PR #131, opened 2025-07-24
Support vLLM backend in Tunix

wang2yn84:
This PR integrates the vLLM backend to Tunix through vLLM sampler and rollout worker. Right now there are couple of limitations in the vLLM backend: 
1. The sharded initial random weights is not properly supported
2. No lora support
3. No logprobs returned
4. It doesn't take external mesh

So for now in all the examples, it runs on the same mesh with the main process and use only TP sharding. Chunked prefill and prefix caching is turned on by default. 

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

### DIFF
--- a/scripts/grpo_demo_llama3.py
+++ b/scripts/grpo_demo_llama3.py
@@ -1,0 +1,796 @@
+import argparse
+import contextlib
+import functools
+import json
+import logging
+import os
+from pprint import pprint
+import re
+
+from flax import nnx
+import grain
+import humanize
+import jax
+from jax import numpy as jnp
+import optax
+from orbax import checkpoint as ocp
+from qwix import lora
+from tqdm.auto import tqdm
+from transformers import AutoTokenizer
+from huggingface_hub import hf_hub_download
+from huggingface_hub import list_repo_files
+from tunix.models.llama3 import model as llama_lib
+from tunix.models.llama3.model import ModelConfig as llama_model_config
+from tunix.models.llama3.params import create_model_from_safe_tensors
+from tunix.rl import rl_cluster as rl_cluster_lib
+from tunix.rl.grpo.grpo_learner import GrpoConfig, GrpoLearner
+from tunix.rl.rollout import base_rollout
+from tunix.sft import metrics_logger
+
+
+print(f"This script is still WIP and you'll need to download all the data to"
+      "local first. Functionality and performance is not guaranteed. Try at "
+      "your own discretion")
+
+os.environ["TPU_BACKEND_TYPE"] = "jax"
+os.environ["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"
+
+# Disable precompilation for faster iteration, need to toggle it back for official run
+os.environ["SKIP_JAX_PRECOMPILE"] = "1"
+
+# vLLM recommends use the old model design
+# os.environ["NEW_MODEL_DESIGN"]= "True"
+
+
+# Parse command line options
+parser = argparse.ArgumentParser(description="Arguments for GRPO demo")
+parser.add_argument("--root-dir", type=str, required=False, help="The root dir of model, data, etc.")
+
+# Parse arguments
+args = parser.parse_args()
+
+# ====== Data ======
+# The data is not available in gcs bucket yet, please manually copy the
+# following data to your local TRAIN_DATA_PATH:
+# /cns/gg-d/home/qwix-dev/rl/grpo/data/gsm8k_train.json
+# /cns/gg-d/home/qwix-dev/rl/grpo/data/gsm8k_test.json
+
+GCS_BUCKET_PREFIX="gcs://tunix/"
+TRAIN_DATA_PATH_SUBDIR = "rl/grpo/data/gsm8k_train.json"
+TEST_DATA_PATH_SUBDIR =  "rl/grpo/data/gsm8k_test.json"
+HF_MODEL_VERSION = "meta-llama/Llama-3.1-8B-Instruct"
+
+TRAIN_FRACTION = 1.0
+
+# Derived Data Path
+GCS_TRAIN_DATA_PATH = os.path.join(GCS_BUCKET_PREFIX, TRAIN_DATA_PATH_SUBDIR)
+GCS_TEST_DATA_PATH= os.path.join(GCS_BUCKET_PREFIX, TEST_DATA_PATH_SUBDIR)
+
+TRAIN_DATA_PATH = os.path.join(args.root_dir, TRAIN_DATA_PATH_SUBDIR)
+TEST_DATA_PATH = os.path.join(args.root_dir, TEST_DATA_PATH_SUBDIR)
+
+VLLM_MODEL_SUBDIR = "rl/grpo/models/"
+VLLM_MODEL_VERSION = os.path.join(args.root_dir, VLLM_MODEL_SUBDIR, HF_MODEL_VERSION)
+if "8B" in HF_MODEL_VERSION:
+  SIMPLIFIED_MODEL_VERSION = "8b"
+else:
+  SIMPLIFIED_MODEL_VERSION = "1b"
+
+# ====== Base Model ======
+NNX_CKPT_DIR=os.path.join(args.root_dir, "rl/grpo/models/", HF_MODEL_VERSION)
+
+# ====== Reproducibility ======
+SEED = 42
+
+# ====== LoRA ======
+RANK = 64
+ALPHA = 64.0
+
+# ====== Sharding ======
+MESH = [(1, jax.device_count()), ("fsdp", "tp")] #YY
+
+# ====== GRPO ======
+# === Generation during GRPO training ===
+MAX_PROMPT_LENGTH = 256
+TOTAL_GENERATION_STEPS = 1024 #YY 768
+# Important to keep a high-ish temperature for varied, diverse responses during
+# training.
+TEMPERATURE = 0.9
+TOP_P = 1.0  # implies we don't do nucleus sampling
+TOP_K = 50
+# The number of times the policy generates multiple responses for a given prompt
+# within a single training step. This corresponds to `G` in Algorithm 1 in the
+# paper. The "group" in GRPO comes from here.
+NUM_GENERATIONS = 4
+
+# === other GRPO configs ===
+# The number of iterations per batch (𝜇 in GRPO algo 1).
+NUM_ITERATIONS = 1
+# The coefficient for the KL divergence penalty (𝛽) in the GRPO loss function.
+# Important to keep a high enough value for this, otherwise, the KL divergence
+# can increase unchecked.
+BETA = 0.08
+# Epsilon value for clipping (𝜀 in GRPO loss in paper). Similar to PPO, for
+# stable updates.
+EPSILON = 0.2
+
+# ====== Training ======
+BATCH_SIZE = 2 # 2 is the max we can do on v5e-8 with llama3 8B model.
+NUM_BATCHES = 1869 # To speed up for quick workflow validation, we can change it to e.g. 2
+# Keep `NUM_TEST_BATCHES` low so that evaluation runs quickly. It can be
+# increased to a max. of 330 (if batch size is 4).
+NUM_TEST_BATCHES = 50 # To speed up for quick workflow validation, we can change it to e.g. 1
+
+EVAL_EVERY_N_STEPS = 10  # this doesn't matter if `TRAIN_FRACTION = 1.0`.
+NUM_EPOCHS = 1  # can potentially train for more epochs
+
+# Number of training steps.
+MAX_STEPS = int(NUM_BATCHES * NUM_ITERATIONS * TRAIN_FRACTION * NUM_EPOCHS)
+
+# === AdamW, warmup, cosine scheduler ===
+LEARNING_RATE = 3e-6
+B1 = 0.9  # Adam beta1
+B2 = 0.99  # Adam beta2
+WEIGHT_DECAY = 0.1
+# == Cosine decay with warmup scheduler ==
+# Linearly increase learning rate from 0. to 5e-6 in the first 10% training
+# steps, and then gradually decrease the learning rate to 0 using cosine
+# scheduler.
+WARMUP_STEPS = 0.1 * MAX_STEPS
+# == Grad clipping ==
+# Grad clipping to prevent large gradients. Found this
+# important to keep KL divergence in check.
+MAX_GRAD_NORM = 0.1
+
+# ====== Checkpoint saving ======
+CKPT_DIR = os.path.join(args.root_dir,
+    "rl/grpo/demo/experiments/llama3/training_runs/2"
+)
+
+SAVE_INTERVAL_STEPS = 500 # To speed up for quick workflow validation, we can change it to e.g. 2
+MAX_TO_KEEP = 4
+DO_MEM_PROFILING = False
+DO_MODEL_DISPLAY = False
+
+# ====== Inference ======
+GENERATION_CONFIGS = {
+    # greedy search
+    "greedy": {"temperature": 1e-2, "top_k": 1, "top_p": 1.0},
+    # some randomness
+    "standard": {"temperature": 0.7, "top_k": 50, "top_p": 0.95},
+    # liberal
+    "liberal": {"temperature": 0.85, "top_k": 2000, "top_p": 1.0},
+}
+
+# ====== Profiler ======
+PROFILER_PATH = os.path.join(args.root_dir, "tunix/rl/grpo/demo/experiments/llama3/profiler")
+
+import gc
+for name, obj in list(globals().items()):
+    if isinstance(obj, jnp.ndarray):
+        del globals()[name]
+gc.collect()
+
+
+# Download data
+def download_hf_checkpoint(repo_id, local_dir):
+  all_files = list_repo_files(repo_id)
+  filtered_files = [f for f in all_files if not f.startswith("original/")]
+
+  for filename in filtered_files:
+    hf_hub_download(
+        repo_id=repo_id,
+        filename=filename,
+        local_dir=local_dir
+    )
+  print(f"Downloaded {filtered_files} to: {local_dir}")
+
+download_hf_checkpoint(HF_MODEL_VERSION, VLLM_MODEL_VERSION)
+
+def download_from_gcs(zip_gcs_path, target_path):
+  return f"""
+    echo "{write_download_from_gcs_sh(zip_gcs_path, target_path)}" > download_from_gcs.sh
+    bash download_from_gcs.sh
+  """
+
+def write_download_from_gcs_sh(zip_gcs_path, target_path):
+  # pylint: disable=anomalous-backslash-in-string
+  return f"""GCS_READ_SUCCESS=0
+while [ \$GCS_READ_SUCCESS -eq 0 ]
+do
+  {{ # try
+      gsutil cp {zip_gcs_path} {target_path} &&
+      echo 'Code download from GCS successful!' && GCS_READ_SUCCESS=1
+  }} || {{ # catch
+      echo 'Failed to read GCS via gsutil, trying again'
+      sleep 10
+  }}
+done"""
+
+# download_from_gcs(GCS_TRAIN_DATA_PATH, TRAIN_DATA_PATH)
+# download_from_gcs(GCS_TEST_DATA_PATH, TEST_DATA_PATH)
+
+
+def load_json_from_local(path):
+  # with gfile.Open(path, "rb") as f:
+  with open(path, "rb") as f:
+    return json.loads(f.read())
+
+
+def show_hbm_usage(title=""):
+  fmt_size = functools.partial(humanize.naturalsize, binary=True)
+  # Force a GC sweep to catch recently deallocated arrays
+  gc.collect()
+  try:
+    import pathwaysutils
+    print(f"Using Pathways compatible HBM stats collector")
+
+    # Track usage per device
+    usage_by_device = {d: 0 for d in jax.local_devices()}
+
+    for obj in gc.get_objects():
+        try:
+            if isinstance(obj, jax.Array):
+                device = obj.device()
+                usage_by_device[device] += obj.size * obj.dtype.itemsize
+        except Exception:
+            continue  # Skip objects that raise
+
+    for d, used in usage_by_device.items():
+        print(f"{title} -- Using {fmt_size(used)} on {d}")
+  except:
+    print(f"Pathways not available. Using non Pathways HBM stats collector")
+
+    for d in jax.local_devices():
+      stats = d.memory_stats()
+      used = stats['bytes_in_use']
+      limit = stats['bytes_limit']
+      print(f'{title} -- Using {fmt_size(used)} / {fmt_size(limit)} ({used/limit:%}) on {d}')
+
+show_hbm_usage()
+
+reasoning_start = "<reasoning>"
+reasoning_end = "</reasoning>"
+solution_start = "<answer>"
+solution_end = "</answer>"
+
+
+SYSTEM_PROMPT = f"""You are given a problem. Think about the problem and \
+provide your reasoning. Place it between {reasoning_start} and \
+{reasoning_end}. Then, provide the final answer (i.e., just one numerical \
+value) between {solution_start} and {solution_end}."""
+
+TEMPLATE = """<start_of_turn>user
+{system_prompt}
+
+{question}<end_of_turn>
+<start_of_turn>model"""
+
+def extract_hash_answer(text: str) -> str | None:
+  if "####" not in text:
+    return None
+  return text.split("####")[1].strip()
+
+
+def get_dataset(path: str) -> grain.MapDataset:
+
+  data = load_json_from_local(path)
+
+  dataset = (
+      grain.MapDataset.source(data)
+      .shuffle(seed=SEED)
+      .map(
+          lambda x: {
+              # passed to model forward pass
+              "prompts": TEMPLATE.format(
+                  system_prompt=SYSTEM_PROMPT, question=x["question"]
+              ),
+              # passed to reward functions
+              "question": x["question"],
+              # passed to reward functions
+              "answer": extract_hash_answer(x["answer"]),
+          }
+      )
+  )
+  return dataset
+
+dataset = get_dataset(TRAIN_DATA_PATH).batch(BATCH_SIZE)[:NUM_BATCHES]
+
+if TRAIN_FRACTION == 1.0:
+  train_dataset = dataset.repeat(NUM_EPOCHS)
+  val_dataset = None
+else:
+  train_dataset = dataset[: int(len(dataset) * TRAIN_FRACTION)]
+  train_dataset = train_dataset.repeat(NUM_EPOCHS)
+
+  val_dataset = dataset[int(len(dataset) * TRAIN_FRACTION) :].repeat(NUM_EPOCHS)
+
+test_dataset = get_dataset(TEST_DATA_PATH).batch(BATCH_SIZE)[:NUM_TEST_BATCHES]
+
+len(train_dataset), len(val_dataset) if val_dataset is not None else 0, len(
+    test_dataset
+)
+
+for ele in train_dataset[:1]:
+  pprint(ele)
+
+MODEL_CONFIG = {
+    "1b": llama_lib.ModelConfig.llama3_1b,
+    "8b": llama_lib.ModelConfig.llama3_8b,
+}
+
+def get_llama_model(ckpt_path, mesh):
+  return create_model_from_safe_tensors(ckpt_path, llama_model_config.llama3_8b(), mesh)
+
+def get_ref_model():
+  ckpt_path = os.path.join(NNX_CKPT_DIR)
+  mesh = jax.make_mesh(*MESH)
+  model_config = MODEL_CONFIG[SIMPLIFIED_MODEL_VERSION]()
+  model = get_llama_model(ckpt_path, mesh)
+  return model, mesh, model_config
+
+def get_lora_model(base_model, mesh=None):
+  if isinstance(base_model, llama_lib.Llama3):
+    module_path = ".*q_proj|.*k_proj|.*v_proj|.*o_proj|.*gate_proj|.*down_proj|.*up_proj"
+  else:
+    module_path = ".*q_einsum|.*kv_einsum|.*gate_proj|.*down_proj|.*up_proj|.*attn_vec_einsum"
+
+  lora_provider = lora.LoraProvider(
+      module_path=(
+          module_path
+      ),
+      rank=RANK,
+      alpha=ALPHA,
+  )
+
+  model_input = base_model.get_model_input()
+  lora_model = lora.apply_lora_to_model(
+      base_model, lora_provider, **model_input
+  )
+
+  with mesh:
+    state = nnx.state(lora_model)
+    pspecs = nnx.get_partition_spec(state)
+    sharded_state = jax.lax.with_sharding_constraint(state, pspecs)
+    nnx.update(lora_model, sharded_state)
+
+  return lora_model
+
+# Reference model
+transformer, mesh, model_config = get_ref_model()
+if DO_MODEL_DISPLAY:
+  nnx.display(transformer)
+
+# Policy model
+# TODO(b/434959964): Supports lora in vLLM Jax backend
+# lora_transformer = get_lora_model(transformer, mesh=mesh)
+
+lora_transformer = transformer
+if DO_MODEL_DISPLAY:
+  nnx.display(lora_transformer)
+
+show_hbm_usage("After creating the reference lora model")
+
+match_format = re.compile(
+    rf"^[\s]{{0,}}"
+    rf"{reasoning_start}.+?{reasoning_end}.*?"
+    rf"{solution_start}(.+?){solution_end}"
+    rf"[\s]{{0,}}$",
+    flags=re.MULTILINE | re.DOTALL,
+)
+
+match_format.search(
+    f"{reasoning_start}Let me"
+    f" think!{reasoning_end}{solution_start}2{solution_end}",
+)
+
+def match_format_exactly(prompts, completions, **kargs):
+  scores = []
+  for completion in completions:
+    score = 0
+    response = completion
+    # Match if format is seen exactly!
+    if match_format.search(response) is not None:
+      score += 3.0
+    scores.append(score)
+  return scores
+
+def match_format_approximately(prompts, completions, **kargs):
+  scores = []
+
+  for completion in completions:
+    score = 0
+    response = completion
+    # Count how many keywords are seen - we penalize if too many!
+    # If we see 1, then plus some points!
+    score += 0.5 if response.count(reasoning_start) == 1 else -0.5
+    score += 0.5 if response.count(reasoning_end) == 1 else -0.5
+    score += 0.5 if response.count(solution_start) == 1 else -0.5
+    score += 0.5 if response.count(solution_end) == 1 else -0.5
+    scores.append(score)
+  return scores
+
+def check_answer(prompts, completions, answer, **kargs):
+  responses = completions
+
+  extracted_responses = [
+      guess.group(1) if (guess := match_format.search(r)) is not None else None
+      for r in responses
+  ]
+
+  scores = []
+  for guess, true_answer in zip(extracted_responses, answer):
+    score = 0
+    if guess is None:
+      scores.append(0)
+      continue
+    # Correct answer gets 3 points!
+    if guess == true_answer:
+      score += 3.0
+    # Match if spaces are seen
+    elif guess.strip() == true_answer.strip():
+      score += 1.5
+    else:
+      # We also reward it if the answer is close via ratios!
+      # Ie if the answer is within some range, reward it!
+      try:
+        ratio = float(guess) / float(true_answer)
+        if ratio >= 0.9 and ratio <= 1.1:
+          score += 0.5
+        elif ratio >= 0.8 and ratio <= 1.2:
+          score += 0.25
+        else:
+          score -= 1.0  # Penalize wrong answers
+      except:
+        score -= 0.5  # Penalize
+    scores.append(score)
+  return scores
+
+match_numbers = re.compile(
+    rf"{solution_start}.*?([\d\.]{{1,}})", flags=re.MULTILINE | re.DOTALL
+)
+match_numbers.findall(f"{solution_start}  0.34  {solution_end}")
+
+def check_numbers(prompts, completions, answer, **kargs):
+  question = kargs["question"]
+  responses = completions
+
+  extracted_responses = [
+      guess.group(1) if (guess := match_numbers.search(r)) is not None else None
+      for r in responses
+  ]
+
+  scores = []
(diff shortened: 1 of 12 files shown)


scripts/grpo_demo_llama3.py
```diff
@@ -0,0 +1,724 @@
...
+      "your own discretion")
+
+os.environ["TPU_BACKEND_TYPE"] = "jax"
+os.environ["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"
+os.environ["SKIP_JAX_PRECOMPILE"] = "1"  # Disable precompilation
+# os.environ["JAX_RANDOM_WEIGHTS"] = "True"
+
+# ====== Data ======
+TRAIN_DATA_PATH = "/workspace/rl/grpo/data/gsm8k_train.json"
+TEST_DATA_PATH = "/workspace/rl/grpo/data/gsm8k_test.json"
```
annyan09023: can you use argparse to allow user to specify their TRAIN_DATA_PATH and TEST_DATA_PATH? I don't like forcing creating a workspace/rl/grpo/data folder, alternatively, you can have assets folder and just check in gsm8k dataset. Maybe the latter is more commonly used

scripts/grpo_demo_llama3.py
```diff
@@ -0,0 +1,724 @@
...
+
+VLLM_MODEL_VERSION="/workspace/rl/grpo/models/meta-llama/8b/"
+VLLM_MODEL_VERSION="meta-llama/Llama-3-8B-Instruct"
+
+os.environ["TPU_BACKEND_TYPE"] = "jax"
+os.environ["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"
```
annyan09023: duplicated lines with L37-39

scripts/grpo_demo_llama3.py
```diff
@@ -0,0 +1,724 @@
...
+
+# ====== Reproducibility ======
+SEED = 42
+
+# ====== LoRA ======
+RANK = 64
+ALPHA = 64.0
+
+# ====== Sharding ======
+MESH = [(1, 8), ("fsdp", "tp")] #YY
```
annyan09023: to be a bit more flexible, do (1, len(jax.devices())

scripts/grpo_demo_llama3.py
```diff
@@ -0,0 +1,724 @@
...
+# steps, and then gradually decrease the learning rate to 0 using cosine
+# scheduler.
+WARMUP_STEPS = 0.1 * MAX_STEPS
+# == Grad clipping ==
+# Grad clipping to prevent large gradients. Found this
+# important to keep KL divergence in check.
+MAX_GRAD_NORM = 0.1
+
+# ====== Checkpoint saving ======
+CKPT_DIR = (
```
annyan09023: can you take a base_dir from argparser and then use that folder/ckpt as the ckpt storage place

scripts/grpo_demo_llama3.py
```diff
@@ -0,0 +1,724 @@
...
+    # greedy search
+    "greedy": {"temperature": 1e-2, "top_k": 1, "top_p": 1.0},
+    # some randomness
+    "standard": {"temperature": 0.7, "top_k": 50, "top_p": 0.95},
+    # liberal
+    "liberal": {"temperature": 0.85, "top_k": 2000, "top_p": 1.0},
+}
+
+# ====== Profiler ======
+PROFILER_PATH = "/workspace/tunix/rl/grpo/demo/experiments/llama3/profiler"
```
annyan09023: ditto, take in a base_dir and structure folders as needed

scripts/grpo_demo_llama3.py
```diff
@@ -0,0 +1,724 @@
...
+def show_hbm_usage(title=""):
+  fmt_size = functools.partial(humanize.naturalsize, binary=True)
+
+  for d in jax.local_devices():
+    stats = d.memory_stats()
+    used = stats['bytes_in_use']
+    limit = stats['bytes_limit']
+    print(f'{title} -- Using {fmt_size(used)} / {fmt_size(limit)} ({used/limit:%}) on {d}')
+
+show_hbm_usage()
```
annyan09023: this is not functional with pathways right?

tests/generate/vllm_sampler_test.py
```diff
@@ -0,0 +1,204 @@
...
+class vLLMSamplerTest(absltest.TestCase):
+
+  def setUp(self) -> None:
+    super().setUp()
+    mesh_shape = (1, len(jax.devices()))  # e.g., (1, 8) for v2-8
+    axis_names = ("fsdp", "tp")  #
+    self.mesh = jax.make_mesh(mesh_shape,
+                          axis_names,
+                          devices=jax.devices())
+    self.model_path = "meta-llama/Llama-3-8B-Instruct"
```
annyan09023: how to download this llama3-8b model, is that from hf, if so, in the setup, you can use hf api to download it? and then delete it after tear down

tunix/generate/base_sampler.py
```diff
@@ -0,0 +1,60 @@
...
+    logits: list[jax.Array] | jax.Array
+
+    # Tokens corresponding to the generated samples.
+    tokens: list[jax.Array] | jax.Array
+
+    # Left padded prompt tokens.
+    padded_prompt_tokens: jax.Array
+
+
+class SamplerBase(abc.ABC):
```
annyan09023: nit: rename the file as sampler_base instead of base_sampler

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+class vLLMSampler(base_sampler.SamplerBase):
+  def __init__(
+        self,
+        model: nnx.module,
+        tokenizer: Any,
+        mesh: jax.sharding.Mesh,
+        max_model_len: int,
+        lora_config: Optional[Dict[str, Any]],
+        model_version: Optional[str], # We still need the model version for now, will remove it later
+        gpu_memory_utilization: Optional[float] = 0.3,
```
annyan09023: do we need this?

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+import os
+
+class vLLMSampler(base_sampler.SamplerBase):
+  def __init__(
+        self,
+        model: nnx.module,
+        tokenizer: Any,
+        mesh: jax.sharding.Mesh,
+        max_model_len: int,
+        lora_config: Optional[Dict[str, Any]],
```
annyan09023: nit: move this after model_version and provide a default value None

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+        model_version: Optional[str], # We still need the model version for now, will remove it later
```
annyan09023: what's the expected model_version, model_version from hf?

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+from tunix.generate import base_sampler
+import os
+
+class vLLMSampler(base_sampler.SamplerBase):
+  def __init__(
+        self,
+        model: nnx.module,
+        tokenizer: Any,
+        mesh: jax.sharding.Mesh,
+        max_model_len: int,
```
annyan09023, line 57: what's max_model_len?

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+    self.gpu_memory_utilization = gpu_memory_utilization
+
+    self.args = self._vllm_config()
+    self.llm = LLM(**self.args)
+
+    self.mappings = self.get_to_hf_mappings
+    if lora_config is not None:
+      self.mappings |= self.get_lora_to_hf_mappings
+
+    # This is how we load the weights to vLLM now, consider load the orbax checkpoint in vLLM directly
```
annyan09023: please add a ticket for that

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+
+  def update_params(self, updated_weights: nnx.Module):
+    self.llm.llm_engine.model_executor.collective_rpc("sync_weights", args=(updated_weights, self.mappings, self.get_to_hf_transpose_keys))
+
+  def _vllm_config(self):
+    args = {}
+    args["additional_config"] = {}
+    args["model"] = self.model_version
+    args["max_model_len"] = self.max_model_len
+    args["tensor_parallel_size"] = self.mesh.shape["tp"]
```
annyan09023, line 99: just curious, for MOE, we also need expert_parallelism, does vllm support that?

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+    return args
+
+  @property
+  def _model_runner(self):
+    return self.llm.llm_engine.model_executor.driver_worker.model_runner
+
+  @property
+  def transformer(self):
+    # vLLM doesn't expose the underlying model
+    return None
```
annyan09023: raise NotImplementedError

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+
+  @property
+  def transformer_state(self):
+    return self._model_runner.state
```
annyan09023, line 115: is this self._model_runner.state the non trainable state from the model? if it's a different concept, better to raise NotImplementedError

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+
+  def tokenize(self, input_string: str) -> List[int]:
+    """Tokenizes the input string."""
+    input_ids = self.tokenizer.encode(input_string)
+    bos_tok = [self.tokenizer.bos_id()] if self.tokenizer.bos_id() else []
```
annyan09023: what if input str already starts with bos_id? do you need to have a check here?

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+  def _get_logprobs_from_vllm_output(self, logprobs: List[Optional[Dict[int, Any]]]):
+      logging.info("vLLM Jax backend doesn't support returnning log probs yet!")
+      return []
+      # Below is to get the log probs from vLLM output
+      # assert logprobs[0] is not None, f"Logprobs are missing"
+      # assert len(logprobs[0]) == 1, f"The log probs contains more than 1 ({len(logprobs[0])} token per position"
+      # return [list(logprob_dict.values())[0].logprob for logprob_dict in logprobs]
+
+  def detokenize(self, input_strings, return_logits, request_outputs: List[RequestOutput]):
+    print("-" * 50)
```
annyan09023: remove?

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+
+  def _get_logprobs_from_vllm_output(self, logprobs: List[Optional[Dict[int, Any]]]):
+      logging.info("vLLM Jax backend doesn't support returnning log probs yet!")
+      return []
+      # Below is to get the log probs from vLLM output
+      # assert logprobs[0] is not None, f"Logprobs are missing"
+      # assert len(logprobs[0]) == 1, f"The log probs contains more than 1 ({len(logprobs[0])} token per position"
+      # return [list(logprob_dict.values())[0].logprob for logprob_dict in logprobs]
+
+  def detokenize(self, input_strings, return_logits, request_outputs: List[RequestOutput]):
```
annyan09023: can you add type annotation for inputs and returns, for tensor, also add shape info, python is hard to read w/o type annotation

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+    print("-" * 50)
+    generations = len(request_outputs[0].outputs)
+    decoded_outputs = [[] for _ in range(generations)]
+    out_logits = [[] for _ in range(generations)]
+    out_tokens = [[] for _ in range(generations)]
+    for input_string, multi_sampling_output in zip(input_strings, request_outputs):
```
annyan09023, line 142: in our original sampler, we check on eos and then truncate the output, is that vllm returns already only contain valid chars, so there is no need to check on eos anymore?

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+      prompt_logits = self._get_logprobs_from_vllm_output(multi_sampling_output.prompt_logprobs)
+      for idx, single_output in enumerate(multi_sampling_output.outputs):
+        out_tokens[idx].append(single_output.token_ids)
+        decoded_outputs[idx].append(self.tokenizer.decode(single_output.token_ids))
+        if return_logits:
+          generation_logits = self._get_logprobs_from_vllm_output(single_output.logprobs)
+          out_logits[idx].append(prompt_logits + generation_logits)
+        print(f"Prompt: {input_string!r}\n\nGenerated text: {decoded_outputs[idx][-1]!r}\n\n Generated text length: {len(decoded_outputs[idx][-1])} {len(out_tokens[idx][-1])=}")
```
annyan09023: log debug

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+        print("-" * 50)
```
annyan09023: remove

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+    return base_sampler.SamplerOutput(
+      text=decoded_outputs[0],
+      logits=out_logits[0],
+      tokens=all_output_ids,
+      padded_prompt_tokens=all_input_ids,
+    )
+
+  # For now, we are still passing sharding to vLLM, consider removing it after switching to the reshard API
+  @property
+  def get_to_hf_mappings(self):
```
annyan09023: is this model specific, for example, llama, qwen and deepseek should have different mapping? if so, we should expect this to be a method from model

tunix/rl/rl_cluster.py
```diff
@@ -167,7 +174,19 @@ def _init_cluster(self):
...
+          cache_config=base_rollout.CacheConfig(
+              cache_size=self.cluster_config.rollout_config.kv_cache_size,
+              num_layers=0,  # vLLM doesn't rely on these model config
+              num_kv_heads=0,
+              head_dim=0,
+          ),
+          mesh=self.r2m[Role.ROLLOUT],
+          lora_config=self.cluster_config.rollout_lora_config,
+          model_version=self.cluster_config.model_version
+      )
```
annyan09023, line 188: add else, it's a general good practice to do if, elif, else raise ValueError to keep code robustness

tunix/rl/rollout/vllm_rollout.py
```diff
@@ -0,0 +1,105 @@
...
+      mesh: jax.sharding.Mesh,
+      lora_config: Optional[Dict[str, str]] = None,
+      model_version: str = "meta-llama/Llama-3.1-8B",
+  ):
+    self.model = model
+    self.mesh = mesh
+    self._sampler = vllm_sampler.vLLMSampler(
+      model=model,
+      tokenizer=tokenizer,
+      max_model_len=cache_config.cache_size,
```
annyan09023: why do we still need a cache config? isn't cache managed by vllm with paged attention?

wang2yn84: We only need the cache size from the cache config. The reason why I still have cache_config in the rollout init is to keep consistent with the vanilla rollout API so that it's easier to swap the function call.

tunix/generate/vllm_sampler.py
wang2yn84: It could be either the model version from HF or the local path with the downloaded HF checkpoint.

tunix/generate/base_sampler.py
wang2yn84: Hmm, I'm keeping it consistent with base_rollout, vanilla_rollout. I'm ok either way. Do you want me to update them all?

scripts/grpo_demo_llama3.py
wang2yn84: Thank you for the suggestion! Yea this script was pretty unrefined. That's indeed much better. Updated!

scripts/grpo_demo_llama3.py
wang2yn84: Removed

scripts/grpo_demo_llama3.py
wang2yn84: Updated

scripts/grpo_demo_llama3.py
wang2yn84: Done

scripts/grpo_demo_llama3.py
wang2yn84: Done

scripts/grpo_demo_llama3.py
wang2yn84: Updated to make it compatible for both Pathways and non Pathways env.

tests/generate/vllm_sampler_test.py
wang2yn84: Added. I didn't explicitly delete it because it's now downloading to /tmp and will rely on system to delete them.

tunix/generate/vllm_sampler.py
wang2yn84: Hmm, the highlight here is not accurate. If you mean gpu_memory_utilization, we need it to tune the HBM usage.

tunix/generate/vllm_sampler.py
wang2yn84: Done

tunix/generate/vllm_sampler.py
wang2yn84: Done.

tunix/generate/vllm_sampler.py
wang2yn84: It's copied to the vLLM so it's not trainable. But the transformer_state() is called in update_param as old_params, it still works for vLLM. Wdyt?

tunix/generate/vllm_sampler.py
wang2yn84: Yes it's indeed model specific. Thank you for the suggestion, that makes a lot of sense!

tunix/rl/rl_cluster.py
wang2yn84: Ooops, I missed this one. Added. Thank you!

tunix/generate/vllm_sampler.py
wang2yn84: It defines the maximum length of each batch, including prompt + generations

tunix/generate/vllm_sampler.py
wang2yn84: From what I see, currently Jax backend doesn't support EP yet. But vLLM has the config knob.

tunix/generate/vllm_sampler.py
wang2yn84: Done

tunix/generate/vllm_sampler.py
wang2yn84: Done.

tunix/generate/vllm_sampler.py
wang2yn84: Done

tunix/generate/vllm_sampler.py
wang2yn84: Added.

tunix/generate/vllm_sampler.py
wang2yn84: Done.

tunix/generate/vllm_sampler.py
wang2yn84: Done.

tunix/rl/rollout/vllm_rollout.py
annyan09023: then you can make it cache_config_or_size, then it can also accept an int ony

tunix/generate/vllm_sampler.py
annyan09023: can you add both to the method arguments pydoc (model_version and max_model_len)?

tunix/generate/vllm_sampler.py
annyan09023: so it should be hbm_memory_utilization instead? I know vllm naming is gpu_memory_utilization, but we can give it a better name. And can you also add in pydoc if it's set, what does it mean (if memory % > 30%, it will error out? in that case, when do we need < 100% number) and what does it mean for unsetting?

tunix/generate/vllm_sampler.py
annyan09023: then okay to keep it

tunix/generate/base_sampler.py
annyan09023: Usually the file name is consistent with class name, because your class name is called samplerBase, so better to have the file name in consistent with the class name

scripts/llama3_example.py
```diff
@@ -0,0 +1,65 @@
...
+          add_generation_prompt=True,
+          # enable_thinking=True,
+      )
+    )
+  return out
+
+inputs = templatize(
+    [
+        "tell me about world war 2",
+        "讲几句人话",
```
annyan09023: better to change to another chinese prompt :P

scripts/llama3_example.py
```diff
@@ -0,0 +1,65 @@
...
+from flax import nnx
+from huggingface_hub import snapshot_download
+from transformers import AutoTokenizer
+from tunix.generate import sampler
+from tunix.models.llama3 import params
+from tunix.models.llama3 import model
+
+
+MODEL_VERSION="meta-llama/Llama-3-8B-Instruct"
+MODEL_CP_PATH='/tmp/model/' + MODEL_VERSION
```
annyan09023: this is not tmp folder for every fs, use tempfile.mkdtemp() instead

scripts/grpo_demo_llama3.py
```diff
@@ -0,0 +1,736 @@
+import argparse
```
annyan09023, line 1: instead of adding a new script, you can add to current grpo_demo notebook by using a flag to control VLLM or vanilla rollout?

scripts/llama3_example.py
```diff
@@ -0,0 +1,65 @@
...
+
+print(f"Make sure you logged in to the huggingface cli.")
+
+# snapshot_download(
+#     repo_id=MODEL_VERSION,
+#     local_dir=MODEL_CP_PATH,
+#     local_dir_use_symlinks=False  # optional: avoids symlinks for portability
+# )
+
+MODEL_CP_PATH = "/workspace/rl/grpo/models/meta-llama/8b"
```
annyan09023: is this needed? isn't MODEL_CP_PATH defined in l12?

scripts/llama3_example.py
```diff
@@ -0,0 +1,65 @@
...
+
+mesh = jax.make_mesh((1, 8), ('fsdp', 'tp'))
```
annyan09023: use jax.devices() instead of hard coded 8

scripts/llama3_example.py
```diff
@@ -0,0 +1,65 @@
...
+config = model.ModelConfig.llama3_8b()  # pick corresponding config based on model version
+llama3 = params.create_model_from_safe_tensors(MODEL_CP_PATH, config, mesh)
+# nnx.display(llama3)
+
+tokenizer = AutoTokenizer.from_pretrained(MODEL_CP_PATH)
+tokenizer.pad_token_id = 0
+
+def templatize(prompts):
```
annyan09023, line 33: you can add enable_thinking as a pass

scripts/llama3_example.py
```diff
@@ -0,0 +1,65 @@
...
+  out = []
+  for p in prompts:
+    out.append(
+        tokenizer.apply_chat_template(
+          [
+              {"role": "user", "content": p},
+          ],
+          tokenize=False,
+          add_generation_prompt=True,
+          # enable_thinking=True,
```
annyan09023: remove

tunix/generate/base_sampler.py
tianshub: just BaseSampler as class name? so it's consistent with rollout class names

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,213 @@
...
+        model: nnx.module,
+        tokenizer: Any,
+        mesh: jax.sharding.Mesh,
+        max_model_len: int,
+        model_version: Optional[str], # We still need the model version for now, will remove it later
+        lora_config: Optional[Dict[str, Any]] = None,
+        gpu_memory_utilization: Optional[float] = 0.3,
+    ):
+    os.environ["TPU_BACKEND_TYPE"] = "jax"
+    os.environ["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"
```
tianshub: this should be in the module scope instead of ctor?

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,213 @@
...
+
+    self.to_hf_transpose_keys = None
+    try:
+      self.to_hf_transpose_keys  = model.to_hf_transpose_keys()
+    except:
+      logging.info("to_hf_transpose_keys is not defined in the model")
+
+    # TODO(b/434741253): Consider switch to loading the orbax checkpoint in
+    # vLLM directly
+    _, params = nnx.split(model)
```
tianshub: this can just be `nnx.state(model)`

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,213 @@
...
+    self.llm = LLM(**self.args)
+
+    try:
+      self.mappings = model.to_hf_mappings()
+    except:
+      raise AttributeError("to_hf_mappings is not defined in the model")
+
+    if lora_config is not None:
+      try:
+        lora_to_hf_mappings = model.lora_to_hf_mappings()
```
tianshub: where's this being used?

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,213 @@
...
+    self.max_model_len = max_model_len
+    self.mesh = mesh
+    self.lora_config = lora_config
+    self.gpu_memory_utilization = gpu_memory_utilization
+
+    self.args = self._vllm_config()
+    self.llm = LLM(**self.args)
+
+    try:
+      self.mappings = model.to_hf_mappings()
```
tianshub: to_hf_mappings

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,213 @@
...
+    except:
+      logging.info("to_hf_transpose_keys is not defined in the model")
+
+    # TODO(b/434741253): Consider switch to loading the orbax checkpoint in
+    # vLLM directly
+    _, params = nnx.split(model)
+    self.update_params(updated_weights=params)
+
+  def update_params(self, updated_weights: nnx.Module):
+    self.llm.llm_engine.model_executor.collective_rpc(
```
tianshub, line 83: why does it expect a nnx.Module instead of a pytree for params?

And seems like this will result in doubling the rollout HBM usage right? which I assume is fine if we just release KV cache?

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,213 @@
...
+import jax
+import jax.numpy as jnp
+from tunix.generate import utils
+from tunix.generate import base_sampler
+import os
+
+class vLLMSampler(base_sampler.SamplerBase):
+  def __init__(
+        self,
+        model: nnx.module,
```
tianshub: This model arg could be a bit misleading, as it's not the actual model used for vLLM, but instead only the to_hf model conversion metadata and initial params are needed. Another risk here is if model reference is being accidentally kept alive in the sampler class, it could result in extra HBM occupation and hard to be detected. 

Instead, shall we consider pass in the to_hf_model config as dedicated arg here and force an initial call to update_weights?

tunix/rl/rollout/vllm_rollout.py
```diff
@@ -0,0 +1,103 @@
...
+    self._sampler.update_params(params)
+
+  def pad_id(self) -> int:
+    return self._sampler.tokenizer.pad_id()
+
+  def eos_id(self) -> int:
+    return self._sampler.tokenizer.eos_id()
+
+  def model(self) -> nnx.Module:
+    return self._sampler.transformer
```
tianshub: wouldn't this fail since vllm_sampler doesn't implement transformer property?

tunix/rl/rollout/vllm_rollout.py
```diff
@@ -0,0 +1,103 @@
...
+  def __init__(
+      self,
+      model: nnx.Module,
+      tokenizer: Any,
+      cache_config: base_rollout.CacheConfig,
+      mesh: jax.sharding.Mesh,
+      lora_config: Optional[Dict[str, str]] = None,
+      model_version: str = "meta-llama/Llama-3.1-8B",
+  ):
+    self.model = model
```
tianshub: this looks dangerous as we are holding extra ref to the model which is not needed right?

tests/generate/vllm_sampler_test.py
```diff
@@ -0,0 +1,232 @@
...
+                {"role": "user", "content": p},
+            ],
+            tokenize=False,
+            add_generation_prompt=True,
+        )
+      )
+    return out
+
+  def test_vllm_sampler(self):
+      if "8B" in self.repo_id:
```
tianshub, line 133: isn't it pretty expensive to test on 8B model in unit test?

tests/generate/vllm_sampler_test.py
```diff
@@ -0,0 +1,232 @@
...
+          seed=0,
+          echo=False,
+          pad_output=True,  # Use padding for output
+      )
+      # Print the outputs.
+      print("-" * 50)
+      print(f"Vanilla Generated text: {vanilla_output.text}")
+
+      print("-" * 50)
+      print(f"vLLM Generated text: {vllm_output.text}")
```
tianshub: for greedy sampling, are the outputs same between the two?

scripts/grpo_demo_llama3.py
wang2yn84: The difference is not just the vllm part. There are google imports and other imcompatibilities. In the long run, I'd like to convert the notebook to python version ([go/colab/python](https://goto.google.com/colab/python). Personally, I feel like the script version is much easier to execute in the 3P env.

scripts/llama3_example.py
wang2yn84: Removed.

scripts/llama3_example.py
wang2yn84: haha, changed.

scripts/llama3_example.py
wang2yn84: Updated. End up using gettempdir() because mkdtemp() returns a different temp dir every time, it consumes too much space.

tunix/generate/vllm_sampler.py
wang2yn84: Sure, moved to the module scope.

tunix/generate/vllm_sampler.py
wang2yn84: It won't take effect until the vLLM Jax backend supports lora. I'm in discussion with them to add a simple Qwix based lora model support. Added a bug tracker for that.

tunix/generate/vllm_sampler.py
wang2yn84: Changed to property. Is that what you mean? Or are you suggesting to rename the self.mappings to self.to_hf_mappings?

tunix/rl/rollout/vllm_rollout.py
wang2yn84: Good idea! Updated!

tunix/generate/vllm_sampler.py
wang2yn84: Good catch! I'm actually using the Pytree on the vLLM Jax backend. Missed this one. Updated!

For the HBM usage, yea right now we are copying all the weights from trainer to vLLM so we are doubling the weights. The reason is because there are couple of tensor that's transposed. One optimization we can do in the future is to share those untransposed tensors between trainer and sampler and only copy the tranposed ones. To release KV cache on the fly, current vLLM Jax backend doesn't support it, there are additional work there.

tunix/generate/vllm_sampler.py
wang2yn84: Hmm, sure I can do that, but the user experience looks weird cuz it's not very natural to sync weights after creating the sampler. Shall we add a new API called load_checkpoints that called into update_params for now? Ideally, we probably should support loading weights from checkpoints directly.

tunix/rl/rollout/vllm_rollout.py
wang2yn84: Good catch! Updated the sampler impl when resolving the previous comments and missed this one. Updated.

tunix/rl/rollout/vllm_rollout.py
wang2yn84: We do need this to calculate the log probs. But vLLM supports returning log probs end of last week and we don't need to hold the model reference any more. Will update soon.

tests/generate/vllm_sampler_test.py
wang2yn84: Yea agree on that, unfortunately the vLLM Jax backend has some issue with the llama 1B model and it's not working ...

tests/generate/vllm_sampler_test.py
wang2yn84: Unfortunately no, with exactly the same weights, they outputs differently. I talked to them in the group chat and they acknowledge the difference. They are going to implement 1 layer logits checker to guarantee the model quality.

tunix/generate/vllm_sampler.py
wang2yn84: Ah, I see. Updated.

tunix/generate/base_sampler.py
wang2yn84: Yea, updated to be consistent. Done.

tunix/generate/vllm_sampler.py
wang2yn84: True, changed to hbm_utilization instead and added the pydoc to explain the usage. Basically if you set it to 0.3, the vLLM engine will make sure the HBM usage (weights + kv cache) ~= 30% of available HBM.

tunix/generate/vllm_sampler.py
wang2yn84: Sure. Done

tunix/generate/vllm_sampler.py
wang2yn84: Yes vLLM has this feature by setting the sampling param. Added.

tunix/generate/vllm_sampler.py
wang2yn84: Changing to property doesn't work, I guess that might be related to we split and merge the model. Changing back to the static method.

tunix/rl/rollout/vllm_rollout.py
wang2yn84: I tested the returning log probs feature in vLLM and it's not fully functional yet. We might still have to keep the model reference for now

tests/generate/vllm_sampler_test.py
wang2yn84: I added the assertions to check against the text outputs. You can see the diff in the outputs.

tunix/generate/vllm_sampler.py
wang2yn84: Actually vllm still returns 1 eos id even if we enabled the stop on eos feature. Added the logics to remove the eos_id from vllm outputs.

tunix/generate/vllm_sampler.py
wang2yn84: Added the mappings as init inputs. Added the load_checkpoint API. Ptal.

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,236 @@
...
+  # TODO(b/434975493): Consider Release KV cache on the fly
+  def update_params(self, updated_weights: jaxtyping.PyTree):
+    self.llm.llm_engine.model_executor.collective_rpc(
+      "sync_weights",
+      args=(updated_weights, self.mappings, self.to_hf_transpose_keys))
+
+  def load_checkpoint(self, path_or_weights: str | jaxtyping.PyTree):
+    # TODO(b/434741253): Consider support orbax checkpoint loading
+    if isinstance(path_or_weights, jaxtyping.PyTree):
+      self.update_params(updated_weights=path_or_weights)
```
annyan09023, line 90: add else and throw not implemented error

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,236 @@
...
+          self.sampling_params.top_p = top_p
+      if top_k is not None:
+          self.sampling_params.top_k = top_k
+      if seed is not None:
+        self.sampling_params.seed = seed
+
+    prompt_ids = [self.tokenize(x) for x in input_strings]
+    outputs = self.llm.generate(
+       prompts=None,
+       prompt_token_ids=prompt_ids,
```
annyan09023, line 201: why we are providing prompt_token_ids instead of prompts? I assume providing prompts is less error prone, so that vllm can do padding, using the right tokenizer, generating the right attention mask?

tunix/rl/rollout/vllm_rollout.py
wang2yn84: Since we are now using vllm returned logprobs, remove the additional reference we keep in the vllmRollout class. Updated.

annyan09023 (approved):

tunix/generate/vllm_sampler.py
wang2yn84: Discussed offline, the initial motivation is because vLLM only supports HF tokenizer. For other type of tokenizers, you'll need to run through their script to convert it to be HF compatible. Using our tokenizer adapter enables user to use external tokenizer programmatically.

tunix/generate/vllm_sampler.py
tianshub: just rename mappings to to_hf_mappings

tunix/generate/vllm_sampler.py
```diff
@@ -0,0 +1,232 @@
...
+  def __init__(
+        self,
+        tokenizer: Any,
+        mesh: jax.sharding.Mesh,
+        max_model_len: int,
+        model_version: str,
+        to_hf_mappings: Optional[Dict[str, str]] = None,
+        lora_to_hf_mappings: Optional[Dict[str, str]] = None,
+        to_hf_transpose_keys: Optional[Dict[str, Tuple[int]]] = None,
+        lora_config: Optional[Dict[str, Any]] = None,
```
tianshub: consider wrapping these in a standalone config class, since these should be all provided or not provided.

tianshub (approved):

tunix/generate/vllm_sampler.py
wang2yn84: Done.

tunix/generate/vllm_sampler.py
wang2yn84: Good suggestion. Done!

tunix/generate/vllm_sampler.py
wang2yn84: We will need to combine the lora mapping with the hf mapping and therefore I give it a generic name here.

PR #148, opened 2025-07-31
Init vLLM model with random weights, bypassing from loading from HF.

wang2yn84:
We don't really need to load weights from HF because we will sync weights from trainer to sampler anyway. Bypass it will speed up bring up time and space.

### DIFF
--- a/scripts/grpo_demo_llama3.py
+++ b/scripts/grpo_demo_llama3.py
@@ -50,16 +50,9 @@ print(
     "your own discretion"
 )
 
-os.environ["TPU_BACKEND_TYPE"] = "jax"
-os.environ["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"
-
 # Disable precompilation for faster iteration, need to toggle it back for
 # official run
 os.environ["SKIP_JAX_PRECOMPILE"] = "1"
-
-# vLLM recommends use the old model design
-# os.environ["NEW_MODEL_DESIGN"]= "True"
-
 
 # Parse command line options
 parser = argparse.ArgumentParser(description="Arguments for GRPO demo")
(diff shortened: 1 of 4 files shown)


tianshub (approved):

scripts/grpo_demo_llama3.py
```diff
@@ -56,7 +56,8 @@
 # Disable precompilation for faster iteration, need to toggle it back for
 # official run
 os.environ["SKIP_JAX_PRECOMPILE"] = "1"
-
```
annyan09023, line 59: can it be part of vllm rollout worker or it has to stay with user code?

wang2yn84: That's true. Moved to the vllm sampler, only keep the SKIP_JAX_COMPILE in the user code. Updated!

tunix/generate/vllm_sampler.py
```diff
@@ -27,8 +27,14 @@
 from vllm.outputs import RequestOutput
 
 
+# Enable Jax backend
 os.environ["TPU_BACKEND_TYPE"] = "jax"
+# Colocate vllm engine and worker in the main process
 os.environ["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"
+# Init vLLM model with random weights
```
annyan09023: Init vLLM model with random weights because model weights are synced from trainer later on

tests/generate/vllm_sampler_test.py
```diff
@@ -30,14 +30,11 @@
 os.environ["TPU_BACKEND_TYPE"] = "jax"
 os.environ["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"
 os.environ["SKIP_JAX_PRECOMPILE"] = "1"
+os.environ["JAX_RANDOM_WEIGHTS"] = "True"
```
annyan09023: remove?

tunix/generate/vllm_sampler.py
wang2yn84: Updated.

tests/generate/vllm_sampler_test.py
wang2yn84: Done.

PR #192, opened 2025-08-12
Add weight sync utils for scanned model

wenxindongwork:


tunix/generate/vllm_sampler.py
```diff
@@ -81,14 +82,18 @@ def __init__(
...
+          from the mesh. Defaults to -1.
     """
     self.tokenizer = tok_adapter.TokenizerAdapter(tokenizer)
     self.model_version = model_version
     self.max_model_len = max_model_len
     self.mesh = mesh
     self.lora_config = mapping_config.lora_config
     self.hbm_utilization = hbm_utilization
-
+    self.tp = self.mesh.shape["tp"] if "tp" in self.mesh.shape else tp
```
lc5211: Hmm, it feels a little bit hacky to ask for a  specifc custom `tp` sharding dim...
With what kind of mesh configuration that `tp` is missed?

tunix/generate/utils.py
```diff
@@ -463,6 +472,71 @@ def process_entry(src_keys, src_val):
...
+      if layer_axis == 0:
+          return tensor[layer_idx]
+      elif layer_axis == 1:
+          return tensor[:, layer_idx]
+      elif layer_axis == 2:
+          return tensor[:, :, layer_idx]
+      elif layer_axis == 3:
+          return tensor[:, :, :, layer_idx]
+      else:
+          raise ValueError(f"Unsupported layer axis: {layer_axis}")
```
lc5211: Maybe add the tensor shape as well for more information to debug?

tunix/generate/utils.py
```diff
@@ -359,12 +359,28 @@ def build_flat_dict(
           else:
             src_parts.append(part)
         actual_src = '.'.join(src_parts)
-        new_flat_dict[actual_src] = v, sharding
+        # Check if this is a scanned parameter (has 'layer' in sharding spec)
+        if sharding and 'layer' in sharding:
```
lc5211, line 363: Is it always the case that vLLM doesn't support scanned layer?
I wonder if it's possible for us to control when to enable scanned layer.

lc5211 (commented):
Thanks for the changes!

tunix/generate/vllm_sampler.py
wenxindongwork: I see... I was passing in the MaxText mesh, that was why. I reverted this change. 

```
TOTAL_TPU_TO_USE = 8
MESH = [(1, TOTAL_TPU_TO_USE), ("fsdp", "tp")]  # YY

mesh =   model_mesh = jax.make_mesh(*MESH, devices=jax.devices()[:TOTAL_TPU_TO_USE])
```

Tried with this mesh instead and everything works. 

I will assume that the mesh will always have the tp dimension.

tunix/generate/utils.py
wenxindongwork: good point! updated this function to work with any tensor rank.

wenxindongwork:
closed in favor of https://github.com/google/tunix/pull/202

PR #196, opened 2025-08-14
Removes the partition spec from the mappings.

wang2yn84:
We no longers uses the partition spec in the mapping to control the vllm sharding. Instead, we follow the vllm's existing sharding binded to the jax array.


lc5211 (approved):

wang2yn84:
We might still need these sharding spec. Close this PR for now.

PR #202, opened 2025-08-18
Weight sync utils: scanned model support and weight hook functions

wenxindongwork:
This PR adds weight sync logic for supporting scanned parameters and applying weight transformation hooks. 

**Summary**

*Support for scanned (layer-wise) parameters:*
- Updated `transfer_state_with_mappings` in `tunix/generate/utils.py` to properly handle scanned parameters, allowing extraction and assignment of layer-specific tensors based on sharding specifications. Helper functions were added to identify layer axes and extract layers, and the logic now supports both scanned and regular parameters. 

*Weight transformation hooks:*
- Introduced a `key_mapping_hook_fns` argument to `transfer_state_with_mappings`, enabling custom transformation functions to be applied to parameter values (e.g., scaling, normalization) after optional transpose but before assignment. 

*Test coverage:*
- Added unit tests in `utils_test.py` for transferring scanned layers, covering various scenarios including layer-wise weights, biases, and regular parameters. Existing tests were updated to use the new mock state/parameter classes and to validate the new hook and scanning logic.

### DIFF
--- a/tunix/generate/utils.py
+++ b/tunix/generate/utils.py
@@ -545,3 +545,47 @@ def transfer_state_with_mappings(
     process_entry(src_keys, src_val)
 
   return dst_state.from_flat_path(tgt_flat)
+
+
+def verify_state_closeness(golden_state, state, atol=1e-2):
+  """Check if the golden NNX state is close to the other NNX state.
+  Helper function for validating weight mapping correctness.
+  """
+  golden_state_flatten = {
+      '.'.join(str(key) for key in keys): v
+      for keys, v in golden_state.flat_state()
+  }
+
+  state_flatten = {
+      '.'.join(str(key) for key in keys): v for keys, v in state.flat_state()
+  }
+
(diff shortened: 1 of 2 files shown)


tunix/generate/utils.py
```diff
@@ -383,6 +400,9 @@ def transfer_state_with_mappings(
     key_mappings: A dictionary defining how to map keys from the source state to
       the target state. The keys of the dictionary are the source keys, and the
       values are tuples containing the target key and the sharding information.
+    key_mapping_hook_fns: A dictionary mapping keys to hook functions that
+      modify the values before assignment. The hook fn will be called after
```
lc5211, line 404: When do we need a key_mapping_hook_fns?

wenxindongwork: when source model implementation is different from vLLM, scaling or permutation for some weights is needed. This is the case for MaxText Llama3.

PR #204, opened 2025-08-21
add new grpo demo with bugs fixed

hgao327:
Added more comments and user guide.
Import and pip install bug fixed.


lc5211 (commented):
Can we merge it to the current grpo_demo.ipynb? It will be easier to see the difference and fixes added.

PR #230, opened 2025-08-27
Support for vllm and MaxText

A9isha:
Modification for supporting vllm and MaxText

### DIFF
--- a/tunix/generate/utils.py
+++ b/tunix/generate/utils.py
@@ -486,15 +486,6 @@ def transfer_state_with_mappings(
     new_value = (
         reshard_fn(value, tgt_param.value.sharding) if reshard_fn else value
     )
-    atol = 1e-2
-    if not jax.numpy.allclose(
-        tgt_param.value, new_value, atol=atol
-    ):
-      logging.debug('Weights for key %s do not match, Updating weights.', 
-                    tgt_param)
-      logging.debug(
-          'Previous value: %s, new value: %s',
-           tgt_param.value.ravel()[:10], new_value.ravel()[:10])
     tgt_param.value = new_value
 
   def _extract_layer_from_scanned_tensor(tensor, layer_idx, layer_axis):


examples/grpo_llama3_demo1.py
```diff
@@ -0,0 +1,1315 @@
...
+    )
+
+    model_config = llama3_lib.ModelConfig.llama3_1_8b()
+    tunix_model.config = model_config
+
+  return tunix_model, mesh, model_config
+
+
+
+def create_maxtext_to_vllm_mappings():
```
wenxindongwork: SInce these weight mapping logic is quite long, do you think they should exist inside MaxText.integration.tunix.tunix_adaptor.TunixMaxTextLlama instead?

I don't have strong preference over this.

examples/grpo_llama3_demo1.py
```diff
@@ -0,0 +1,1315 @@
...
+          'model.lm_head',
+          (None, 'model'),
+      ),  # checked
+      # Layer-specific mappings (scanned -> unscanned)
+      # MLP components - shard hidden dimensions for TP
+      'base.decoder.layers.mlp.wi_0.kernel': (  # checked
+          'model.layers.*.mlp.gate_proj.kernel',
+          (None, 'layer', 'model'),
+      ),  # gate_proj: (4096, 14336) - shard output
+      'base.decoder.layers.mlp.wi_1.kernel': (  # checked
```
wenxindongwork: Sorry I should have removed the `#checked` comments.

examples/grpo_llama3_demo1.py
```diff
@@ -0,0 +1,1315 @@
...
+
+
+hook_fns = {
+    'base.decoder.layers.self_attention.query.kernel': transform_query_kernel,
+    'base.decoder.layers.self_attention.key.kernel': transform_key_kernel,
+}
+
+
+# Base model
+# gemma, mesh, model_config = get_base_model(
```
wenxindongwork: Is this comment needed still?

tunix/sft/peft_trainer.py
```diff
@@ -22,6 +22,7 @@
 from absl import logging
 import flax
 from flax import nnx
+from flax import linen as nn
```
tianshub, line 25: why we need this?

tunix/sft/peft_trainer.py
```diff
@@ -353,14 +371,31 @@ def _shard_input(self, input_data: TrainingInput) -> TrainingInput:
     if mesh.empty:
       return input_data
 
+    pspec = shd.PartitionSpec(*self.config.data_sharding_axis)
+
+    def _get_sharding(x):
```
tianshub: is this maxtext model only logic? if so can we not put it here?

tunix/generate/vllm_sampler.py
```diff
@@ -48,7 +49,7 @@ class VllmConfig:
   model_version: str
   max_model_len: int
   mesh: jax.sharding.Mesh
-  hbm_utilization: float
+  hbm_utilization: int
```
tianshub: this should be float no?

examples/grpo_llama3_demo1.py
annyan09023: +1, you can follow Tunix model convention, where a method to_hf_mappings() are created, you can add that to the tunix_llama3_adaptor

tunix/generate/utils.py
```diff
@@ -383,6 +401,9 @@ def transfer_state_with_mappings(
     key_mappings: A dictionary defining how to map keys from the source state to
       the target state. The keys of the dictionary are the source keys, and the
       values are tuples containing the target key and the sharding information.
+    key_mapping_hook_fns: A dictionary mapping keys to hook functions that
```
tianshub: why you need this extra fn? why not just provide the proper key_mapping?

examples/grpo_llama3_demo1.py
```diff
@@ -0,0 +1,1315 @@
+# <a href="https://colab.research.google.com/github/google/tunix/blob/main/examples/grpo_demo.ipynb" ><img src="https://colab.research.google.com/assets/colab-badge.svg" alt="Open In Colab"/></a>
```
annyan09023: you cannot run grpo_llama3_demo1.py or maxtext_model_with_vllm notebook in Tunix w/o installing MaxText, let's not merge those 2 files in Tunix for now

tunix/generate/utils.py
```diff
@@ -359,19 +360,36 @@ def build_flat_dict(
...
+      new_flat_dict[key] = ([layer for _, layer in layers], sharding)
+
   return new_flat_dict
 
 
 def transfer_state_with_mappings(
     src_state,
     dst_state,
     key_mappings,
+    key_mapping_hook_fns=None,
```
annyan09023: can you elaborate the different usage between key_mappings and key_mapping_hook_fns, they are not clear to me

tunix/generate/utils.py
```diff
@@ -469,10 +487,113 @@ def process_entry(src_keys, src_val):
...
+
+  def process_entry(src_keys, src_val):
+    flat_key = '.'.join(str(k) for k in src_keys)
+
+    if flat_key not in new_src_dict:
+      # Skip RNG states that don't need mapping
+      if _should_skip_parameter(flat_key):
+        logging.debug('Skipping parameter: %s', flat_key)
+      else:
+        logging.error('!!! No mapping for source key: %s', flat_key)
```
annyan09023: is that a logging error or better to throw an error explictly?

tunix/generate/utils.py
```diff
@@ -469,10 +487,113 @@ def process_entry(src_keys, src_val):
...
+      for keys, v in golden_state.flat_state()
+  }
+
+  state_flatten = {
+      '.'.join(str(key) for key in keys): v for keys, v in state.flat_state()
+  }
+
+  # Check that keys match
+  if not golden_state_flatten.keys() == state_flatten.keys():
+    logging.info('Keys do not match.')
```
annyan09023: do you need to output keys in the logging?

tunix/generate/vllm_sampler.py
```diff
@@ -98,6 +100,8 @@ def __init__(
         and config.mapping_config.lora_to_hf_mappings is not None
     ):
       self.mappings |= config.mapping_config.lora_to_hf_mappings
+    
+    # self.copied_original_state = jax.tree.map(jnp.copy, self.transformer_state)
```
annyan09023: remove unused line?

tunix/generate/vllm_sampler.py
```diff
@@ -107,13 +111,18 @@ def update_params(
       filter_types: Optional[Tuple[Any, ...]] = None,
   ):
     del filter_types
+    print("update_params called...")
```
annyan09023: remove

tunix/generate/vllm_sampler.py
```diff
@@ -107,13 +111,18 @@ def update_params(
...
     utils.transfer_state_with_mappings(
         src_state=updated_weights,
         dst_state=self.transformer_state,
         key_mappings=self.mappings,
+        key_mapping_hook_fns=self.to_hf_hook_fns,
         transpose_keys=self.to_hf_transpose_keys,
         reshard_fn=reshard.reshard_pytree,
     )
+    print("done on update_params... verify the weights are updated")
```
annyan09023: remove

tunix/generate/vllm_sampler.py
```diff
@@ -164,6 +178,12 @@ def tokenize(self, input_string: str) -> List[int]:
         if input_ids[-1] != self.tokenizer.eos_id()
         else []
     )
+    # return bos_tok + input_ids + eos_tok
+    eos_tok = (
```
annyan09023: are you repeating above logic? I think line 181 can be uncommented and then then you don't need to add new logic here?

tunix/generate/vllm_sampler.py
```diff
@@ -200,7 +220,7 @@ def detokenize(
   def __call__(
       self,
       input_strings: List[str],
-      max_generation_steps: int,
+      total_generation_steps: int,
```
annyan09023: don't revert master change

tunix/sft/peft_trainer.py
```diff
@@ -312,12 +313,29 @@ def _shard_optimizer(self, mesh: shd.Mesh) -> None:
...
+        return shd.PartitionSpec(*pspec[:state_ndim])
+      return pspec
+
+    adjusted_pspecs = jax.tree.map(
+        _adjust_pspec, optimizer_pspecs, optimizer_state_arrays
+    )
+    optimizer_shardings = nn.logical_to_mesh_sharding(
+        adjusted_pspecs,
+        mesh,
+    )
```
annyan09023: not sure I understand, what problem are you trying to solve here?

annyan09023 (changes requested):

tunix/sft/peft_trainer.py
A9isha: I was using linen for the line `  optimizer_shardings = nn.logical_to_mesh_sharding` in this file (peft_trainer.py)

examples/grpo_llama3_demo1.py
A9isha: done - thanks - added it [here](https://github.com/AI-Hypercomputer/maxtext/pull/2251/files#diff-380da2b2ec0bc32e77e8ff9cd02952080a6b7ce516cd37b92092a19a6fc6b670R45), and we use to_hf_mappings()  in the script

examples/grpo_llama3_demo1.py
A9isha: Done

examples/grpo_llama3_demo1.py
A9isha: removed now

examples/grpo_llama3_demo1.py
A9isha: yes sure I am moving this file to MaxText/examples

tunix/generate/utils.py
A9isha: @wenxindongwork could you please respond here?

tunix/generate/utils.py
A9isha: @wenxindongwork could you please respond here?

tunix/generate/utils.py
A9isha: @wenxindongwork could you please respond here?

tunix/generate/vllm_sampler.py
A9isha: done

tunix/generate/vllm_sampler.py
A9isha: done

tunix/generate/vllm_sampler.py
A9isha: done

tunix/sft/peft_trainer.py
A9isha: I saw some of the state being scalar (for e.g., learning rate in the optimizer state) which is not a jax array. 
If the pspec is "too long" for the state, it's truncated to match the state's rank. For a scalar (state_ndim=0), this results in an empty PartitionSpec, P(), which correctly signifies that the scalar should be replicated across all devices. The model weights are all jax arrays, unline optimizer state.

tunix/generate/vllm_sampler.py
A9isha: done

tunix/generate/vllm_sampler.py
A9isha: ah yes - sorry - done!

tunix/generate/utils.py
A9isha: @wenxindongwork could you please respond here?

tunix/generate/vllm_sampler.py
A9isha: true, made this into float. @wenxindongwork please let me know if you see any concern

wenxindongwork: this should be a float (and less than 1)

tunix/generate/utils.py
wenxindongwork: We need sometimes to transform the weights (e.g. rescaling, permutation) because the models are implemented differently.

tunix/generate/utils.py
wenxindongwork: We need sometimes to transform the weights (e.g. rescaling, permutation) because the models are implemented differently.
If no transformation is needed for the parameter, we simply supply the key_mapping. Otherwise, we will need to provide a hook fn for that param.

tunix/generate/utils.py
wenxindongwork: I think it's okay to not throw an error here, because some parameters are skipped (e.g. rng) But agreed that this can be made less error prone.

tunix/generate/utils.py
wenxindongwork: That's a good idea! Let's include the key name in the mapping. @A9isha should I make this change?

A9isha: go for it. Could you please push to this branch `maxtext-vllm-integration`?

tunix/generate/utils.py
A9isha: Could you please push the change to this branch `maxtext-vllm-integration`?

tunix/sft/peft_trainer.py
A9isha: Yes, in MaxText we have [logical axes names](https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/configs/base.yml#L321-L322) mapping to physical mesh axes. So we need this otherwise we get an error saying cannot find [vocab](https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/configs/base.yml#L354) among [['data', 'stage', 'fsdp', 'fsdp_transpose', 'sequence', 'context', 'context_autoregressive', 'tensor', 'tensor_transpose', 'tensor_sequence', 'expert', 'autoregressive']](https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/configs/base.yml#L321). Where would you recommend putting it?

tianshub: let's put it to a separate sharding_util.py file for now so the only appearance in this file is one line function call. Btw, is this copy pasting some existing maxtext code? if so we should try to dedup it. One option is to check if the model provides a `_get_sharding ` function and call that when available.

IIUC, Rich is trying to simplify the sharding config in maxtext, and my ask is until that's settled, let's try to control the blast radius of replicating code.

tunix/sft/peft_trainer.py
tianshub: see the comments below, the function doesn't depend on any local var, so we can move it out.

tunix/generate/utils.py
tianshub: got it, saw the explanation in the thread.

tunix/generate/utils.py
wenxindongwork: I'm resubmitting the PR for the weight sync utils. (It got rolled back) 
https://github.com/google/tunix/pull/231

Could you rebase after it's merged?

tunix/generate/utils.py
wenxindongwork: Making this change in https://github.com/google/tunix/pull/231

tunix/generate/utils.py
```diff
@@ -486,6 +486,13 @@ def _process_and_assign_value(value, tgt_param, flat_key, src_keys):
     new_value = (
         reshard_fn(value, tgt_param.value.sharding) if reshard_fn else value
     )
+    atol = 1e-2
+    if not jax.numpy.allclose(
+        tgt_param.value, new_value, atol=atol
```
wenxindongwork: what is the purpose of this code?

A9isha: this was to check if weights are actually getting updated after a train step. I put the logging under `logging.debug`

annyan09023: @A9isha this should be guarded through a unit test instead of some logging, can you remove the logging here and we can add a proper next in next PR?

tunix/sft/peft_trainer.py
A9isha: `optimizer_shardings = nn.logical_to_mesh_sharding` is in the `_shard_optimizer()`. The comments below are for `_get_sharding()` inside `_shard_input()`

tunix/sft/peft_trainer.py
A9isha: I have added `tunix/sft/sharding_utils.py`

This is not a copy of code from MaxText though

tunix/generate/utils.py
A9isha: makes sense - done

tunix/generate/utils.py
A9isha: thanks for rebasing WenXin

PR #231, opened 2025-08-27
[Resubmit] Weight sync utils: scanned model support and weight hook functions

wenxindongwork:
Resubmitting https://github.com/google/tunix/pull/202 as it was rolled back. 

This PR adds weight sync logic for supporting scanned parameters and applying weight transformation hooks.

Summary

Support for scanned (layer-wise) parameters:

Updated transfer_state_with_mappings in tunix/generate/utils.py to properly handle scanned parameters, allowing extraction and assignment of layer-specific tensors based on sharding specifications. Helper functions were added to identify layer axes and extract layers, and the logic now supports both scanned and regular parameters.
Weight transformation hooks:

Introduced a key_mapping_hook_fns argument to transfer_state_with_mappings, enabling custom transformation functions to be applied to parameter values (e.g., scaling, normalization) after optional transpose but before assignment.
Test coverage:

Added unit tests in utils_test.py for transferring scanned layers, covering various scenarios including layer-wise weights, biases, and regular parameters. Existing tests were updated to use the new mock state/parameter classes and to validate the new hook and scanning logic.


lc5211 (approved):
Thank you!

PR #277, opened 2025-09-10
Refactor state transfer for vllm rollout.

lc5211:
Refactor state transfer for vllm rollout.
- Refactor the logic to handle scanned layers by unrolling them before processing. 
- Refactor the reshard logic to batched manner as recommended.

### DIFF
--- a/tunix/generate/utils.py
+++ b/tunix/generate/utils.py
@@ -363,12 +363,13 @@ def build_flat_dict(
         # Check if this is a scanned parameter (has 'layer' in sharding spec)
         if sharding and 'layer' in sharding:
           if actual_src not in new_flat_dict:
-            new_flat_dict[actual_src] = ([], sharding)
+            new_flat_dict[actual_src] = ([], [], sharding)
           layer_number = int(matched.groups()[0])
-          new_flat_dict[actual_src][0].append((layer_number, v, path))
+          new_flat_dict[actual_src][0].append((layer_number, v))
+          new_flat_dict[actual_src][1].append((layer_number, path))
         else:
           # Regular (non-scanned) parameter
-          new_flat_dict[actual_src] = (v, path), sharding
+          new_flat_dict[actual_src] = v, path, sharding
 
         mapped = True
         break
@@ -377,14 +378,13 @@ def build_flat_dict(
       logging.warning('!!! No mapping for flat state: %s', path)
 
   # Sort layers
-  for key, (layers, sharding) in new_flat_dict.items():
+  for key, (layers, paths, sharding) in new_flat_dict.items():
     if isinstance(layers, list):
       layers.sort(key=lambda x: x[0])
-      values = [v for _, v, _ in layers]
-      paths = [p for _, _, p in layers]
+      paths.sort(key=lambda x: x[0])
+      values = [v for _, v in layers]
+      paths = [p for _, p in paths]
       new_flat_dict[key] = (values, paths, sharding)
-    else:
-      new_flat_dict[key] = layers + (sharding,)
 
   return new_flat_dict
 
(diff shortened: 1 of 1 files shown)


tunix/generate/utils.py
```diff
@@ -410,141 +415,122 @@ def transfer_state_with_mappings(
...
+      else:
+        logging.error('!!! No mapping for source key: %s', src_key)
+      continue
+
+    tgt_param, _, sharding_spec = src_to_tgt_map[src_key]
 
-  def _process_and_assign_value(value, tgt_param, flat_key, src_keys):
-    # Optional transpose
+    def _get_layer_axis_from_sharding_spec(sharding_spec):
+      if isinstance(sharding_spec, (list, tuple)):
```
annyan09023, line 442: just curious, why sharding_spec can be either list or tuple?

tunix/generate/utils.py
```diff
@@ -410,141 +415,122 @@ def transfer_state_with_mappings(
...
+        for i, spec in enumerate(sharding_spec):
+          if spec == 'layer':
+            return i
+      return None
+
+    layer_axis = _get_layer_axis_from_sharding_spec(sharding_spec)
+    if layer_axis is not None:
+      num_layers = src_val.value.shape[layer_axis]
+      # values = []
```
annyan09023: remove

tunix/generate/utils.py
```diff
@@ -410,141 +415,122 @@ def transfer_state_with_mappings(
   Returns:
     The target state with the transferred values.
   """
+  tgt_flat_list = dst_state.flat_state()
+  sharding_dict = None
+  if reshard_fn:
```
annyan09023, line 420: you have if reshard_fn logic in L522, maybe move L421 in the if check of L522 as well to move relevant logic close

tunix/generate/utils.py
```diff
@@ -410,141 +415,122 @@ def transfer_state_with_mappings(
...
+    )
 
-  src_flat = src_state.flat_state()
-  tgt_flat = dst_state.flat_state()
   # Maps source keys to target tensor(s) and sharding spec
-  new_src_dict = build_flat_dict(tgt_flat, key_mappings)
+  src_to_tgt_map = build_flat_dict(tgt_flat_list, key_mappings)
+
+  # Flatten the source state, unrolling any scanned layers.
+  unscanned_src_to_tgt_flat = {}
```
wenxindongwork, line 429: One performance issue I see with unscanning the layer all together is duplicating the HBM usage. Since JAX slicing creates a new array instead of reference handle. 

Is this something you think we need to worry about in this stage?

annyan09023 (commented):
I can only try my best to understand the logic, it has a lot of nuances ;-(

tunix/generate/utils.py
lc5211: I need to retrieve the original sharding from tgt_param.value before it's updated in L516.

tunix/generate/utils.py
lc5211: Current implementation also does the slice to a new array right?
Besides, current implementation is not right by operating on the src_val.value with jnp.transpose, jnp.reshape or astype. It will change the trainer's weights in place which will cause training unstable, e.g. we observed maxtext asks trainer weights to be f32 but after the first weight sync, it's converted to bf16.

tunix/generate/utils.py
lc5211: This is how we defined in to_hf_mappings. I assume list or tuple are the acceptable formats. @wenxindongwork  do you have more context on the convention? Thanks!

tunix/generate/utils.py
wenxindongwork: thanks for the explanation!

PR #288, opened 2025-09-11
Add an image and build script to Tunix

annyan09023:
To test with image:
```
sudo docker run -it --rm --net=host --ipc=host --ulimit memlock=-1:-1 -v "$(pwd)":/app --workdir /app --device=/dev/vfio/0       
   --device=/dev/vfio/1 --device=/dev/vfio/2 --device=/dev/vfio/3 --device=/dev/vfio/4 --device=/dev/vfio/5 --device=/dev/vfio/6        │
   --device=/dev/vfio/7 --device=/dev/vfio/vfio tunix_vllm_image bash
```
The scripts/grpo_demo_llama3_qwen2.py is not runnable now because the dataset is not downloaded by the script, instead, it expects to copy to a gcs bucket, I will fix that part

### DIFF
--- a/docker_build_dependency_image.sh
+++ b/docker_build_dependency_image.sh
@@ -1,0 +1,90 @@
+#!/bin/bash
+
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


wang2yn84:
Hi [annyan09023](https://github.com/annyan09023), I think it's easier to build on top of vllm image: vllm/vllm-tpu:v0.11.1, together with tpu-inference==0.11.1

PR #336, opened 2025-09-18
Adding basic tpu github tests

mydatascience:
Adding github tests on tpu machines

Using current tests to run pytest on tpu machines.

Tests need to be refactored after fixes apply. (we need to run all tests after code change)

### DIFF
--- a/.github/workflows/tpu-tests.yml
+++ b/.github/workflows/tpu-tests.yml
@@ -1,0 +1,164 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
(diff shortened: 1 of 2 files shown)


.github/workflows/tpu-tests.yml
```diff
@@ -0,0 +1,207 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: TPU Tests
+
+on:
```
wang2yn84, line 20: Consider call this script from the entrypoint here: https://github.com/google/tunix/blob/688c395c5737a054774152777d3e6e03b636126e/.github/workflows/build_and_test_tunix.yml#L44?

.github/workflows/tpu-tests.yml
```diff
@@ -0,0 +1,207 @@
...
+        pip install -e .
+        pip install pytest pytest-xdist
+    
+    - name: Verify TPU availability
+      run: |
+        python -c "
+        import jax
+        print(f'JAX version: {jax.__version__}')
+        print(f'JAX devices: {jax.devices()}')
+        print(f'TPU available: {len(jax.devices()) > 0}')
```
wang2yn84: If it's running on CPU, this will pass too. Consider check jax.devices()[0].platform == 'tpu'

.github/workflows/tpu-tests.yml
mydatascience: Done

.github/workflows/tpu-tests.yml
mydatascience: Done

wang2yn84 (approved):
Thank you very much!

PR #348, opened 2025-09-22
Fix ReadTheDocs build

melissawm:
The RTD build for latest is currently failing since https://github.com/google/tunix/pull/344 due to the minimum required Python version. This PR bumps the RTD yaml file to also use Python 3.12.

### DIFF
--- a/.github/workflows/run_notebooks.yml
+++ b/.github/workflows/run_notebooks.yml
@@ -1,0 +1,46 @@
+name: Run Notebooks
+
+on:
+  push:
+    branches: [ main ]
+  workflow_dispatch:
+
+jobs:
+  run-notebooks:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/run_notebooks.yml
```diff
@@ -0,0 +1,46 @@
...
+jobs:
+  run-notebooks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
```
annyan09023: should be 3.12 as well?

melissawm: This actually should not be here, apologies. I was experimenting locally but this workflow is still not working. I'll remove from the PR.

PR #367, opened 2025-09-24
properly merge batches with arbitrarily nested data structure

Hanjun-Dai:
When merging multiple micro batches into one, the current code assumes a specific form of data structure that allows concatenation. However when the batch elements are nested dicts (it is very common in many custom prepared datasets), the current code will not be able to handle that.

Since the package already depends on jax, we can directly use the jax's tree map and np concat, so as to handle the arbitrary data structure when merging batches.

**Reference**

https://docs.jax.dev/en/latest/_autosummary/jax.tree_util.tree_map.html

**Checklist**

### DIFF
--- a/tunix/rl/utils.py
+++ b/tunix/rl/utils.py
@@ -25,6 +25,7 @@ from absl import logging
 from flax import nnx
 from flax.nnx import filterlib
 from flax.nnx import statelib
+from grain._src.core import tree_lib
 import humanize
 import jax
 import jax.numpy as jnp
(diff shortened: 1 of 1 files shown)


tunix/rl/utils.py
```diff
@@ -240,7 +241,7 @@ def merge_micro_batches(batches: List[dict[str, Any]]) -> dict[str, Any]:
     if isinstance(all_values[0], list):
       merged[key] = list(chain.from_iterable(all_values))
     else:
-      merged[key] = np.concatenate([np.asarray(v) for v in all_values], axis=0)
+      merged[key] = tree_util.tree_map(lambda *xs: np.concatenate(xs), *all_values)
```
tianshub: thank you @Hanjun-Dai . A minor issue is this failed the pylint formatter internally. I think you can do `pre-commit install` on your side and push again, that should auto format the change.

tianshub: it's a minor line too long issue:
```
merged[key] = tree_util.tree_map(
          lambda *xs: np.concatenate(xs), *all_values
      )
```

kangaroo0123: thank you so much! i did the pre-commit thing and pushed the re-formatted code.

PR #417, opened 2025-09-29
Fix SFT tests. [PR3/N to add all TPU tests to CI] 

wang2yn84:
This PR fixes the following:
1. When logging learning rate, test report type error: flax.nnx.training.optimizer.OptArray, expecting np/jnp array: Here are some details of the failure: https://paste.googleplex.com/6277703556792320.
2. After fixing #1, learning rate becomes unavailable when updating pbar for gradient accumulation tests. The root cause should be the OptArray object lost reference because we didn't keep track of it in the metrics any more. There is a delay when the metrics are logged and update to pbar. Fix by moving learning rate data from TPU to CPU when buffering. 
3. Fix the temp folder in the test. They all assume user will provide the temp path and therefore pytest fails. Provide another option to set it with test temp dir.
4. Refactor the test to share some common variables, such as mesh, etc.

There are 3 DPO tests still failing. In order to get the changes in before release, bypass these 3 tests in CI for now. Should fix these 3 tests asap.

**Reference**

**Colab Notebook**

**Checklist**

- [ ] I have followed [Contribution Guidelines](https://github.com/google/tunix/blob/main/CONTRIBUTING.md).


abheesht17:
Looking into the DPO test errors. I can only reproduce two test failures though

abheesht17:
> Looking into the DPO test errors. I can only reproduce two test failures though

Discussed offline; DPO tests pass. Created this dummy PR to verify: https://github.com/google/tunix/pull/418.

.github/workflows/tpu-tests.yml
```diff
@@ -76,33 +76,12 @@ jobs:
           --ignore=tests/generate/vllm_sampler_test.py \
           --ignore=tests/generate/tokenizer_adapter_test.py
 
-    - name: Run tunix SFT tests (PASSED only)
+    - name: Run tunix SFT tests (Exclude untriaged failing tests for now)
```
abheesht17: Down below, I can see that we run RL tests only for common.py. Any reason why we do this, and don't run all the tests? Do none of the other tests pass?

abheesht17 (commented):
Thanks! Left a comment

.github/workflows/tpu-tests.yml
wang2yn84: I'm fixing the folders one by one cuz each folders takes quite significant time. That's why I'm marking this PR as PR3/N :p. The rl folder has most failures and that's my last goal

wang2yn84:
> > Looking into the DPO test errors. I can only reproduce two test failures though
> 
> Discussed offline; DPO tests pass. Created this dummy PR to verify: #418.

The errors I was seeing was quite different. Most likely due to stale pytest cache. All 3 tests are passing now after some minor precision fix.

PR #493, opened 2025-10-03
Fix the OOM issue for FSDP

wang2yn84:
When we enable full FSDP (8, 1) in grpo example (need to increase batch size accordingly), get the following error: 

```
 jaxlib._jax.XlaRuntimeError: RESOURCE_EXHAUSTED: Error loading program 'jit__decode_fn': Attempting to reserve 10.54G at the bottom of memory. That was not possible. There are 10.37G free, 0B reserved, and 10.37G reservable.: while running replica 0 and partition 0 of a replicated computation (other replicas may have failed as well).
```

Prefill doesn't OOM while decode does, it points to the while loop in the decode function. After investigation, the root cause is the parameters are not explicitly listed in the carry. XLA fused the parameters to the generated graph. It's still a speculation and need to confirm with XLA team. For some reason, TP doesn't expose this issue. Verified that FSDP fully shards the model by loading Gemma3 27B model on 2x4 vlp on Borg without issue. Still need further investigation on this.

**Reference**

**Colab Notebook**

**Checklist**

- [ ] I have followed [Contribution Guidelines](https://github.com/google/tunix/blob/main/CONTRIBUTING.md).


hgao327 (approved):

wang2yn84:
Cannot reproduce this issue any more. b/454740726

Issue #137, opened 2025-07-27
Wandb


shivajid:
## Expected Behavior
Clean execution

## Actual Behavior
Adding the metrics logger throws wandb errors.

wandb: WARNING Tried to log to step 0 that is less than the current step 3. Steps must be monotonically increasing, so this data will be ignored. See https://wandb.me/define-metric to log data out of order.

Later has some some unexpected behavior

```
  File "/home/shivajid/workshop/tensor_crunch/sample_trainer.py", line 223, in <module>
    out_data = sampler(
               ^^^^^^^^
  File "/home/shivajid/tunixenv/lib/python3.11/site-packages/tunix/generate/sampler.py", line 705, in __call__
    sampling_state = self._compiled_prefill_fn(
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/shivajid/tunixenv/lib/python3.11/site-packages/jax/_src/monitoring.py", line 97, in record_scalar
    callback(event, value, **kwargs)
  File "/home/shivajid/tunixenv/lib/python3.11/site-packages/tunix/sft/metrics_logger.py", line 90, in log_to_wandb
    wandb.log({event: scalar_value}, step=current_step)
  File "/home/shivajid/tunixenv/lib/python3.11/site-packages/wandb/sdk/lib/preinit.py", line 36, in preinit_wrapper
    raise wandb.Error(f"You must call wandb.init() before {name}()")
wandb.errors.errors.Error: You must call wandb.init() before wandb.log()
```

## Steps to Reproduce the Problem

1. Add wandb to your import and login
1.
1.

## Specifications

- Version:
- Platform: TPUv6e-8

tianshub:
Did you run wanb.init() before the rest of the execution?

PR #517, opened 2025-10-07
Fix wandb integration issues in MetricsLogger

pujariaditya:
This PR fixes wandb integration issues in MetricsLogger reported in #137:

1. **Prevents wandb re-initialization**: Added check to ensure MetricsLogger doesn't initialize wandb if a run is already active, fixing "You must call wandb.init() before wandb.log()" error when users manually initialize wandb.

2. **Ensures monotonic step tracking**: Implemented global step counter `_last_steps` that tracks steps per metric prefix, preventing "Tried to log to step 0 that is less than the current step 3" warnings.

3. **Conditional wandb.finish()**: Only finishes wandb runs if they were started by MetricsLogger, avoiding interference with user-managed sessions.

**Reference**
- https://github.com/google/tunix/issues/137

**Checklist**

### DIFF
--- a/tunix/sft/metrics_logger.py
+++ b/tunix/sft/metrics_logger.py
@@ -6,6 +6,7 @@ import datetime
 import enum
 import functools
 import logging
+from typing import Dict
 
 import jax
 import numpy as np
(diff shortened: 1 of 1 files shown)


darsnack:
It would be great if these changes could be merged.

PR #557, opened 2025-10-11
Added Dr. GRPO and Magistral improvements

siddagra:
This PR adds several new configuration options to the GRPO learner, enabling support for recent algorithmic improvements from the literature:

- **epsilon_high**: Defaults to Epsilon. Epsilon value for clipping upper bound. If None, defaults to epsilon for symmetric clipping. Set higher than epsilon for asymmetric clipping.  
  *Clip Higher introduced in DAPO ([arxiv:2503.14476](https://arxiv.org/pdf/2503.14476)), also used in Magistral and VAPO ([arxiv:2504.05118](https://arxiv.org/abs/2504.05118)).*

- **dr_grpo**: Default False. If True, use DR-GRPO variant which:
  - Only subtracts mean rewards (no std dev normalization)
  - Normalizes loss by max_tokens instead of actual token counts  
  *See DR-GRPO ([arxiv:2503.20783](https://arxiv.org/pdf/2503.20783)).*

- **max_tokens**: Only used if Dr-GRPO is True. Maximum generation tokens for DR-GRPO normalization. If None and dr_grpo is True, will attempt to infer from rollout config.

- **magistral_adv_norm**: Default False. If True, normalize advantages by mini-batch std dev instead of group std dev after subtracting group mean.  
  *See Magistral ([arxiv:2506.10910](https://arxiv.org/pdf/2506.10910)).*

- **eliminate_non_diverse_groups**: Default False. If True,  groups where all rewards are identical are masked instead, as they provide no/noisey learning signal. Done this by setting loss for this group to 0 via masking.  
  *Introduced in Magistral ([arxiv:2506.10910](https://arxiv.org/pdf/2506.10910)).*

- **debug**: Default False. If True, print detailed debug information during training.

Most of these improvements are supported both theoretically and empirically in the respective papers.

All of these features are kept False by default in order to mantain original functionality unless the user decides to enable some feature.

The reason for keeping these terms independent is twofold:
1. To enable ablation studies with certain features on/off.
2. To allow combining multiple theoretically sound corrections from all of these papers under the same framework.

**Reference**
- [GRPO](https://arxiv.org/abs/2402.03300)
- [GSPO](https://www.arxiv.org/pdf/2507.18071)
- [Magistral](https://arxiv.org/pdf/2506.10910)
- [DR-GRPO](https://arxiv.org/pdf/2503.20783)
- [DAPO (Clip Higher)](https://arxiv.org/pdf/2503.14476)
- [VAPO](https://arxiv.org/abs/2504.05118)

**Colab Notebook**

**Checklist**

- [Y] I have verified that my change does not break existing code and all unit tests pass.
- [Y] I have added all appropriate doc-strings/documentation.
- [Y] My PR is based on the latest changes of the main branch (if unsure, rebase the code).
- [Y] I have signed the [Contributor License Agreement](https://cla.developers.google.com/about).
- [Y] I have followed [Contribution Guidelines](https://github.com/google/tunix/blob/main/CONTRIBUTING.md).

Let me know if you want to further customize or add more details!
Also let me know how I should add unit tests more concretely. I conduct unit tests in a main function but I assume that is not fine for this.


wang2yn84:
Thank you [siddagra](https://github.com/siddagra) for your PR! These are really good impls for popular algos. While we are reviewing the algo, can you add them to our notebook and also include the results in this PR? Thanks a lot!

siddagra:
Thanks for the kind words!

Sure I can share a notebook soon. I wasn't able to get any major "improvements" since the demo notebook does not even improve the evaluation metric if I run it "as is" for me, and I also get a very bad KL spike past 1000 steps with all approaches (including vanilla unmodified code) with gemma-2b-it. But I have run all of the approaches without any issues and have bug tested it quite thoroughly. I will share the notebook and results with you regardless though.

Btw I also understand if the implementations seem scuffed by your standards, I am dedicated to improving them to fit the overall direction you would like to go with this. Thanks!

wang2yn84:
Awesome! That's great you've run through everything. Can't wait to see your notebook.

We are working on the algorithm abstraction on our end. @sizhit2 do you have an ETA for that? If it takes long, we should land this PR first.

tunix/rl/grpo/grpo_helpers.py
```diff
@@ -13,22 +13,75 @@
...
 
 
-def compute_advantages(rewards: jax.Array, num_generations: int) -> jax.Array:
+def compute_advantages(
+    rewards: jax.Array,
+    num_generations: int,
+    dr_grpo: bool = False,
+    magistral_adv_norm: bool = False,
+    eliminate_non_diverse_groups: bool = False,
+    debug: bool = False,
```
sizhit2, line 29: Instead using debug, let's use logging.vlog() to log debug information.

from absl import logging
logging.vlog(2,"debug information")

tunix/rl/grpo/grpo_learner.py
```diff
@@ -79,6 +113,28 @@ def __post_init__(self):
...
+      print(
+          "[GRPO Config] Magistral Adv Norm: enabled={self.magistral_adv_norm}"
+      )
+      print(
+          "[GRPO Config] Eliminate Non-Diverse Groups:"
+          f" enabled={self.eliminate_non_diverse_groups}"
+      )
 
 
 class GRPOLearner(rl_learner.RLLearner):
```
sizhit2, line 140: Let's add unit test for your changes.

sizhit2:
Thanks for adding the improvements. Just checking in on the status of this—are you still planning to move forward with the merge? No worries if you've moved on to other things, we can just close this for now if that's the case.

siddagra:
Hey, here is the notebook I had used:

[grpo_demo (2).ipynb](https://github.com/user-attachments/files/23353641/grpo_demo.2.ipynb)

These are some wandb results:
image
(I had changed various hyperparameters and hence it may not be enough for comparison, but at least it showcases that all variants/combination of variants work without error for long runs)

Let me know if I should do anything else!

siddagra:
I also merged the modified code with pad_id/eos_id, please let me know if its fine.

wang2yn84:
> I also merged the modified code with pad_id/eos_id, please let me know if its fine.

Hi @siddagra , thank you very much for the metrics! Looks decent. Unit tests are important for users and developers too, can you help add them for the parts you changed? Thank you!

PR #760, opened 2025-11-18
reduce safetensor loading time

keshavb96:
This PR speeds up the loading of safetensors into the target sharded pytree of JAX arrays by mmaping the entire safetensor at once and then using indexing to extract the relevant tensor data. The speedup is a little over 2x compared to main.

### DIFF
--- a/tunix/models/safetensors_loader.py
+++ b/tunix/models/safetensors_loader.py
@@ -14,17 +14,19 @@
 
 """Common utilities for loading model weights from safetensors files."""
 
-import concurrent.futures
 import contextlib
 import os
 import re
-import threading
+import mmap
+import numpy as np
+import json
+import struct
 
 from etils import epath
 from flax import nnx
 import jax
 import jax.numpy as jnp
-import safetensors.flax as safetensors
+from ml_dtypes import bfloat16
 # DO NOT CHNAGE THIS IMPORT. This is used in both oss and GOOGLE_INTERNAL_PACKAGE_PATH.
 from tunix.oss import utils
 from tunix.utils import compat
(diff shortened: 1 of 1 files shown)


tunix/models/safetensors_loader.py
```diff
@@ -59,6 +61,51 @@ def path_to_key(path):
...
+
+    file_size = os.path.getsize(filepath)
+    data_size_bytes = file_size - data_block_start_offset_bytes
+    total_bfloat16_elements = data_size_bytes // 2
+
+    f = open(filepath, 'rb')
+    mm = mmap.mmap(f.fileno(), 0, prot=mmap.PROT_READ)
+    contiguous_array = np.frombuffer(
+        mm,
+        dtype=bfloat16,
```
wang2yn84: Are we assuming it's bf16 here?

keshavb96: Hi @wang2yn84! I've generalized it to dtypes plus also confirmed that the safetensors loading test: `tunix/tests/models/safetensors_loader_test.py` passes.

As for the benchmark, on loading llama3.1-8b, the previous version of loading takes ~15 seconds without any sharding and using a single device, and ~20 seconds with sharding and multiple devices. This PR reduces those times to ~8 seconds and ~10 seconds respectively. I use the following code for the benchmark:

```
from tunix.models.llama3 import params
from tunix.models.llama3 import model as model_lib
import time
from jax.sharding import Mesh
import numpy as np
import jax

config = model_lib.ModelConfig.llama3_1_8b()
mesh = Mesh(np.array(jax.devices()).reshape(8, 1), ('tp', 'fsdp'))

start_time = time.perf_counter()
model = params.create_model_from_safe_tensors("/tmp/models/llama3-8b", config, mesh)
duration = time.perf_counter() - start_time
print(f"[safetensors_loader] Loaded model in {duration:.3f} seconds")
```

wang2yn84: That's awesome! Can you squash the commits into 1 before I trigger the merging process on my end?

PR #769, opened 2025-11-19
Fix `TypeError` in example notebooks by replacing `MESH[0]` with `len(MESH[0])`.

rajasekharporeddy:
Currently the notebooks raises a TypeError:

```python
TypeError: can't multiply sequence by non-int of type 'tuple'
```
 due to the definition of mesh:

```python
mesh = jax.make_mesh(*MESH, axis_types=(jax.sharding.AxisType.Auto,) * MESH[0])
```
This PR resolves this issue by replacing `MESH[0]` with `len(MESH[0])`.

### DIFF
--- a/examples/logit_distillation.ipynb
+++ b/examples/logit_distillation.ipynb
@@ -19,10 +19,10 @@
       ]
     },
     {
+      "cell_type": "markdown",
       "metadata": {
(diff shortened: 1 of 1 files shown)


wang2yn84:
Thank you so much!

PR #786, opened 2025-11-21
feat:add device indexes for sglang jax

pathfinder-pf:
Resolves #\<issue_number_goes_here\>
For the sglang jax rollout engine, construct a mesh through device indices.

**Reference**

**Colab Notebook**

**Checklist**

### DIFF
--- a/scripts/grpo_demo_sglang_jax_rollout.py
+++ b/scripts/grpo_demo_sglang_jax_rollout.py
@@ -52,6 +52,15 @@ parser.add_argument(
     required=False,
     help="The model version to use.",
 )
+
+parser.add_argument(
+    "--rollout-data-parallel-size",
+    type=int,
+    default=1,
+    required=False,
+    help="Rollout engine data parallel size.",
+)
+
 args = parser.parse_args()
 
 # ====== Data ======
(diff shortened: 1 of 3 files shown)


scripts/grpo_demo_sglang_jax_rollout.py
```diff
@@ -405,7 +405,12 @@ def load_model(model_version: str, enable_lora: bool = False):
   if "Qwen" in model_version:
     mesh_shape = (1, 2)  # because the num_key_value_heads is 2
   axis_names = ("fsdp", "tp")
-  mesh = jax.make_mesh(mesh_shape, axis_names, devices=jax.devices(), axis_types=(jax.sharding.AxisType.Auto,) * len(mesh_shape))
+  mesh = jax.make_mesh(
+      mesh_shape,
+      axis_names,
+      devices=jax.devices(),
```
wang2yn84: I think you need a subset of the device here because for smaller model you are not using all the devices.

scripts/grpo_demo_sglang_jax_rollout.py
```diff
@@ -421,7 +426,11 @@ def load_model(model_version: str, enable_lora: bool = False):
 ref_model, mesh, model_config = load_model(repo_id)
 show_hbm_usage()
 # Policy model
-
+device_indexes = None
```
wang2yn84: Please follow how we do for vllm sampler and tpu inference. The impl here doesn't guaranteed to work.

scripts/grpo_demo_sglang_jax_rollout.py
pathfinder-pf: Thanks I will correct this.

scripts/grpo_demo_sglang_jax_rollout.py
pathfinder-pf: ok，i will have a look at vllm sampler.

PR #834, opened 2025-12-03
Enable multi-process/multi-controller support for GPUs

### LINKED ISSUE #698: GRPO training fails in multi-process mode
**Expected Behavior**
GRPO training with Tunix in multi-process SPMD (e.g., 2 GPU nodes) should generate rollouts and compute advantages without failures.

**Actual Behavior**
Crash during rollout sampling when iterating a distributed `jax.Array`:
- `AssertionError` at `jax/_src/array.py:380`: `assert self.is_fully_replicated or self.is_fully_addressable`
- Triggered in `tunix/generate/sampler.py` while doing `zip(out_tokens, lengths)`

**Steps to Reproduce the Problem**
1. Add `jax.distributed.initialize()` at the beginning of the GRPO example script.
2. Run the GRPO demo
```
srun -u --label --ntasks=2 --ntasks-per-node=1 -c${SLURM_CPUS_ON_NODE} python <grpo-script>
```

**Checklist**
- [x] I have searched the existing issues for a similar bug report.
- [x] I have provided all the required information in the "Environment" section.
- [x] I have provided a minimal, reproducible example.

**Would you like to help us fix it?**

yhtang:
Supports #705 and #706
Fixes #698 

**Overview**

This PR enables Tunix GRPO training to run in multi-controller mode on GPUs, where multiple processes join the same `jax.distributed.initialize` session with each controlling distinct GPUs.

**Background**

Running multi-controller on GPUs (e.g., 2 processes × N GPU each) previously failed with:
- `AssertionError` when iterating over non-fully-addressable arrays
- `ValueError: device_put's first argument must be a fully addressable array`

**Changes**

1. `tunix/generate/sampler.py`
  - Decode strings from local shards (`addressable_shards`) instead of iterating over global arrays
  - Maintains DP sharding: each process decodes only its local batch slice
  - Fallback to `device_get` for single-controller/fully-addressable cases
2. `tunix/sft/peft_trainer.py`
  - Updated `_shard_input` to handle global JAX arrays from rollout
  - Use `jax.device_put(x, target_sharding)` for cross-mesh resharding instead of `make_array_from_process_local_data`
  - Preserves existing behavior for NumPy/local arrays

**Status**

This PR should be considered as a proof-of-concept that demonstrates multi-controller GRPO on GPU. Please provide any feedback and comments on how it interacts with the existing workflow for TPU/Pathways.

Further optimization is needed for efficient multi-GPU scaling, as my preliminary experiments indicates near-flat strong scaling timing.

**Testing**

The following gist provides a minimal GRPO pipeline and associated run script for testing multi-process GPU runs: https://gist.github.com/yhtang/95ed9cbb6068a545aef2976fd73aab64

**Reference**
#705 
#706 

**Checklist**

### DIFF
--- a/tunix/generate/sampler.py
+++ b/tunix/generate/sampler.py
@@ -29,15 +29,33 @@ from flax.nnx import graph
 from flax.nnx import statelib
 import jax
 import jax.numpy as jnp
+from jax.experimental import multihost_utils as mhu
+from jax.lax import with_sharding_constraint as wsc
+from jax.sharding import PartitionSpec as P
 import jaxtyping
 import numpy as np
 from tunix.generate import base_sampler
 from tunix.generate import utils
 import tunix.generate.beam_search as beam_search_lib
 import tunix.generate.tokenizer_adapter as tok_adapter
+from tunix.sft import sharding_utils as shd_utils
+from typing import Any as _TypingAny  # for helper typing isolation
 
 LayerCache = dict[str, jaxtyping.Array]
 Cache = dict[str, LayerCache]
+
+
+def _assert_tp_replicated(arr: _TypingAny) -> None:
+  """Assert that a global array is replicated across tensor-parallel axis.
+
+  This checks that the sharding spec does not include the 'tp' axis.
+  If the array does not have a NamedSharding/spec, the check is skipped.
+  """
+  sh = getattr(arr, 'sharding', None)
+  spec = getattr(sh, 'spec', None)
+  if spec is None:
+    return
+  assert 'tp' not in spec, f'Expected TP-replicated array; got spec={spec}'
 
 
 @flax.struct.dataclass
@@ -508,6 +526,8 @@ class Sampler(base_sampler.BaseSampler):
         sampler_state.cache,
         attention_mask,
     )
+    # Ensure full-vocab logits are replicated across tp for sampling/beam.
+    logits = wsc(logits, P('fsdp', None, None))
     token_buffer = sampler_state.token_buffer
     done = sampler_state.done
     positions = sampler_state.positions
@@ -608,6 +628,8 @@ class Sampler(base_sampler.BaseSampler):
         sampler_state.cache,
         attention_mask,
     )
+    # Ensure full-vocab logits are replicated across tp for sampling.
+    logits = wsc(logits, P('fsdp', None, None))
     updated_sampler_state = self._sample(
         logits=logits,
         cache=cache,
@@ -696,8 +718,11 @@ class Sampler(base_sampler.BaseSampler):
 
     tokens = [self.tokenize(x) for x in input_strings]
     max_tokens_length = max(len(x) for x in tokens)
-    if max_prompt_length is None or max_prompt_length < max_tokens_length:
-      max_prompt_length = utils.next_power_of_2(max_tokens_length)
+    # Compute a global max across processes and choose a global prompt length.
+    local_max = np.array([max_tokens_length], dtype=np.int32)
+    max_prompt_len_global = int(mhu.process_allgather(local_max).max())
+    if max_prompt_length is None or max_prompt_length < max_prompt_len_global:
+      max_prompt_length = utils.next_power_of_2(max_prompt_len_global)
 
     all_input_ids = jnp.array([
         utils.pad_to_length(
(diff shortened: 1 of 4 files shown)


yhtang:
One notable limitation of this PR is the assumption of a particular mesh shape.

tunix/rl/grpo/grpo_learner.py
```diff
@@ -214,6 +216,22 @@ def _generate_and_compute_advantage(
     prompt_ids = rollout_output.left_padded_prompt_tokens
     completion_text = rollout_output.text
 
+    # Slice prompts and other inputs to match local completion count (multi-controller safe)
+    local_batch_size = len(completion_text)
+    if local_batch_size != len(training_input["prompts"]):
```
wang2yn84: It's also possible that we apply DP on prompts so it will break the assumption here?

tunix/rl/grpo/grpo_learner.py
```diff
@@ -214,6 +216,22 @@ def _generate_and_compute_advantage(
...
+      # In multi-controller: prompts are replicated but completions are local
```
wang2yn84: Training inputs and completion text should have the same sharding, isn't it?

tunix/rl/grpo/grpo_learner.py
```diff
@@ -258,13 +276,23 @@ def _generate_and_compute_advantage(
...
+          prompts=local_training_input["prompts"],
           completions=completion_text,
           mode=mode,
-          **{k: v for k, v in training_input.items() if k != "prompts"},
+          **{k: v for k, v in local_training_input.items() if k != "prompts"},
       )
+      
+      # Create globally sharded array from process-local rewards
+      mesh = self.rl_cluster.r2m[rl_cluster_lib.Role.ACTOR]
+      if not mesh.empty:
```
wang2yn84: This will almost always eval to True because Actor mesh should always be available. Are you still trying to identify if it's multi controller env here?

yhtang: Thanks for the clarification. I wasn’t sure whether the Actor mesh is always present across different offloading/rollout configurations. I’ve gone ahead and removed the predicate as you suggested.

tunix/rl/grpo/grpo_learner.py
yhtang: I think this is an important API contract we should clarify.

Under FSDP, what do we want `training_input` as seen by `RLLearner._generate_and_compute_advantage()` to represent?

1. Process-local: each FSDP process only sees its local shard of the batch, or
2. Global: each process sees the full set of prompts/completions.

Either approach can work, but we should make a clear choice and document it so users know how to structure and shard their datasets.

My preference is **(1) process-local**: each rank operates on its local slice of `training_input` and `completion_text` with matching sharding. That avoids broadcasting prompts across processes and keeps the mental model closer to "local batch in, local completions out".

Please let me know what you think and I'll revise this PR accordingly.

tunix/rl/grpo/grpo_learner.py
yhtang: Could you clarify whether "DP" here refers to FSDP, or to an additional data-parallel layer where the model is replicated on top of FSDP?

tunix/rl/grpo/grpo_learner.py
wang2yn84: +1, I will go with option 1 that FSDP process only sees the local shards for advantage calculation. First the advantage will be feed into the loss function of training and therefore it should align with the definition of FSDP. Second, we don't need the cross batch calculation until the gradient update and it makes sense to keep only the local shards. Therefore in the process of rewards -> advantage -> loss, we should shard it properly the moment when we transfer it from host to device.

tunix/rl/grpo/grpo_learner.py
wang2yn84: Oh sorry about the confusion, here DP refers to the sharding on batch dimension which is the same as FSDP. It seems you are assuming the inputs are replicated and outputs are sharded by FSDP/DP. I think we should shard the inputs also with FSDP/DP. If you observe the current inputs are replicated, it's a mistake and should be fixed.

tunix/generate/sampler.py
```diff
@@ -608,6 +628,8 @@ def _sample_step(
         sampler_state.cache,
         attention_mask,
     )
+    # Ensure full-vocab logits are replicated across tp for sampling.
+    logits = wsc(logits, P('fsdp', None, None))
```
wang2yn84: Redundant?

PR #837, opened 2025-12-04
fix the issue when eager mode jax is triggered in undesired places

Hanjun-Dai:
Existing implementation will trigger the jax eager mode, which compiles many small pieces of code on-demand  (in order to get it run on TPU) at every training iteration. This will cause several issues:

## Before the fix:

- We see a lot of complains from wandb: WARNING Tried to log to step 0 that is less than the current step xxx. Steps must be monotonically increasing, so this data will be ignored. See https://wandb.me/define-metric to log data out of order.
This is because the jax compilation message also gets monitored and logged into wandb. But jax doesn't know the step and default step 0 is used. That's why it prints a lot of these warnings during training.

- TPU utilization: the eager mode especially the re-compilation at every iteration takes time, which can make things slow and lower the TPU utilization
Screenshot 2025-12-04 at 2 07 00 AM

- Probably too many cached binaries will be created until all kinds of shapes have been compiled.

## After the fix:

- wandb warning is gone

- TPU utilization is higher and it should run faster
Screenshot 2025-12-04 at 2 04 34 AM

**Checklist**

### DIFF
--- a/tunix/generate/utils.py
+++ b/tunix/generate/utils.py
@@ -26,6 +26,7 @@ from flax import nnx
 import jax
 from jax import lax
 import jax.numpy as jnp
+import numpy as np
 
 
 def compute_attention_masks(
@@ -88,12 +89,12 @@ def next_power_of_2(x: int) -> int:
 
 
 def pad_to_length(
-    x: jax.Array,
+    x: np.ndarray,
     target_length: int,
     pad_value: int = 0,
     left=False,
     axis: int = 0,
-) -> jax.Array:
+) -> np.ndarray:
   """Pads a JAX array to a specified target length along a given axis.
 
   Args:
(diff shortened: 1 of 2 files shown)


tunix/generate/utils.py
```diff
@@ -88,12 +89,12 @@ def next_power_of_2(x: int) -> int:
 
 
 def pad_to_length(
```
wang2yn84, line 91: There are couple of other callersites of this utility function. E.g. the vllm_sampler and sglang_jax_sampler wraps prompt id as Jax array. Can you update them accordingly. The other callersites such as dpo_trainer you can leave them there. We need to refactor them to use the generate/utils.py.

wang2yn84: Actually let me just merge this PR, it's very important to us and I'd like to land it ASAP. I will fix the rest from internal.

PR #841, opened 2025-12-04
Support custom MaxText (vLLM) models in sampler and rollout.

NicoGrande:
Enables direct weight synchronization and configuration for custom MaxText models within the vLLM sampler and rollout workers. This allows users to supply a `maxtext_config` via `additional_config` without needing explicit key mappings.

This PR enables support for the MaxText on vLLM workstream, by allowing a MaxText native model served on vLLM to be used for RL sampling in Tunix. This feature allows for custom MaxText models, which do not have a corresponding huggingface implementation, to be finetuned using Tunix + vLLM.

To do this, this PR adds support for a few new config entries required to initialize the MaxText on vLLM model. Additionally, it adds a new weight sync mapping named  `transfer_state_directly` which is used when there is no mapping required to sync weights between vLLM and the trainer.

**Reference**

For more details please refer to [go/maxtext-vllm-eng-plan](http://goto.google.com/maxtext-vllm-eng-plan)

**Checklist**

### DIFF
--- a/tunix/generate/utils.py
+++ b/tunix/generate/utils.py
@@ -20,7 +20,7 @@ import gc
 import logging
 import math
 import re
-from typing import Any, Dict, Iterator, List, Optional, Tuple
+from typing import Any, Callable, Dict, Iterator, List, Mapping, Optional, Tuple
 
 from flax import nnx
 import jax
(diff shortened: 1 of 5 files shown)


tunix/rl/common.py
```diff
@@ -151,7 +151,7 @@ def selective_log_softmax(logits: jax.Array, input_ids: jax.Array) -> jax.Array:
     Selected log probabilities.
   """
   logps = jax.nn.log_softmax(logits, axis=-1)
-  per_token_logps = jnp.take_along_axis(logps, input_ids[..., None], axis=-1)
+  per_token_logps = jnp.take_along_axis(logps, input_ids.astype(jnp.int32)[..., None], axis=-1)
```
tianshub: input_ids should already by in int type, why do you need this cast?

tunix/generate/vllm_sampler.py
```diff
@@ -135,14 +138,38 @@ def update_params(
...
+          src_state=updated_weights,
+          dst_state=self.transformer_state,
+          key_mappings=self.to_hf_key_mappings,
+          key_mapping_hook_fns=self.to_hf_hook_fns,
+          transpose_keys=self.to_hf_transpose_keys,
+          reshard_fn=reshard.reshard_pytree,
+      )
+    else:
+      # Direct Weight Sync (e.g. MaxText -> MaxText)
+      logging.info(
```
tianshub: this logging will show up on every sync weight, make it debug? or log once

tunix/generate/vllm_sampler.py
```diff
@@ -135,14 +138,38 @@ def update_params(
...
+
+      additional_config = self.config.additional_config or {}
+      if 'maxtext_config' not in additional_config:
+        raise ValueError(
+            "Direct weight synchronization is currently supported only for "
+            "MaxText models. The required 'maxtext_config' key is missing "
+            "from 'additional_config'."
+        )
+
+      utils.transfer_state_directly(
```
tianshub, line 165: I'm trying to understand why can't we reuse the transfer_state_with_mappings for maxtext model and vLLM model? if we need to make that function more generic to support some customized maxtext model structure, can we do it there?

tunix/generate/utils.py
```diff
@@ -701,6 +701,97 @@ def transfer_state_with_mappings(
...
+  def to_pure_spec(node):
+    if hasattr(node, 'to_pure_dict'):
+      node = node.to_pure_dict()
+    if isinstance(node, dict):
+      return {k: to_pure_spec(v) for k, v in node.items()}
+    if hasattr(node, 'value'):  # Unwrap Param/Cache/Variable
+      return node.value
+    return node
+
+  # Helper: Intersect Trees (Handle KVCache/RNG mismatches)
```
tianshub, line 760: why is KV cache relevant during state transfer?

tunix/rl/common.py
NicoGrande: It shouldn't be necessary, I believe this was an empirical observation we made but most likely due to a bug in our early implementation

tunix/generate/vllm_sampler.py
NicoGrande: Changed to debug.

tunix/generate/vllm_sampler.py
NicoGrande: You are right that this functionality could be merged into the same `transfer_state_with_mappings` method given null mappings. We can introduce this in a follow up PR.

PR #853, opened 2025-12-07
allow users to specify data module outside of tunix; refactor a bit

Hanjun-Dai:
This PR allows users to specify their own data module, so as to setup dataset outside of tunix folder.

The ways to specify a custom dataset module through cli is to set `data_module` arg, with possible variants like:
- `data.coding` # create_dataset is the default function to call in order to get a dataset obj
- `data.coding:create_dataset`   # explicitly specify the function to call
- `data.coding:get_my_dataset`  # explicitly specify any function name to replace the default one
- `data.coding:create_dataset(name='coding_v0')`  # function can have their own args
- `data.coding:create_dataset('coding_v0', split='train')`  # and mixed args+kwargs
- `/home/user/project/data/coding.py:get_dataset`  # or specify a path to the file, plus all above options on func spec

Meanwhile, the `post_init_dataset` is added to do the following things by default:
- filter out the prompts if it is longer than the max_prompt_length
- batching
- limit num_batches when necessary

The filter caused by `max_prompt_length` added above should be the only place that will affect existing examples. 

**Reference**

**Colab Notebook**

**Checklist**

### DIFF
--- a/tunix/cli/utils/data.py
+++ b/tunix/cli/utils/data.py
@@ -1,0 +1,116 @@
+import ast
+import importlib
+import functools
+import grain
+import os
+from typing import Any
+from tunix.generate.tokenizer_adapter import TokenizerAdapter
+
+
+def apply_chat_template(x, tokenizer: TokenizerAdapter) -> dict[str, Any]:
+  return {
+      "prompts": tokenizer.apply_chat_template(
+          x["prompt"], tokenize=False, add_generation_prompt=True
+      ),
+      **{k: v for k, v in x.items() if k != "prompt"},
+  }
+
+
+def parse_call_string(arg_string: str) -> tuple[list[Any], dict[str, Any]]:
+  if not arg_string.strip():
+    return [], {}
+
+  fake_expression = f"dummy_func({arg_string})"
+  try:
+    tree = ast.parse(fake_expression)
+  except SyntaxError:
+    raise ValueError(f"Invalid argument syntax: {arg_string}")
+
+  call_node = tree.body[0].value
+  
+  parsed_args = []
+  for node in call_node.args:
+    parsed_args.append(ast.literal_eval(node))
+
+  parsed_kwargs = {}
+  for keyword in call_node.keywords:
+    parsed_kwargs[keyword.arg] = ast.literal_eval(keyword.value)
+
+  return parsed_args, parsed_kwargs
+
+
+def get_dataset_from_module(specifier: str, tokenizer: TokenizerAdapter):
+  """Get dataset from module.
+
+  Examples of specifier:
+    - "data.coding" # create_dataset is the default function
+    - "data.coding:create_dataset"
+    - "data.coding:get_my_dataset"
+    - "data.coding:create_dataset(name='coding_v0')"
+    - "data.coding:create_dataset('coding_v0', split='train')"
+    - "/home/user/project/data/coding.py:get_dataset"
+
+  Args:
+    specifier: The specifier of the module.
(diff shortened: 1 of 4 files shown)


wang2yn84:
We support loading from parquet file, does it help in your case?

Hanjun-Dai:
> We support loading from parquet file, does it help in your case?

This is a different kind of usage which directly returns a grain dataset, and is more straightforward to use than parquet files (e.g., programmatically change the data mixture, data weights, etc, which is not possible with parquet but is in critical needed for tuning model recipe)

tunix/cli/utils/data.py
```diff
@@ -0,0 +1,120 @@
+import ast
```
tianshub, line 1: thanks for adding the support @Hanjun-Dai , very neat.
Can we add some type hints to the new APIs?

tunix/cli/utils/data.py
```diff
@@ -0,0 +1,120 @@
...
+from asyncio import streams
+import importlib
+import functools
+import grain
+import os
+
+
+def apply_chat_template(x, tokenizer):
```
wang2yn84: Shall we use the apply_chat_template here: https://github.com/google/tunix/blob/0863146dc20c01241b82a8620586f50fd019b753/tunix/generate/tokenizer_adapter.py#L138?

tunix/cli/utils/data.py
```diff
@@ -0,0 +1,120 @@
...
+  return {
+      "prompts": tokenizer.apply_chat_template(
+          x["prompt"], tokenize=False, add_generation_prompt=True
+      ),
+      **{k: v for k, v in x.items() if k != "prompt"},
+  }
+
+
+def get_dataset_from_parquet(parquet_path, tokenizer):
```
wang2yn84: We've refactored this, it's redundant with https://github.com/google/tunix/blob/0863146dc20c01241b82a8620586f50fd019b753/tunix/examples/data/math_dataset.py#L161

tunix/cli/utils/data.py
```diff
@@ -0,0 +1,120 @@
...
+  tokenizer,
+  batch_size: int,
+  num_batches: int | None,
+  max_prompt_length: int | None,
+):
+  if max_prompt_length is not None and max_prompt_length > 0:
+    def prompt_length_filter(x):
+      tokens = tokenizer.tokenize(x["prompts"])
+      return len(tokens) <= max_prompt_length
+    dataset = dataset.filter(prompt_length_filter).to_iter_dataset()
```
wang2yn84, line 109: Thank you for this! It's an important fix.

tunix/cli/utils/data.py
Hanjun-Dai: yes! added type hints to the functions in this data util

tunix/cli/utils/data.py
Hanjun-Dai: great point, I'm using the TokenizerAdapter now which exactly points to the implementation under generate/tokenizer_adapter.py; this function now is a mapper that points to TokenizerAdapter but also keeps other fields untouched, and translate 'prompt' to 'prompts' following the naming convention between different platforms.

tunix/cli/utils/data.py
Hanjun-Dai: deleted!

tunix/cli/utils/data.py
Hanjun-Dai: 😄

Hanjun-Dai:
thank you Tianshu and Lance for the review, this is very helpful. I've made the changes accordingly!

PR #858, opened 2025-12-08
add qwen4b model config which uses tie embedding

Hanjun-Dai:
qwen4b (and models smaller than that) uses tie embedding [1]
but somehow the 0.6b and 1.7b models have duplicated lm_head which makes loading easier.
This PR fixes the issue of complains about not finding lm_head in qwen4b (since it uses tie embedding)

Tested 4b models and it works.

**Reference**
[1] https://qwenlm.github.io/blog/qwen3/

**Checklist**

### DIFF
--- a/tests/models/naming_test.py
+++ b/tests/models/naming_test.py
@@ -231,6 +231,18 @@ _MODEL_INFO_MAP = {
         id='qwen3_1p7b',
         category='qwen3',
     ),
+    'qwen3-4b': ModelTestInfo(
+        family='qwen3',
+        version='4b',
+        id='qwen3_4b',
+        category='qwen3',
+    ),
+    'qwen3-4b-2507': ModelTestInfo(
+        family='qwen3',
+        version='4b_2507',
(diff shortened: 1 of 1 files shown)


tunix/models/qwen3/model.py
```diff
@@ -128,6 +129,36 @@ def qwen3_1p7b(cls):  # qwen3-1.7B
         rope_theta=1_000_000,
     )
 
+  @classmethod
+  def qwen3_4b(cls):  # qwen3-4B
```
tianshub, line 133: thanks for adding these @Hanjun-Dai , we have a test for the model names, so please add the new one to here https://github.com/google/tunix/blob/c1fb42be652c8f461cf56c40d710b0b8ed0de13d/tests/models/naming_test.py#L233

Hanjun-Dai: sure, just added to the naming_test, could you please trigger the workflow again? thank you!

Issue #902, opened 2025-12-16
RL utils.py has an incorrect check for is_positive_integer


Ajay-26:
Hi! Raising this issue to note that there might be a bug in initializing cluster config for RL. 

The following snippet crashes when I run it:

```
# Training config
cluster_config = rl_cluster_lib.ClusterConfig(
    role_to_mesh={
        rl_cluster_lib.Role.ACTOR: mesh,
        rl_cluster_lib.Role.REFERENCE: mesh,
        rl_cluster_lib.Role.ROLLOUT: mesh,
    },
    rollout_engine='vanilla',
    offload_to_cpu=False,
    training_config=rl_cluster_lib.RLTrainingConfig(
        actor_optimizer=optimizer,
        eval_every_n_steps=EVAL_EVERY_N_STEPS,
        max_steps=MAX_STEPS,
        mini_batch_size=TRAIN_MICRO_BATCH_SIZE,
        train_micro_batch_size=TRAIN_MICRO_BATCH_SIZE,
        # metrics logging
        metrics_logging_options=metrics_logging_options,
        # checkpoint saving
        checkpoint_root_directory=CKPT_DIR,
        checkpointing_options=checkpointing_options,
    ),
    rollout_config=base_rollout.RolloutConfig(
        max_tokens_to_generate=TOTAL_GENERATION_STEPS,
        max_prompt_length=MAX_PROMPT_LENGTH,
        kv_cache_size=MAX_PROMPT_LENGTH + TOTAL_GENERATION_STEPS + 256,
        temperature=TEMPERATURE,
        top_p=TOP_P,
        top_k=TOP_K,
        eos_tokens=EOS_TOKENS,
    ),
)

grpo_config = GRPOConfig(
    num_generations=NUM_GENERATIONS,
    num_iterations=NUM_ITERATIONS,
    beta=BETA,
    epsilon=EPSILON,
)
```

**Expected Behavior**
The given snippet should simply work. 

**Actual Behavior**
The snippet raises the following error:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_181/593643255.py in <cell line: 0>()
      8     rollout_engine='vanilla',
      9     offload_to_cpu=False,
---> 10     training_config=rl_cluster_lib.RLTrainingConfig(
     11         actor_optimizer=optimizer,
     12         eval_every_n_steps=EVAL_EVERY_N_STEPS,

[...]

PR #903, opened 2025-12-17
Fix AttributeError in is_positive_integer on Python 3.11

bug39:
Remove .is_integer() call as int.is_integer() was only added in Python 3.12.
pyproject.toml specifies: requires-python = ">=3.11", yet according to https://docs.python.org/3.12/library/stdtypes.html  `int.is_integer()` was only added in Python 3.12 for duck-type compatibility with `float.is_integer()

Resolves #\<[902](https://github.com/google/tunix/issues/902)\>

### DIFF
--- a/tunix/rl/utils.py
+++ b/tunix/rl/utils.py
@@ -31,11 +31,14 @@ import numpy as np
 Mesh = jax.sharding.Mesh
 NamedSharding = jax.sharding.NamedSharding
 
-
-def is_positive_integer(value: int | None, name: str):
+def is_positive_integer(value: int | float | None, name: str) -> None:
   """Checks if the value is a positive integer."""
-  if value is not None and value <= 0:
+  if value is not None and (not is_integer_value(value) or value <= 0):
     raise ValueError(f"{name} must be a positive integer. Got: {value}")
+
+def is_integer_value(value) -> bool:
+  """Checks if the value is an integer"""
+  return (isinstance(value, int) and not isinstance(value, bool)) or (isinstance(value, float) and value.is_integer())
 
 
 def check_divisibility(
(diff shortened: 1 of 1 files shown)


wang2yn84 (commented):
Thank you for the PR! Let's try the following and add a unit test here.

def is_integer_value(x):
    return isinstance(x, int) or (
        isinstance(x, float) and x.is_integer()
    )

bug39:
> Thank you for the PR! Let's try the following and add a unit test here.
> 
> def is_integer_value(x): return isinstance(x, int) or ( isinstance(x, float) and x.is_integer() )

Addressed by adding an is_integer_value helper function, also adding a check that x is not a boolean (is_instance(bool, int) would return True). I'm not sure if you wanted me to include an assert block. Please let me know!

wang2yn84:
Thank you for the update. Can you add a simple test cover various data types to https://github.com/google/tunix/blob/main/tests/rl/rl_utils_test.py?

wang2yn84:
And also please squash the commits before merge, thank you!

rajasekharporeddy:
Hi @bug39, Could you please add a sample test and squash the commits?

kitrakrev:
Hi @bug39 — gentle ping. Happy to pick this up if you're no longer working on it: add the requested test in `tests/rl/rl_utils_test.py` covering int/float/bool/None/negatives, then squash. Will open a follow-up PR referencing this one and crediting your fix. Let me know if you'd prefer to finish it yourself.

rajasekharporeddy:
Merged in [e5f4a04](https://github.com/google/tunix/commit/e5f4a04e2f5f1ff7b2aee69f39a480a156bd2b26)

PR #926, opened 2025-12-20
refactor rl config: 1) remove unnecessary params; 2) model_config consistency

Hanjun-Dai:
There are actor_model_config, reference_model_config and rollout_model_config where there value might conflict with each other, if user don't overwrite them all explicitly. The benefits are:

# Consistency

Current
Screenshot 2025-12-20 at 3 38 54 PM
where user overwrites the model to qwen-0.6b but the rest of the configs still show llama-8b, which is very confusing. 

After this patch
Screenshot 2025-12-20 at 2 00 35 PM

# Reusing checks in cli/config.py

A lot of checks / validations are based on model_config. Right now it seems for RL the config is set through reference_model_config. So to make it consistent, we can overwrite the model_config when no explicit configuration are set to model_config.

**Checklist**

### DIFF
--- a/tunix/cli/base_config.yaml
+++ b/tunix/cli/base_config.yaml
@@ -12,6 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
+################################# LoRa #################################
+_lora_config: &base_lora_config
+  module_path: ".*q_einsum|.*kv_einsum|.*gate_proj|.*down_proj|.*up_proj"
+  rank: 16
+  alpha: 2.0
+  weight_qtype: "nf4"
+  tile_size: 256
+
+################################## MESH ##################################
+_mesh: &base_mesh
+  shape: "(2,2)"
+  # "('fsdp',)"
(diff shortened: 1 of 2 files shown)


wang2yn84:
Thank you for your PR! Can you add the correspondent test cases to tunix/cli/config_test.py? Thank you!

wang2yn84:
Also, can you provide the command to generate the "current" model config? Do you overwrite the model of the base config?

Hanjun-Dai:
> Thank you for your PR! Can you add the correspondent test cases to tunix/cli/config_test.py? Thank you!

working on it!

Hanjun-Dai:
> Also, can you provide the command to generate the "current" model config? Do you overwrite the model of the base config?

so the existing scripts should still work. If there's no direct modification to model_config (e.g., through cli), then model_config will take the value from the reference_model_config (reference_model_config is what's being used right now when setting up RL)

Hanjun-Dai:
closing this one as https://github.com/google/tunix/pull/960 is providing a better solution

PR #1006, opened 2026-01-27
Adding new sharding and performance args for vLLM

NicoGrande:
This PR adds new flags to enable attn_dp in the vLLM mesh, as well as flags to tune the performance of vLLM generation. More information on these flags and their effects can be found here:
https://docs.vllm.ai/en/stable/configuration/optimization/#performance-tuning-with-chunked-prefill

**Checklist**


tunix/rl/rollout/base_rollout.py
```diff
@@ -142,6 +142,13 @@ class RolloutConfig:
   rollout_vllm_hf_config_path: str | None = None
   rollout_vllm_additional_config: dict[str, Any] | None = None
 
+  # Whether to enable data parallel in attention for vLLM rollout engine.
+  rollout_vllm_enable_dp_attention: bool = False
+
+  # Maximum number of batched tokens and sequences for vLLM rollout engine.
```
wang2yn84: Can you put more detailed description here?

tunix/rl/rollout/base_rollout.py
```diff
@@ -142,6 +142,13 @@ class RolloutConfig:
...
+  rollout_vllm_max_num_batched_tokens: int = 2048
+  rollout_vllm_max_num_seqs: int = 256
```
wang2yn84: ditto

tunix/generate/vllm_sampler.py
```diff
@@ -67,6 +67,9 @@ class VllmConfig:
   async_scheduling: bool = False
   tensor_parallel_size: int = -1
   data_parallel_size: int = -1
+  enable_dp_attention: bool = False
+  max_num_batched_tokens: int = 2048
+  max_num_seqs: int = 256
```
lc5211: Why do we set the default values for 2048 and 256 respectively? I wonder if users should be forced to set the correct values for these two instead of depending on default values.

tunix/rl/rollout/base_rollout.py
NicoGrande: Done.

tunix/rl/rollout/base_rollout.py
NicoGrande: Done.

tunix/generate/vllm_sampler.py
NicoGrande: I changed these values to be optional. If no value is specified, then the underlying default from vLLM will be used.

PR #1143, opened 2026-02-24
Adding fix to logical axis cm for RL.

NicoGrande:
Resolves [b/487002385](https://b.corp.google.com/487002385)

Ensure that the correct logical axis rules are in place to avoid global context managers.

**Checklist**

### DIFF
--- a/tunix/rl/rl_cluster.py
+++ b/tunix/rl/rl_cluster.py
@@ -408,13 +408,18 @@ class RLCluster:
         raise ValueError("Rollout vllm model version or path is missing!")
 
       # TODO(linchai): maybe support offloading for vllm rollout.
-      self._rollout = vllm_rollout.VllmRollout(
-          self.rollout_actor,
-          self.tokenizer,
-          cache_config_or_size=max_kv_cache_size,
-          mesh=self.r2m[Role.ROLLOUT],
-          rollout_config=loaded_vllm_config,
-      )
+      # vLLM handles model initialization and loading internally, so we need to provide
+      # logical axis rules for vLLM to correctly shard the model on the rollout mesh.
+      # This is important for out-of-tree models in vLLM that are implemented with custom
+      # logical axis rules, like is the case for MaxText models.
+      with self._get_logical_axis_rules_cm(Role.ROLLOUT):
+        self._rollout = vllm_rollout.VllmRollout(
+            self.rollout_actor,
+            self.tokenizer,
+            cache_config_or_size=max_kv_cache_size,
+            mesh=self.r2m[Role.ROLLOUT],
+            rollout_config=loaded_vllm_config,
+        )
     elif self.cluster_config.rollout_engine == "sglang_jax":
       from tunix.rl.rollout import sglang_jax_rollout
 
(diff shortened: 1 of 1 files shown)


tunix/rl/rl_cluster.py
```diff
@@ -408,13 +408,14 @@ def _init_cluster(self):
...
 
       # TODO(linchai): maybe support offloading for vllm rollout.
-      self._rollout = vllm_rollout.VllmRollout(
-          self.rollout_actor,
-          self.tokenizer,
-          cache_config_or_size=max_kv_cache_size,
-          mesh=self.r2m[Role.ROLLOUT],
-          rollout_config=loaded_vllm_config,
-      )
+      with self._get_logical_axis_rules_cm(Role.ROLLOUT):
```
wang2yn84, line 415: It makes sense here to only pass the logical axis because vLLM uses its own mesh.

tunix/rl/rl_cluster.py
```diff
@@ -520,16 +521,19 @@ def _init_cluster(self):
...
-        training_config=actor_config,
-        custom_checkpoint_metadata_fn=lambda: {
-            "global_step": self.global_steps + 1
-        },  # offset by 1 since global_step is incremented after the training loop in rl_learner. # pylint: disable=line-too-long
-        metrics_logger=self._rl_metrics_logger,
-        perf_tracer=self._perf,
-    )
+    with self.cluster_config.role_to_mesh[
+        Role.ACTOR
+    ], self._get_logical_axis_rules_cm(Role.ACTOR):
```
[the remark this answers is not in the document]
wang2yn84: I think it's a good idea, but considering it might block you, can you add a TODO to refactor with another PR? I will go ahead and approve this PR

tunix/rl/rl_cluster.py
wang2yn84: Shall we do the same thing for critics trainer?

tunix/rl/rl_cluster.py
```diff
@@ -408,13 +408,14 @@ def _init_cluster(self):
...
       # TODO(linchai): maybe support offloading for vllm rollout.
-      self._rollout = vllm_rollout.VllmRollout(
-          self.rollout_actor,
-          self.tokenizer,
-          cache_config_or_size=max_kv_cache_size,
-          mesh=self.r2m[Role.ROLLOUT],
-          rollout_config=loaded_vllm_config,
-      )
+      with self._get_logical_axis_rules_cm(Role.ROLLOUT):
+        self._rollout = vllm_rollout.VllmRollout(
```
wang2yn84, line 416: Can you add some comments to explain why you only add the logical axis rule to vllm rollout, but rather vanilla and sglang backends?

NicoGrande: Done

tunix/rl/rl_cluster.py
NicoGrande: Done

tunix/rl/rl_cluster.py
NicoGrande: Yes exactly :)

tunix/rl/rl_cluster.py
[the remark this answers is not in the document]
NicoGrande: Added a TODO.

tunix/rl/rl_cluster.py
wang2yn84: Hi Nico, I was trying to ask for comments about why this feature is specific to vLLM, but not vanilla and sglang, I think the logical axis context here is for MaxText managed model in vLLM?

tunix/rl/rl_cluster.py
```diff
@@ -1031,6 +1040,7 @@ def get_rewards(
           eos_id,
       )
 
+  # TODO(lancewang): Combine mesh and logical axis rules into a single context manager.
```
wang2yn84: Can you create a bug tracker instead of assigning it to me?

NicoGrande: Done :)

tunix/rl/rl_cluster.py
NicoGrande: Thats right - more broadly for any model plugged into vLLM which has custom logical axis tules.

PR #1169, opened 2026-02-26
Add support for vllm sampler kwargs.

NicoGrande:
Add support for sampling kwargs to be passed in via vLLMConfig. This mirrors the recent engine_kwargs argument added to vLLMConfig.

**Checklist**

### DIFF
--- a/tests/generate/vllm_sampler_test.py
+++ b/tests/generate/vllm_sampler_test.py
@@ -31,6 +31,7 @@ import transformers
 from tunix.generate import mappings
 from tunix.generate import sampler as vanilla_sampler
 from tunix.generate import vllm_sampler
+from tunix.models.dummy_model_creator import create_dummy_model
 from tunix.models.llama3 import model as llama_lib
 from tunix.models.llama3 import params as llama_params
 from tunix.sft import utils as base_utils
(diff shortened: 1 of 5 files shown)


tunix/generate/vllm_sampler.py
```diff
@@ -418,16 +421,21 @@ def __call__(
...
               logging.INFO,
               "Received additional kwargs that are not explicitly defined in"
               f" the method signature: {kwargs}. These will be forwarded to the"
               " underlying sampler, but please ensure that they are valid.",
               1,
-          )
+          )   
+          for key, value in kwargs.items():
+            logging.info(
+                "Sampler kwargs setting key '%s' with value '%s'.", key, value
```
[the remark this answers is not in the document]
NicoGrande: Done.

tunix/generate/vllm_sampler.py
```diff
@@ -418,16 +421,21 @@ def __call__(
...
-          sampling_params.update(**kwargs)
           logging.log_first_n(
               logging.INFO,
               "Received additional kwargs that are not explicitly defined in"
               f" the method signature: {kwargs}. These will be forwarded to the"
               " underlying sampler, but please ensure that they are valid.",
               1,
-          )
+          )   
+          for key, value in kwargs.items():
```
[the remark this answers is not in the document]
NicoGrande: Done.

tunix/generate/vllm_sampler.py
```diff
@@ -71,6 +71,9 @@ class VllmConfig:
       init=False, default_factory=dict
   )
 
+  # vLLM sampler args that can be directly passed in without additional processing, e.g. temperature, stop etc.
+  sampler_kwargs: Dict[str, Any] = dataclasses.field(default_factory=dict)
```
wang2yn84: Can we rename to sampling_kwargs?

tunix/rl/rollout/base_rollout.py
```diff
@@ -157,9 +157,12 @@ class RolloutConfig:
   # Maximum number of concurrent sequences allowed to be processed in vLLM.
   rollout_vllm_max_num_seqs: Optional[int] = None
 
-  # Additional keyword arguments forwarded directly to the vLLM sampler/engine.
+  # Additional keyword arguments forwarded directly to the vLLM engine constructor.
   rollout_vllm_kwargs: dict[str, Any] = dataclasses.field(default_factory=dict)
 
+  # Additional keyword arguments forwarded directly to the vLLM sampler.
+  rollout_vllm_sampler_kwargs: dict[str, Any] = dataclasses.field(default_factory=dict)
```
wang2yn84: Sampling?

tunix/generate/vllm_sampler.py
NicoGrande: Done.

tunix/rl/rollout/base_rollout.py
NicoGrande: Done.

NicoGrande:
> Can you add a test for this?

I added a couple of unit tests. LMK what you think.

tests/generate/vllm_sampler_test.py
```diff
@@ -357,6 +357,167 @@ async def dispatch_requests():
         ),
     )
 
+  def test_vllm_sampler_sampling_kwargs(self):
+    """Test that sampling kwargs are correctly applied to sampling_params."""
+    tunix_model, _ = self.load_llama3_model(
+        self.repo_id, enable_lora=self.enable_lora
+    )
+
+    base_utils.show_hbm_usage("After loading tunix model")
```
wang2yn84: Let's remove this in the test?

tests/generate/vllm_sampler_test.py
```diff
@@ -357,6 +357,167 @@ async def dispatch_requests():
...
+    captured_sampling_params = []
+
+    def mock_generate(prompts, sampling_params, **kwargs):
+      captured_sampling_params.append(sampling_params)
+      return original_generate(prompts, sampling_params, **kwargs)
+
+    vl_sampler.llm.generate = mock_generate
+
+    # Call with additional method kwargs
+    method_sampling_kwargs = {"min_tokens": 10}
```
wang2yn84, line 419: It's ok to test it here, but I'd like to point our this PR doesn't really pass the kwargs in generate, it only passes in during init. Just want to make sure it aligns with what you are trying to implement.

tests/generate/vllm_sampler_test.py
NicoGrande: Sounds good! Will remove!

tests/generate/vllm_sampler_test.py
```diff
@@ -357,6 +357,167 @@ async def dispatch_requests():
...
+    # Check method kwargs
+    self.assertEqual(sampling_params.min_tokens, 10)
+
+  def test_vllm_sampler_sampling_kwargs_override(self):
+    """Test that method kwargs override config sampling_kwargs."""
+    tunix_model, _ = self.load_llama3_model(
+        self.repo_id, enable_lora=self.enable_lora
+    )
+
+    base_utils.show_hbm_usage("After loading tunix model")
```
wang2yn84: Ditto

tests/generate/vllm_sampler_test.py
NicoGrande: Yes, agreed! I just wanted to make sure this case was handled correctly since we have two sources of kwargs now. Does this behavior look good to you?

tests/generate/vllm_sampler_test.py
```diff
@@ -357,6 +357,167 @@ async def dispatch_requests():
         ),
     )
 
+  def test_vllm_sampler_sampling_kwargs(self):
+    """Test that sampling kwargs are correctly applied to sampling_params."""
+    tunix_model, _ = self.load_llama3_model(
```
wang2yn84: Since we are not testing the correctness of output, shall we put dummy_model_creator that Tunix offered instead of the real HF model?

tests/generate/vllm_sampler_test.py
NicoGrande: Done.

tests/generate/vllm_sampler_test.py
NicoGrande: Done.

tunix/generate/vllm_sampler.py
```diff
@@ -424,20 +426,25 @@ def __call__(
       if seed is not None:
         sampling_params.seed = seed
 
-      if kwargs:
+      self.config.sampling_kwargs.update(kwargs)
```
wang2yn84: engine_kwargs ?

tunix/generate/vllm_sampler.py
```diff
@@ -424,20 +426,25 @@ def __call__(
...
+      if self.config.sampling_kwargs:     
         try:
-          sampling_params.update(**kwargs)
           logging.log_first_n(
```
wang2yn84, line 433: The vllm config init will not be called more than once?

tests/generate/vllm_sampler_test.py
wang2yn84: Sure, yea it's good to test in the unit test.

tunix/generate/vllm_sampler.py
NicoGrande: This one should be `sampling_kwargs` since its not going to the llm constructor but to the sampling_params object instead

tunix/generate/vllm_sampler.py
NicoGrande: No - I think it will only be called once. The `kwargs` provided to the __call__ method could change theoretically which is why we want keep the update call here

tunix/generate/vllm_sampler.py
wang2yn84: I mean the kwargs, where does it come from?

tunix/generate/vllm_sampler.py
wang2yn84: Yea I mean if it will only be called once, maybe we don't need to use logging.log_first_n?

tunix/generate/vllm_sampler.py
NicoGrande: Those are the kwargs that are coming from the the `generate()` method. They can be passed in dynamically.

tunix/generate/vllm_sampler.py
NicoGrande: Sounds good! Will update this.

NicoGrande: Done.

tunix/generate/vllm_sampler.py
```diff
@@ -433,23 +425,25 @@ def __call__(
...
-      if kwargs:
+      
+      sampling_kwargs = self.config.sampling_kwargs.copy()
+      sampling_kwargs.update(kwargs)
+      if sampling_kwargs:     
         try:
-          sampling_params.update(**kwargs)
-          logging.log_first_n(
-              logging.INFO,
+          logging.info(
```
wang2yn84: This message will be printed for each call, that's too many, let's use log first n for this.

tunix/generate/vllm_sampler.py
```diff
@@ -433,23 +425,25 @@ def __call__(
...
               " underlying sampler, but please ensure that they are valid.",
-              1,
-          )
-        except Exception as e:
-          logging.log_first_n(
-              logging.INFO,
-              f"Failed to update sampling_params with kwargs: {kwargs}."
+          )   
+          for key, value in sampling_kwargs.items():
+            logging.debug(
```
wang2yn84: Is there anyway to limit this too? Even for debug, that's unnecessary if all the messages are the same.

tunix/generate/vllm_sampler.py
NicoGrande: Reverted this back.

tunix/generate/vllm_sampler.py
NicoGrande: Limited it also using log_first_n.

PR #1192, opened 2026-03-04
Fix AttributeError: 'int' object has no attribute 'is_integer'

### LINKED ISSUE #944: Wrong Python version causing Attribute Error is_integer() in utils.py
https://github.com/google/tunix/blob/3b558520bb65e4cd649f8199362f9dc144e541da/pyproject.toml#L8

This version should be set to '>=3.12' (or higher) at least, because tunix/tunix/rl/utils.py calls value.is_integer() which is safely supported for the base `int` type since Python v3.12.  For example, see below:
https://github.com/google/tunix/blob/3b558520bb65e4cd649f8199362f9dc144e541da/tunix/rl/utils.py#L37

I tested it with v3.11 and confirmed it causes Attribute Error. You should upgrade the default Python version to 3.12 (or higher) in pyproject.toml or substitute all value.is_integer() with the more primitive and safe isinstance().

thakoreh:
## Problem
Fixes #944

The `is_positive_integer()` function in `tunix/rl/utils.py` line 35 calls `value.is_integer()`, which raises an `AttributeError` when `value` is an `int` in Python < 3.12.

The `is_integer()` method exists only on `float` objects, not `int`.

## Solution
Replace `value.is_integer()` with `isinstance(value, int)` which:
- Works correctly for all Python versions (3.11+)
- Properly validates that the value is an integer
- Maintains backward compatibility

## Changes
- Updated type hint to `int | float | None` to accept both types
- Changed validation from `not value.is_integer()` to `not isinstance(value, int)`

## Testing
The function will now correctly:
- Accept positive integers
- Reject floats (even if they represent whole numbers like 5.0)
- Reject non-positive values
- Accept None (no validation)

### DIFF
--- a/tunix/rl/utils.py
+++ b/tunix/rl/utils.py
@@ -32,9 +32,9 @@ Mesh = jax.sharding.Mesh
 NamedSharding = jax.sharding.NamedSharding
 
 
-def is_positive_integer(value: int | None, name: str):
+def is_positive_integer(value: int | float | None, name: str):
   """Checks if the value is positive."""
-  if value is not None and (not value.is_integer() or value <= 0):
+  if value is not None and (not isinstance(value, int) or value <= 0):
     raise ValueError(f"{name} must be a positive integer. Got: {value}")
 
 


thakoreh:
feel like duplicate so closing

PR #1217, opened 2026-03-09
add qwen3 grpo example script with simplemath rewards

andytwigg:
add qwen3 grpo example script with simplemath rewards


examples/rl/grpo/gsm8k/run_qwen3_simplereward.sh
```diff
@@ -0,0 +1,94 @@
+# Copyright 2025 Google LLC
```
tianshub, line 1: run_qwen3_simple_reward.sh?

PR #1238, opened 2026-03-12
allow passing eval_ds parameters through tunix cli, plus bugfixes

andytwigg:
- allow passing eval_ds parameters through cli (for running eval dataset alongside grpo train)

plus bugfixes:
- use correct qwen3 lora regex in examples/rl/grpo/gsm8k/run_qwen3.sh
- add model_download_path to rl/examples scripts to avoid downloading models to same paths


tunix/cli/grpo_main.py
```diff
@@ -145,12 +145,64 @@ def create_perf_config(self, cluster_config: rl_cluster_lib.ClusterConfig):
...
+        perf_config.custom_export_fn_v2 = (
+            perf_export_v2.PerfMetricsExport(
+                perf_metrics_options.log_dir
+                if perf_metrics_options.log_dir
+                else None
+            ).export_metrics
+        )
     return perf_config
 
+  def _load_and_init_dataset(self, tokenizer, data_config_prefix: str = "", split: str = "train"):
```
s-noghabi: The tokenizer parameter should be type-annotated, also the return type.

s-noghabi: can you use the * pseudo-parameter to force keyword-only argument passing for data_config_prefix and split. This helps prevent confusion, especially as they share the same type and have default values.

tunix/cli/grpo_main.py
```diff
@@ -199,33 +251,25 @@ def run_grpo_trainer(self):
     )
 
     tokenizer = grpo_trainer.rl_cluster.tokenizer
-    if self.config.get("data_module", None):
-      dataset = data_lib.get_dataset_from_module(
-          self.config["data_module"],
-          tokenizer,
+
```
s-noghabi: there is some trailing white spaces in this file. Can you format and remove them?

tunix/cli/grpo_main.py
```diff
@@ -199,33 +251,25 @@ def run_grpo_trainer(self):
...
+    
+    # Get dataset splits from config with defaults
+    train_split = self.config.get("train_split", "train")
+    eval_split = self.config.get("eval_split", "test")
+    
+    # Load training dataset
+    dataset = self._load_and_init_dataset(tokenizer, split=train_split)
+    
+    # Load eval dataset if configured
+    eval_dataset = None
```
s-noghabi: can you replace the to a single-assignment form for eval_dataset (e.g., by using an if/else block or a conditional expression) instead of assigning None and then re-binding it.

tunix/cli/grpo_main.py
andytwigg: this isn't trailing whitespace, its just line separation. PTAL

PR #1257, opened 2026-03-17
Add support for dynamically setting the number of steps for GRPO.

niting:
The existing implementation requires these to be specified by the user. We want users to be able to point to their dataset and the implementation should identify the length of dataset. The dataset length is then used to adjust the number of steps required provided the batch size.

Updates the Qwen script to use the feature.

**Reference**

**Colab Notebook**

**Checklist**

This change has been tested locally by doing a GRPO run and running the Qwen script.

### DIFF
--- a/examples/rl/grpo/gsm8k/run_qwen3.sh
+++ b/examples/rl/grpo/gsm8k/run_qwen3.sh
@@ -17,25 +17,16 @@ set -x # Enable xtrace
 
 # specify at cmd line to override defaults, e.g.
 model_name=${model_name:-"Qwen3-1.7B-base"}
-batch_size=${batch_size:-1}
-num_batches=${num_batches:-3738}
+batch_size=${batch_size:-8}
 num_train_epochs=${num_train_epochs:-1}
 warmup_ratio=${warmup_ratio:-0.1}
-train_fraction=${train_fraction:-1.0}
+train_fraction=${train_fraction:-0.8}
 
 echo "Using parameters:"
 echo "  Batch Size: $batch_size"
-echo "  Num Batches: $num_batches"
 echo "  Num Epochs: $num_train_epochs"
 echo "  Warmup Ratio: $warmup_ratio"
 echo "  Train Fraction: $train_fraction"
-
-max_steps_float=$(awk "BEGIN {print $batch_size * $num_batches * $num_train_epochs * $train_fraction}")
-max_steps=$(printf "%.0f" "$max_steps_float")
-warmup_steps=$(awk "BEGIN {printf \"%.0f\", $warmup_ratio * $max_steps}")
-
-echo "Max steps: $max_steps"
-echo "Rounded warmup steps: $warmup_steps"
 
 python3 -m tunix.cli.grpo_main \
   base_config.yaml \
@@ -58,7 +49,6 @@ python3 -m tunix.cli.grpo_main \
   tokenizer_config.add_bos=false \
   dataset_name="gsm8k" \
   batch_size=$batch_size \
-  num_batches=$num_batches \
   num_test_batches=100 \
   num_train_epochs=$num_train_epochs \
   rl_training_config.actor_optimizer_config.opt_type="adamw" \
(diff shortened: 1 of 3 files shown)


tunix/cli/grpo_main.py
```diff
@@ -216,6 +250,7 @@ def run_grpo_trainer(self):
           dataset=self.config["dataset_name"],
           tfds_download=self.config["tfds_download"],
       )
+    self.compute_params(len(dataset))
```
wang2yn84: I believe not all the dataset implements __len__ and that's why we might have to rely on the config to provide the accurate length if we don't want to go through the dataset once.

tunix/cli/grpo_main.py
```diff
@@ -191,14 +187,52 @@ def create_rl_cluster(self):
...
+          num_batches,
+          batch_size,
+      )
+    num_train_epochs = self.config.get("num_train_epochs")
+    if not num_train_epochs:
+      num_train_epochs = 1
+    train_fraction = self.config.get("train_fraction")
+    if not train_fraction:
+      train_fraction = 0.8
+    if not max_steps:
```
wang2yn84, line 232: Shall we check the max_steps against int(num_batches * num_train_epochs * train_fraction) if max_steps is available?

tunix/cli/grpo_main.py
```diff
@@ -191,14 +187,52 @@ def create_rl_cluster(self):
...
+      logging.info(
+          "Dynamically computed num_batches=%d with batch_size=%d",
+          num_batches,
+          batch_size,
+      )
+    num_train_epochs = self.config.get("num_train_epochs")
+    if not num_train_epochs:
+      num_train_epochs = 1
+    train_fraction = self.config.get("train_fraction")
+    if not train_fraction:
```
wang2yn84, line 222: Shall we check the train_fraction range if it's available?

examples/rl/grpo/gsm8k/run_qwen3.sh
```diff
@@ -58,23 +49,18 @@ python3 -m tunix.cli.grpo_main \
...
   batch_size=$batch_size \
-  num_batches=$num_batches \
   num_test_batches=100 \
   num_train_epochs=$num_train_epochs \
   rl_training_config.actor_optimizer_config.opt_type="adamw" \
   rl_training_config.actor_optimizer_config.peak_value=3e-6 \
   rl_training_config.actor_optimizer_config.schedule_type="warmup_cosine_decay_schedule" \
   rl_training_config.actor_optimizer_config.init_value=0.0 \
   rl_training_config.actor_optimizer_config.end_value=0.0 \
-  rl_training_config.actor_optimizer_config.warmup_ratio=$warmup_ratio \
```
wang2yn84, line 69: Maybe we should still setup warm_up ratio to 0.1 instead of relying on the default value?

examples/rl/grpo/gsm8k/run_qwen3.sh
```diff
@@ -17,26 +17,17 @@ set -x # Enable xtrace
 
 # specify at cmd line to override defaults, e.g.
 model_name=${model_name:-"Qwen3-1.7B-base"}
-batch_size=${batch_size:-1}
-num_batches=${num_batches:-3738}
+batch_size=${batch_size:-8}
 num_train_epochs=${num_train_epochs:-1}
 warmup_ratio=${warmup_ratio:-0.1}
 train_fraction=${train_fraction:-1.0}
```
wang2yn84: I think we should set train_fraction? The default value is 0.8.

tunix/cli/grpo_main.py
andytwigg: That makes sense; if the len(..) is not implemented, and num_steps is not specified, then we should throw an error, but if dataset.__len__ exists, then we should allow to not specify num_steps?

niting: Pretty sure more things will break if __len__ was not implemented. See tunix/cli/utils/data.py:177 which splits the train and test sets. It would be odd for a dataset to not have that implemented since they are typically just iterator types.

tunix/cli/grpo_main.py
niting: I can, but I was assuming that the user might specify max_steps when they want to really try out the behavior with different steps. I could potentially cap the max_steps to that value or just leave it as is for now. What do you prefer?

tunix/cli/grpo_main.py
niting: Done. I now fallback to a default of 0.8 in case its out of range.

examples/rl/grpo/gsm8k/run_qwen3.sh
niting: Done. Reverted this change.

examples/rl/grpo/gsm8k/run_qwen3.sh
niting: The train_fraction was 1.0. I updated it to 0.8.

tunix/cli/grpo_main.py
andytwigg: grain supports datasets without len()

niting: Done. I now check if __len__ is available and enforce that max_steps is required when it's not. Note that post_init_dataset in tunix/cli/utils/data.py will still break if __len__ is not available, can fix that separately since that's unrelated to this PR.

wang2yn84: Thank you! The fix makes sense to me.

tunix/cli/grpo_main.py
wang2yn84: Let's cap it and error out early if it exceeds the max.

niting: Done.

PR #1385, opened 2026-04-09
Adding support for fused moe kernels in sampler.

NicoGrande:
This PR introduces full support for the Fused MoE Kernel integrated to MaxText in https://github.com/AI-Hypercomputer/maxtext/pull/3627. 

More specifically, this PR introduces the ability to fuse MoE kernel weights in the MaxText model during weight sync to match the required shapes for the `tpu-inference` Fused MoE kernel. 

Additionally, this PR adds some optimizations for resharding, which are helpful when with large models. First, this PR removes the call to `jax.clear_caches()` in `vllm_sampler.py`. This removes the need to clear Jax compilation caches, which speeds up both training and rollout steps. The downside of this, is that there is more memory fragmentation present in sampler TPUs after clearing the KV cache. To get around fragmentation OOMs, we introduce chunked resharding to reduce the peak HBM consumed during reshard operations for large models.

**Checklist**

### DIFF
--- a/tunix/rl/rollout/vllm_rollout.py
+++ b/tunix/rl/rollout/vllm_rollout.py
@@ -56,6 +56,7 @@ class VllmRollout(base_rollout.BaseRollout):
             tensor_parallel_size=rollout_config.tensor_parallel_size,
             data_parallel_size=rollout_config.data_parallel_size,
             expert_parallel_size=rollout_config.expert_parallel_size,
+            reshard_chunk_size=rollout_config.rollout_vllm_reshard_chunk_size,
             engine_kwargs={
                 "model": rollout_config.rollout_vllm_model_version,
                 "max_model_len": cache_config_or_size,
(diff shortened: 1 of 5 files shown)


tunix/rl/rollout/vllm_rollout.py
```diff
@@ -56,6 +56,7 @@ def __init__(
             tensor_parallel_size=rollout_config.tensor_parallel_size,
             data_parallel_size=rollout_config.data_parallel_size,
             expert_parallel_size=rollout_config.expert_parallel_size,
+            rollout_chunk_size=rollout_config.rollout_vllm_reshard_chunk_size,
```
wang2yn84: rollout_chunk_size seems doesn't need need any special handling, just plumbing it through, shall we pass it via rollout_vllm_kwargs?

NicoGrande: you are right, we only need to plumb it through - however it is not a vLLM engine argument so it shouldn't be passed to the `LLM()` constructor. If we can remove it before passing rollout_vllm_kwargs to the constructor that would work, otherwise adding a new argument for it may be cleaner. LMK your thoughts :)

wang2yn84: My bad, I was looking at the wrong place. Plumbing it through is the right way to go.

PR #1448, opened 2026-04-28
Adding support for TP for fused MoE kernel.

NicoGrande:
This PR adds support for padding MoE model's hidden dimension to be compatible with the GMM_v2 kernel in `tpu-inference`. The corresponding MaxText PR can be found here: https://github.com/AI-Hypercomputer/maxtext/pull/3762

Additionally, this PR introduces some correctness fixes:
- Wrap call to `.unsafe_buffer_pointer()` in try-catch block to avoid errors when calling from Pathways environement.
- Generalizes function which applies KV head replication to also handle MoE padding
- Optimizes implementation of these functions by fusing consecutive Jax operations in a JIT function.
- Adds new unit tests to ensure coverage of all codepaths modifying the model shape as part of weight sync. 

**Checklist**

### DIFF
--- a/tunix/generate/utils.py
+++ b/tunix/generate/utils.py
@@ -703,8 +703,8 @@ def _align_shape(
 
 
 def _apply_dtype_cast(
-    val: jnp.ndarray, tgt_dtype: jnp.dtype, src_key: str
-) -> jnp.ndarray:
+    val: jax.Array | np.ndarray, tgt_dtype: jnp.dtype, src_key: str
+) -> jax.Array | np.ndarray:
   if val.dtype != tgt_dtype:
     logging.log_first_n(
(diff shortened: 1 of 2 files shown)


tunix/generate/utils.py
lc5211, line 1: Can we (maybe as a followup) refine the change a bit to seperate the logics for scanned layers. moe fusion, attention qkv special handlings. Feel like this changes is going wild as we keep adding things for optimizations and special handling. It will be hard to maintain once people loss context.

lc5211 (approved):
Change

tunix/generate/utils.py
NicoGrande: Yes definitely!

PR #1523, opened 2026-05-15
Agentic GRPO: TIS correction, eval dedup, flash-attn segment_ids

colincai-mc:
### Bug fixes

**`tunix/rl/agentic/agentic_rl_learner.py` — eval deduplication**
The eval condition `train_steps % eval_every_n_steps == 0` evaluates to `True` for every micro-iteration within a `grad_accum_steps` window at a step boundary, causing the eval dataset to be replayed that many times per boundary. A `_last_eval_train_step` guard skips duplicate evals.

### Features

**Token-level truncated importance-sampling (TIS) correction** (`sampler_is='token'`)
In multi-turn agentic rollouts the rollout engine and trainer recompute the same sequence with slightly different numerical paths (temperature scaling, attention kernel, padding). The resulting log-probability drift accumulates across turns and acts as unintended noise on the importance ratio. With `sampler_is='token'`, the trainer applies a per-token weight `min(exp(trainer_logp − rollout_logp), threshold)` before aggregating the policy gradient loss. This keeps the effective IS ratio grounded at the trainer's start-of-step distribution rather than the rollout engine's.

**`sampler_is_weights` field in `TrainExample` + `aggregate_loss` application**
The per-token weights computed in `agentic_grpo_learner` are attached to `TrainExample.sampler_is_weights` and applied in `grpo_loss_fn` before loss aggregation, so they affect the gradient through loss magnitude without introducing a stop-gradient bias on the ratio.

**Sampler-trainer `prob_diff` and `pearson` metrics**
Logged every training step. `prob_diff` is the mean absolute probability difference `|softmax(rollout_logit) − softmax(trainer_logit)|` over completion tokens; `pearson` is the per-batch correlation between rollout and trainer log-probabilities. These serve as numerical alignment health checks for the two compute paths.

**Qwen3 model: `segment_ids` for flash attention (splash kernel)**
The splash attention kernel requires per-position segment IDs to correctly compute the causal mask for left-padded sequences. Without this, padding tokens on the left contaminate real-token attention outputs during trainer recompute. This PR plumbs `segment_ids` derived from the non-pad mask through the Qwen3 forward pass.

**`vllm_sampler` EOS token fix**
Prevents an off-by-one in the completion mask when the sampler returns the EOS token at the boundary.

**Trajectory collection: sampler logprobs in output**
The `trajectory_collect_engine` now includes the rollout engine's per-token logprobs in the collected trajectory so the GRPO learner can apply TIS without a second forward pass through the rollout engine.

### DIFF
--- a/tunix/models/qwen3/model.py
+++ b/tunix/models/qwen3/model.py
@@ -486,6 +486,7 @@ class Attention(nnx.Module):
       segment_pos: jaxtyping.Array,
       cache: LayerCache | None,
       attn_mask: jaxtyping.Array | None,
+      segment_ids: jaxtyping.Array | None = None,
   ) -> tuple[LayerCache | None, jaxtyping.Array]:
     seq_len = x.shape[1]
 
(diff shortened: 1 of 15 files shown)


tunix/models/qwen3/model.py
```diff
@@ -486,6 +486,7 @@ def block(
       segment_pos: jaxtyping.Array,
       cache: LayerCache | None,
       attn_mask: jaxtyping.Array | None,
+      segment_ids: jaxtyping.Array | None = None,
```
tianshub, line 489: this is used for sequence packing right? right now we haven't enabled that yet since the seq packing support is still WIP

tunix/rl/agentic/trajectory/trajectory_collect_engine.py
```diff
@@ -272,26 +272,52 @@ async def collect(self, mode: str = "Conversation") -> Any:
       conversation_tokens, conversation_masks, logprobs = [], [], []
       prompt_tokens = getattr(self.agent.trajectory, "prompt_tokens", [])
 
+      # vLLM is configured with `include_stop_str_in_output=True`, so
+      # assistant_tokens from vllm DO include the trailing eos (`<|im_end|>`
+      # for Qwen3). But the chat template separates messages with `\n` after
```
tianshub: chat template refers to the [parser here](https://github.com/google/tunix/blob/90bb1c5b9624f3677edfccb6afd2399713af9b57/tunix/rl/agentic/parser/chat_template_parser/parser.py#L153)? shall we consider removing \n there instead of manually fixing it here?

tunix/rl/rl_cluster.py
```diff
@@ -1109,26 +1118,16 @@ def get_actor_per_token_logps(
...
-      )
+      # NOTE: removed the actor-to-pinned-host offload dance and replaced with
+      # a straight compute using the current actor weights (already on device).
+      # The original code moved actor → host, resharded anchor → device,
+      # computed, moved actor → back. That round-trip left the actor's
+      # layernorm weights on `float32<host>` after the call returned, then
+      # the next train_step's `mul` op crashed with memory_space mismatch.
+      # For sampler-trainer diff metric (single-iteration GRPO), the
+      # "anchor" state equals the current actor state anyway, so we just
+      # compute directly from the actor model.
```
tianshub: when there're multiple model updates (e.g. mini_batch < global_batch), wouldn't we want to use the actor model weights at the very beginning of the global_step?

tunix/models/qwen3/model.py
colinc9: yeah, here it's not for packing, but needed for per-batch padding masks. So the splash attention kernel only accepts a static causal mask. The per-batch dynamic pad positions can only flow in through SegmentIds. Without this, left-padded prompts leak attention from real queries onto pad keys, making softmax garbage that shifts the logits. We use segment_id=0 for pad and segment_id=1 for real tokens to avoid it

tunix/rl/agentic/trajectory/trajectory_collect_engine.py
colincai-mc: addressed in new commit

tunix/rl/rl_cluster.py
colincai-mc: addressed in new commit

lc5211 (approved):

tianshub (approved):

PR #1534, opened 2026-05-22
Add Trackio rollout trace logging

abidlabs:
Hi folks! This PR adds trace logging via Trackio, the free, local-first experiment tracking library from Hugging Face 🤗

This PR follows Tunix's existing rollout and trajectory logging patterns. Specifically, I did this:

- added optional Trackio trace settings to `AlgorithmConfig`
- added PPO/GRPO rollout prompt-completion logging as `trackio.Trace` records with step/reward/advantage metadata
- added agentic trajectory logging as `trackio.Trace` records with trajectory metadata
- documented Trackio rollout trace logging
- added tests with mocked Trackio

Here's what it looks like:

image

AI assistance was used to prepare this PR.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -103,6 +103,27 @@ With the above, agentic_grpo_learner will by default start an async trajectory
 logger which logs the trajectories including prompts, responses, etc. to the
 specified `log_dir`.
 
+### Trackio Trajectory Backend
+
+RL algorithms can also log rollout prompt/completion pairs as Trackio traces
+through the trajectory logger. Set `trackio_project` and a positive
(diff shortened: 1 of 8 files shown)


lc5211:
Thanks for the PR! Can we make trackio as one backend of TrajectoryLogger to make the logging system more manageable? Thanks!

abidlabs:
Thanks for the suggestion @lc5211. I updated this PR to make Trackio a backend of the existing trajectory logger instead of using a separate Trackio-specific logger. The Trackio backend now lives in `tunix/utils/trajectory_logger.py`, standard RL and agentic RL both dispatch traces through `AsyncTrajectoryLogger`

I also tested and confirmed with and end-to-end example that everything is logged correclty:

image

PR #1560, opened 2026-06-03
Add optional output_clamp_value to compute_kl_divergence

py4:
## Summary

`compute_kl_divergence` currently has no upper bound on its output. The three estimators are unbiased on the math, but rare outliers happen in practice — e.g. when the trained policy briefly drifts far from the reference and the `low_var_kl` estimator's `exp(ref_logp - logp)` term saturates bf16 (or, less commonly, overflows fp32 even with the internal cast). When that happens the per-token KL becomes `Inf`/`NaN`, gets aggregated into `kl_loss`, multiplied by `β`, added to the policy-gradient loss, and poisons the gradient for the rest of the step.

This PR:

1. **Adds an optional `clamp_value: float | None = None` kwarg** to `compute_kl_divergence`. When set to a positive float, the returned KL is symmetrically clamped: `clip(kl, -clamp_value, +clamp_value)`. Default `None` is a strict no-op and preserves existing behavior bit-for-bit.

2. **Adds a `kl_clamp_value: float | None = None` field** to `AlgorithmConfig` (the base class inherited by `GRPOConfig`, `PPOConfig`, and `AgenticRLConfig`) so users can set the clamp from a recipe.

3. **Wires the field through to the three existing call sites**:
   - `tunix/rl/algo_core.py` — `ppo_policy_loss_fn` (uses defensive `getattr(algo_config, "kl_clamp_value", None)` since that path reads `kl_coef` via `getattr` too)
   - `tunix/rl/algo_core.py` — `grpo_loss_fn` (reads `algo_config.kl_clamp_value` directly, matching the existing `algo_config.kl_loss_mode` access next to it)
   - `tunix/rl/ppo/ppo_learner.py` (reads `self.algo_config.kl_clamp_value`)

Default `None` everywhere means no behavior change for existing recipes. Users opt in by setting e.g. `kl_clamp_value=10000.0` in their algo config.

## Why opt-in (not a behavior change)

Default `None` keeps existing recipes bit-identical. Reviewers who want to flip the default to a numeric value (e.g. `10000.0`) in a follow-up — happy to do so; that's a one-line change.

## Tests

**`tests/rl/common_test.py`** — 4 new methods on `CommonTest` targeting `compute_kl_divergence`:

| Test | What it checks |
|---|---|
| `test_compute_kl_divergence_clamp_default_is_no_op` (parameterized over `kl` / `mse_kl` / `low_var_kl`) | Calling with `clamp_value=None` returns byte-identical output vs not passing the kwarg at all |
| `test_compute_kl_divergence_clamp_caps_outliers` (parameterized over the three methods) | With `clamp_value=10.0` and ±50 log-prob deltas, every element of the result is in `[-10, +10]` |
| `test_compute_kl_divergence_clamp_passes_through_in_range` | A large clamp (`10000.0`) on small in-range inputs returns identical values to the unclamped path |
| `test_compute_kl_divergence_clamp_symmetric` | A `+100` input clamps to `+5.0`, the matching `-100` input clamps to `-5.0` |

**`tests/rl/algorithm_config_test.py`** — 2 new methods on `AlgorithmConfigTest`:

| Test | What it checks |
|---|---|
| `test_kl_clamp_value_default_is_none` | `AlgorithmConfig().kl_clamp_value is None` |
| `test_kl_clamp_value_round_trips` (parameterized over `10000.0` / `1.0` / `None`) | Value set on the config is what the field returns |

## Files

| File | Δ |
|---|---|
| `tunix/rl/common.py` | +20 / -4 — new kwarg + docstring |
| `tunix/rl/algorithm_config.py` | +8 — new config field |
| `tunix/rl/algo_core.py` | +6 / -1 — wire at both KL call sites |
| `tunix/rl/ppo/ppo_learner.py` | +1 — wire `self.algo_config.kl_clamp_value` |
| `tests/rl/common_test.py` | +71 — 4 new test methods |
| `tests/rl/algorithm_config_test.py` | +15 — 2 new test methods |

## Checklist
- [x] Default (`None`) is bit-identical to prior behavior at every call site
- [x] Configurable from a recipe via `AlgorithmConfig.kl_clamp_value`
- [x] Wired at all three existing `compute_kl_divergence` call sites (PPO loss, GRPO loss, PPO learner)
- [x] Tests cover the kwarg semantics + the config field
- [x] No new dependencies

### DIFF
--- a/tunix/rl/algorithm_config.py
+++ b/tunix/rl/algorithm_config.py
@@ -29,6 +29,14 @@ class AlgorithmConfig:
   advantage_estimator: str = "grpo"
   policy_loss_fn: str = "grpo"
   reward_manager: str = "sequence-level"
+  # Optional symmetric clamp applied to per-token KL inside
+  # `common.compute_kl_divergence`. `None` (default) disables the clamp and
+  # preserves prior behavior bit-for-bit. Set to a positive float (e.g.
+  # `10000.0`) to bound rare outliers — useful when the trained policy
+  # briefly drifts far from the reference and the `low_var_kl` estimator's
+  # `exp(diff)` term saturates bf16 / overflows fp32 and poisons the loss
+  # for the rest of the step.
+  kl_clamp_value: float | None = None
 
   def __post_init__(self):
     valid_algo_variants = [
--- a/tunix/rl/common.py
+++ b/tunix/rl/common.py
@@ -117,6 +117,7 @@ def compute_kl_divergence(
     per_token_logps: jax.Array,
     ref_per_token_logps: jax.Array,
     method: str = "low_var_kl",
+    clamp_value: float | None = None,
 ) -> jax.Array:
   """Compute per token KL divergence between trained and reference policy.
 
@@ -132,6 +133,13 @@ def compute_kl_divergence(
     per_token_logps: Per token log probabilities from the trained policy.
     ref_per_token_logps: Per token log probabilities from the reference policy.
     method: KL penalty method. Defaults to "low_var_kl".
+    clamp_value: Optional symmetric clamp applied to the returned KL,
+      i.e. `clip(kl, -clamp_value, +clamp_value)`. `None`
+      (default) disables the clamp and preserves prior behavior. Set to a
+      positive float (e.g. `10000.0`) to bound rare outliers — useful when
+      the trained policy briefly drifts far from the reference and the
+      `low_var_kl` estimator's `exp(diff)` term can overflow fp32 / saturate
+      bf16.
 
   Returns:
     KL divergence.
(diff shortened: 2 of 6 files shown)


tunix/rl/algorithm_config.py
```diff
@@ -29,6 +29,14 @@ class AlgorithmConfig:
...
   policy_loss_fn: str = "grpo"
   reward_manager: str = "sequence-level"
+  # Optional symmetric clamp applied to per-token KL inside
+  # `common.compute_kl_divergence`. `None` (default) disables the clamp and
+  # preserves prior behavior bit-for-bit. Set to a positive float (e.g.
+  # `10000.0`) to bound rare outliers — useful when the trained policy
+  # briefly drifts far from the reference and the `low_var_kl` estimator's
+  # `exp(diff)` term saturates bf16 / overflows fp32 and poisons the loss
+  # for the rest of the step.
+  kl_output_clamp_value: float | None = None
```
wang2yn84: kl_clamp_value should be sufficient

tunix/rl/common.py
```diff
@@ -117,6 +117,7 @@ def compute_kl_divergence(
     per_token_logps: jax.Array,
     ref_per_token_logps: jax.Array,
     method: str = "low_var_kl",
+    output_clamp_value: float | None = None,
```
wang2yn84: We don't plan to add any intermediate clamping, so "clamp_value" should be sufficent?

wang2yn84 (commented):
Thank you Pooya! 2 comments about naming.

py4:
@wang2yn84 thanks — addressed in ae5c495. Renamed `output_clamp_value` → `clamp_value` (kwarg on `compute_kl_divergence`) and `kl_output_clamp_value` → `kl_clamp_value` (field on `AlgorithmConfig`), both arguments updated at every call site and in the tests.

PR #1623, opened 2026-06-26
Fix bug in trainer loop and optimize parallel rollout

AntonyMei:
- Fixed bug in trainer loop. Instead of repeating #iterations times continuously on one microbatch, we now iterate through the full batch and repeat #iteration times.
- Implement a performance optimization to parallelize rollout with first round training.

**Reference**
https://arxiv.org/pdf/2402.03300

**Checklist**

### DIFF
--- a/tunix/rl/agentic/agentic_rl_learner.py
+++ b/tunix/rl/agentic/agentic_rl_learner.py
@@ -665,9 +665,8 @@ class AgenticRLLearner(abc.ABC, Generic[TConfig]):
                 batch_results=batch,
                 mode=rl_cluster_lib.Mode.TRAIN,
             )
-            for _ in range(self._num_iterations()):
-              for train_example in train_examples:
-                train_data_queue.put(train_example)
+            for train_example in train_examples:
+              train_data_queue.put(train_example)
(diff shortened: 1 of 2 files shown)


tunix/rl/agentic/agentic_rl_learner.py
```diff
@@ -915,22 +876,78 @@ async def _eval_runner_async(current_eval_orchestrator):
...
+        iterations = self._num_iterations()
+
+        for i in range(1, iterations):
+          # TODO(b/483779605) Sub-step checkpointing.
+          self._iter_steps += len(full_batch_chunks)
+
+          # TODO(yixuanm): Eval during iteration too. Skipping for now as we 
+          # will refactor the learner soon.
+          self.rl_cluster.update_actor(
+              full_batch_chunks, None, skip_jit
```
wang2yn84, line 942: We are sending the full batch chunks to trainer anyway, shall we send full_batch_chunks * _num_iterations altogether?

wang2yn84: Actually nvm, let's keep it as is, just to keep it simple.

PR #1627, opened 2026-06-27
Guard duplicate-key check and write in safetensors loader with a lock

### LINKED ISSUE #1259: TOCTOU race condition in concurrent safetensors loading bypasses duplicate key detection
**Expected Behavior**

When two distinct PyTorch tensor keys map to the same JAX key via `torch_key_to_jax_key` during concurrent safetensors loading, the loader should raise a `ValueError` indicating a duplicate key, preventing silent weight corruption.

**Actual Behavior**

The duplicate key check (`if jax_key_mapped in file_loaded_tensors`) and the subsequent dictionary write (`file_loaded_tensors[jax_key_mapped] = current_arr`) in `tunix/models/safetensors_loader.py:219-223` are not protected by any lock, while running inside a `ThreadPoolExecutor(max_workers=os.cpu_count())`.

This creates a TOCTOU (time-of-check-to-time-of-use) race condition:

Thread A: reads file_loaded_tensors[key] → not found     (CHECK passes)
Thread B: reads file_loaded_tensors[key] → not found     (CHECK passes)
Thread A: writes file_loaded_tensors[key] = tensor_A
Thread B: writes file_loaded_tensors[key] = tensor_B     (silently overwrites A)

The `ValueError` on line 220 that is intended to catch duplicates never fires. The model loads successfully with incorrect weights and no error is raised, making this extremely difficult to debug.

The existing `file_lock` only protects `sf_file.get_tensor()` (line 198-199), leaving the check and write on lines 219-223 completely unguarded:

```python
def process_key(k_name, f, sf_file, file_loaded_tensors):
    with file_lock:
        v = sf_file.get_tensor(k_name)           # protected

    jax_key_mapped, transform = torch_utils.torch_key_to_jax_key(key_map, k_name)
    # ... transform tensor ...

    if jax_key_mapped in file_loaded_tensors:     # CHECK — no lock
        raise ValueError(...)
    file_loaded_tensors[jax_key_mapped] = current_arr  # WRITE — no lock
```

Steps to Reproduce the Problem

1. Provide a key mapping where two distinct PyTorch keys resolve to the same JAX key via torch_key_to_jax_key
2. Load a safetensors checkpoint that contains both keys using load_params_from_safetensors
3. Under concurrent execution (os.cpu_count() threads), both threads pass the duplicate check before either writes  the second thread silently overwrites the first tensor

The race is timing dependent and more likely to trigger on machines with higher core counts (larger thread pool) and when the mapped keys are close together in iteration order.

```python
dict_lock = threading.Lock()

def process_key(k_name, f, sf_file, file_loaded_tensors):
    with file_lock:
        v = sf_file.get_tensor(k_name)

    # ... key mapping and transforms ...

    with dict_lock:
        if jax_key_mapped in file_loaded_tensors:
            raise ValueError(f'Duplicate key {jax_key_mapped} found within file {f.name}.')
        file_loaded_tensors[jax_key_mapped] = current_arr
```

**Environment**

OS: Platform independent (threading behavior)
Project Version: main branch (v0.1.6+, commit beca051)
Python: 3.11+

**Checklist**

- [x] I have searched the existing issues for a similar bug report.
- [x] I have provided all the required information in the "Environment" section.
- [x] I have provided a minimal, reproducible example.

AMR5210:
Resolves #1259

This fixes a TOCTOU race in `load_and_create_model_orig`
(`tunix/models/safetensors_loader.py`). `process_key` runs across a
`ThreadPoolExecutor(max_workers=os.cpu_count())`, but the duplicate-key check
`if jax_key_mapped in file_loaded_tensors` and the following write were not
protected by a lock. Two threads whose source keys map to the same jax key can
both pass the check before either writes, so the second silently overwrites the
first and the intended `ValueError` never fires. The model then loads with
incorrect weights and no error.

The fix adds a `dict_lock` and wraps only the check and write in it. The tensor
read, transform, and dtype cast stay outside the lock, so loading remains
parallel with no meaningful speed impact.

**Reference**
- Issue: #1259

**Colab Notebook**
N/A. This is a bug fix to existing loading code and adds no new API.

**Checklist**

### DIFF
--- a/tests/models/safetensors_loader_test.py
+++ b/tests/models/safetensors_loader_test.py
@@ -168,6 +168,45 @@ class SafetensorsLoaderTest(parameterized.TestCase):
         loaded_state,
     )
 
+  def test_load_and_create_model_raises_on_duplicate_key(self):
+    # Two distinct source keys that both map to the same jax key. In the
+    # 'original' loader these are processed on separate threads, so without a
+    # lock around the duplicate-key check and write the second write can
+    # silently overwrite the first (issue #1259). With the lock the duplicate
+    # is always detected.
+    try:
+      st_dir = self.create_tempdir().full_path
+    except Exception:  # pylint: disable=broad-except
+      st_dir = tempfile.TemporaryDirectory().name
+      os.makedirs(st_dir, exist_ok=True)
+
+    tensors = {
+        'lm_head.kernel': np.zeros((2, 2), dtype=np.float32),
+        'lm_head.weight': np.zeros((2, 2), dtype=np.float32),
+    }
+    filename = os.path.join(st_dir, 'model.safetensors')
+    stnp.save_file(tensors, filename)
+
+    def duplicate_key_mapping(config):
+      del config
+      return {
+          r'^lm_head\.kernel$': ('lm_head.kernel', None),
+          r'^lm_head\.weight$': ('lm_head.kernel', None),
+      }
+
+    # The duplicate ValueError is re-raised wrapped as RuntimeError by the
+    # loader, so match on RuntimeError. mode='original' is required because the
+    # default optimized path is single-threaded and does not run this check.
+    with self.assertRaisesRegex(RuntimeError, 'Duplicate key'):
+      safetensors_loader.load_and_create_model(
+          st_dir,
+          test_common.ToyTransformer,
+          self.model.config,
+          duplicate_key_mapping,
+          dtype=jnp.float32,
+          mode='original',
+      )
+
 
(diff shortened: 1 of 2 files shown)


AMR5210:
Opened #1627 to fix this. The fix adds a `dict_lock` around the duplicate-key 
check and write in `load_and_create_model_orig` so both operations are atomic 
under the existing `ThreadPoolExecutor`. Includes a regression test that 
confirms the duplicate is detected deterministically. All 7 tests pass on 
Kaggle TPU infrastructure.

PR #1632, opened 2026-06-30
Verify model supports segment-ids

yatlas20098:


tunix/rl/common.py
```diff
@@ -568,12 +568,25 @@ def compute_score(
       segment_positions,
   )
 
+  try:
```
tianshub: worth to move it out as a helper function and cache it? so it's not part of the core logic.

e.g. 
```
@functools.cache
def call_contains(model, target_arg):
  ...
```

yatlas20098: I've implemented caching but key'd it based on the model type instead of the model instance. Since nnx.merge constructs the model instance during each computation step, caching the model instance directly would result in zero hits.

PR #1635, opened 2026-06-30
Fix degenerate group masking for multi-group batching in GRPO learner

AntonyMei:
Fixes a bug where degenerate group masking was applied incorrectly to multi-group batches (e.g. when process_in_consumer=True), resulting in either no masking or incorrectly masking out the entire batch. Adds a unit test for cases where the input contain multiple groups.

**Problem**

See also: https://b.corp.google.com/issues/529897561

When process_in_consumer is enabled, the GRPO learner's _process_results method receives a concatenated batch containing trajectories from multiple generation groups.

The previous degenerate group masking implementation:

1. Checked for degeneracy across the entire batch at once via jnp.all(jnp.isclose(advantages, 0.0)). This meant groups were only masked if all groups in the entire batch had all-0 advantages.
2. When triggered, it masked out the entire batch using completion_mask = jnp.zeros_like(completion_mask) instead of masking only the specific degenerate group.

The correct implementation should check and apply masking at group level.

**Solution**

1. Reshaped advantages to [num_groups, num_generations] to evaluate each group's degeneracy independently.
2. Replaced jnp.all checking with a per-group check on the reshaped advantages.
3. Created a boolean mask that is expanded/repeated num_generations times to selectively mask only the sequences corresponding to degenerate groups inside completion_mask.
4. Group IDs for any filtered degenerate groups are logged individually.

**Checklist**

### DIFF
--- a/tests/rl/agentic/agentic_grpo_learner_test.py
+++ b/tests/rl/agentic/agentic_grpo_learner_test.py
@@ -996,6 +996,95 @@ class AgenticGrpoLearnerTest(parameterized.TestCase):
       [res_group4] = learner._process_results(group4)
       self.assertTrue(jnp.any(res_group4.old_per_token_logps == 0.0))
 
+  def test_process_results_masks_zero_advantage_multi_group(self):
+    class MockTraj:
+
+      def __init__(self, index, group_id, reward):
+        self.traj = {
+            "conversation_text": [
+                {"role": "user", "content": "user query"},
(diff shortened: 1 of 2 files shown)


wang2yn84:
Thank you Yixuan for this PR! I'm removing the degen in https://github.com/google/tunix/pull/1635, so this fix is no longer needed.

AntonyMei:
Thank you! I will close this PR.

Issue #1685, opened 2026-07-15
Adopt OpenTelemetry behind the MetricsLogger interface


erain:
## Problem

Tunix currently couples `MetricsLogger` to Metrax backends through process-global JAX monitoring listeners. Backend configuration is backend-specific and lives inside a training-library object, and there is no standard path for correlated metrics, traces, and structured events. [OpenTelemetry](https://opentelemetry.io/) would give Tunix a vendor-neutral instrumentation layer.

## Updated plan: phased adoption

*(Edited: the original proposal was a full replacement of the Metrax backends. Per maintainer feedback, we are not ready to fully commit to OpenTelemetry yet — we want to try it first. The plan is therefore split into phases so each step can be evaluated before the next.)*

### Phase 1 — opt-in double-write (#1686)

Add an experimental, **off-by-default** flag `MetricsLoggerOptions.enable_opentelemetry`:

- The existing Metrax/`jax.monitoring` backends stay the default and are completely unchanged. With the flag unset, behavior is identical to today and OpenTelemetry does not even need to be installed (it is a new optional extra: `pip install 'google-tunix[otel]'`, not a core dependency).
- With the flag set, every scalar passed to `MetricsLogger.log()` is **double-written**: once through the existing backends, and once as an OpenTelemetry gauge with stable `tunix.*` instrument names, low-cardinality `tunix.metrics.prefix` / `tunix.training.mode` attributes, and a separate `tunix.training.step` gauge for the logical step.
- OpenTelemetry providers and exporters are application-owned. Tunix never configures exporters and never flushes or shuts down providers; a provider can also be injected for tests/embedding.
- **Weights & Biases compatibility**: W&B natively ingests OTLP traces (Weave endpoint) but has no OTLP endpoint for run metrics, so Phase 1 ships `tunix.sft.otel_wandb.WandbMetricsExporter` — an OpenTelemetry SDK metric exporter that forwards the gauges to `wandb.log` using the familiar `{prefix}/{mode}/{name}` chart keys. Since the Metrax `WandbBackend` also keeps running, both pipelines can be compared side by side in the same W&B project.

### Phase 2 — spans and structured events (follow-up)

Once double-write metrics look good: add `span()` / `event()` helpers to the same facade, instrument high-value operations (training step, RL rollout), and fix shared-logger ownership so a logger injected into actor and critic trainers is not closed by each trainer.

### Phase 3 — switch (only after validation)

[...]

PR #1686, opened 2026-07-15
[Tunix] Add opt-in OpenTelemetry double-write to MetricsLogger

erain:
Part of #1685 (Phase 1 of the phased plan).

## Summary

This PR adds an **opt-in OpenTelemetry double-write path** to `MetricsLogger`. It deliberately does *not* replace the existing Metrax/`jax.monitoring` backends — they remain the default and are unchanged. The goal of this step is to let us try OpenTelemetry alongside the current pipeline and compare both outputs (including in Weights & Biases) before deciding on any switch.

- New flag `MetricsLoggerOptions.enable_opentelemetry` (default `False`). With the flag unset, behavior is byte-for-byte identical to today and OpenTelemetry does not need to be installed.
- With the flag set, every scalar passed to `MetricsLogger.log()` is written twice: once through the existing backends via `jax.monitoring.record_scalar` (unchanged), and once as an OpenTelemetry gauge measurement.
- New `tunix.sft.otel_wandb.WandbMetricsExporter` bridges the OpenTelemetry metrics to a Weights & Biases run, since W&B has no OTLP endpoint for run metrics (its native OTLP ingestion is traces-only, via Weave).
- OpenTelemetry is a new optional extra — `pip install 'google-tunix[otel]'` — not a core dependency. Core dependencies are unchanged (`google-metrax` stays).

### Double-write path

`log()` keeps its exact existing behavior (local history + `jax.monitoring.record_scalar`) and then, only when the flag is enabled, additionally emits through an OpenTelemetry `MeterProvider`:

- Known metrics map to stable instrument names:

  | Tunix metric key | OpenTelemetry instrument |
  | --- | --- |
  | `loss` | `tunix.training.loss` |
  | `perplexity` | `tunix.training.perplexity` |
  | `learning_rate` | `tunix.training.learning_rate` |
  | `grad_norm` | `tunix.training.gradient.norm` |

  Other metric keys are normalized into the `tunix.*` namespace (`rewards/score mean` → `tunix.rewards.score.mean`).
- The legacy prefix and mode become low-cardinality attributes (`tunix.metrics.prefix`, `tunix.training.mode`). The logical step is emitted as a separate `tunix.training.step` gauge rather than an attribute, to avoid unbounded time-series cardinality.
- Synchronous gauges are used because these are non-monotonic current values.
- Emission is best effort: conversion/emission failures are logged and can never interrupt training; non-scalar values stay in local history with a warning.
- Matching the Metrax backend policy, only JAX process 0 emits.

### Provider ownership

Tunix never configures exporters and never flushes or shuts down providers. Applications configure the global `MeterProvider`, or inject one through the new keyword-only `otel_meter_provider` constructor argument (used by tests). `MetricsLogger.close()` behavior is unchanged: it closes the Metrax backends and leaves OpenTelemetry providers running.

### Weights & Biases interop

`WandbMetricsExporter` is an OpenTelemetry SDK `MetricExporter` that groups gauge data points by prefix/mode, resolves the logical step from the `tunix.training.step` gauge, and forwards each group to `wandb.log` with the familiar `{prefix}/{mode}/{name}` chart keys (e.g. `actor/train/tunix.training.loss`), in ascending step order. Because the Metrax `WandbBackend` keeps running by default, both pipelines can be pointed at the same W&B project and compared chart by chart.

## Non-goals (deferred to later phases of #1685)

- Replacing or deprecating the Metrax backends, `backend_kwargs`, or any existing option field — nothing about the default path changes.
- Spans / structured events and trainer instrumentation (Phase 2).
- Trainer/cluster logger-ownership changes (Phase 2).
- Making OpenTelemetry a core dependency or configuring any SDK/exporter inside Tunix.

## Validation

```shell
python -m pytest tests/sft/metrics_logger_test.py tests/sft/otel_wandb_test.py -q
```

- All pre-existing `MetricsLogger` tests pass unchanged (default-off path is untouched).
- New in-memory OpenTelemetry SDK tests cover: double-write emitting both paths, default-off emitting nothing, instrument-name normalization, non-scalar handling, step-gauge dedup, process-zero policy, the missing-API `ImportError`, and provider lifecycle across `close()`.
- New `WandbMetricsExporter` unit tests cover: logger → provider → exporter → `wandb.log` flow with correct keys and step, ascending step ordering, export failure isolation (returns `FAILURE`, never raises), and lazy resolution of the global `wandb.run`.
- A real-client **end-to-end test** (`WandbOfflineEndToEndTest`) double-writes several training steps and an eval metric through the actual `wandb` package using `wandb.init(mode="offline")` — CI-safe, no account or network needed — then reads the run's `.wandb` transaction log back with wandb's own reader and asserts the per-step history (e.g. step 3 carries `actor/train/tunix.training.loss`, the normalized `actor/train/tunix.rewards.score.mean`, and `actor/eval/tunix.training.loss`, all step-aligned). A legacy Metrax-protocol backend registered on the same logger receives every scalar in the same run, demonstrating both sides of the double-write against real clients. The test skips gracefully when `wandb` is not installed; `wandb` was added to the `[test]` extra for it.
- Trainer, cluster, docs-example, and YAML config surfaces are untouched relative to `main`.

**Reference**

- Issue: #1685
- OpenTelemetry Python documentation: https://opentelemetry.io/docs/languages/python/
- W&B OTLP (traces-only) ingestion: https://docs.wandb.ai/weave/guides/tracking/otel

**Colab Notebook**

Not applicable. The default usage surface is unchanged; the opt-in flag and W&B exporter are documented in `docs/metrics.md` and covered by in-memory SDK tests.

**Checklist**

- [x] I have verified that my change does not break existing code and all applicable unit tests pass.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -210,6 +210,90 @@ depends on the execution environment.
     *   `log_dir`: Directory where event files are written.
     *   `flush_every_n_steps`: Frequency of flushing logs to disk (default: 100).
 
+### Experimental: OpenTelemetry double-write
+
+Tunix is evaluating [OpenTelemetry](https://opentelemetry.io/) as a
+vendor-neutral instrumentation layer. As an opt-in first step, `MetricsLogger`
+supports a **double-write** mode: every scalar passed to `log()` is emitted
+both through the existing Metrax/`jax.monitoring` backends (unchanged, still
+the default) and as an OpenTelemetry gauge measurement. This lets you compare
+the two pipelines side by side before any switch is made.
+
+Enable it with the `enable_opentelemetry` flag (requires
+`pip install 'google-tunix[otel]'`):
+
+```python
+# OTLP exporter: pip install opentelemetry-exporter-otlp
+from opentelemetry.exporter.otlp.proto.grpc.metric_exporter import (
+    OTLPMetricExporter,
+)
+from opentelemetry.sdk.metrics import MeterProvider
+from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
+from tunix.sft import metrics_logger
+
+reader = PeriodicExportingMetricReader(OTLPMetricExporter())
+provider = MeterProvider(metric_readers=[reader])
+
+options = metrics_logger.MetricsLoggerOptions(
+    log_dir="/tmp/logs",          # Existing backends keep working unchanged.
+    enable_opentelemetry=True,    # Additionally emit OpenTelemetry gauges.
+)
+logger = metrics_logger.MetricsLogger(options, otel_meter_provider=provider)
+```
+
+Key properties of the double-write path:
+
+*   **Off by default.** With the flag unset, behavior is identical to previous
+    releases; the OpenTelemetry API does not even need to be installed.
+*   **Application-owned providers.** Tunix never configures exporters or shuts
+    down providers. Configure the global `MeterProvider` (or inject one via the
+    keyword-only `otel_meter_provider` argument) and own its flush/shutdown.
+    `MetricsLogger.close()` only closes the Metrax backends.
+*   **Process policy.** Like the Metrax backends, only JAX process 0 emits.
+*   **Naming.** Known metrics map to stable instrument names
+    (`loss` → `tunix.training.loss`, `perplexity` →
+    `tunix.training.perplexity`, `learning_rate` →
+    `tunix.training.learning_rate`, `grad_norm` →
+    `tunix.training.gradient.norm`); other metric keys are normalized into the
+    `tunix.*` namespace (e.g. `rewards/score mean` →
+    `tunix.rewards.score.mean`). The legacy prefix and mode become the
+    low-cardinality attributes `tunix.metrics.prefix` and
+    `tunix.training.mode`, and the logical step is emitted as a separate
+    `tunix.training.step` gauge.
+
+#### OpenTelemetry → Weights & Biases
+
+W&B ingests OpenTelemetry traces natively (via the
+[Weave OTLP endpoint](https://docs.wandb.ai/weave/guides/tracking/otel)), but
+has no OTLP endpoint for run metrics. Tunix therefore ships
+`tunix.sft.otel_wandb.WandbMetricsExporter`, an OpenTelemetry SDK metric
+exporter that forwards the gauges above to `wandb.log`, using the familiar
+`{prefix}/{mode}/{name}` chart keys (e.g. `actor/train/tunix.training.loss`)
+and the `tunix.training.step` gauge as the W&B step:
+
+```python
+import wandb
+from opentelemetry.sdk.metrics import MeterProvider
+from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
+from tunix.sft import metrics_logger
+from tunix.sft import otel_wandb
+
+run = wandb.init(project="my-project")
+reader = PeriodicExportingMetricReader(
+    otel_wandb.WandbMetricsExporter(run), export_interval_millis=10_000
+)
+provider = MeterProvider(metric_readers=[reader])
+
+options = metrics_logger.MetricsLoggerOptions(
+    log_dir="/tmp/logs",
+    enable_opentelemetry=True,
+)
+logger = metrics_logger.MetricsLogger(options, otel_meter_provider=provider)
+```
+
+Because the Metrax `WandbBackend` also remains active by default, you can point
+both pipelines at the same W&B project and compare the charts directly.
+
 ### Custom metric logger
 
 You can integrate any logging service by creating a custom backend that conforms
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,9 +69,16 @@ test = [
   "pytest-xdist",
   "gcsfs",
   "chex",
+  "opentelemetry-sdk>=1.25.0",
+  "wandb",  # Offline end-to-end test for the OpenTelemetry W&B exporter
 ]
 cli = [
   "tensorflow",
+]
+otel = [
+  # Experimental OpenTelemetry double-write path for MetricsLogger.
+  "opentelemetry-api>=1.25.0",
+  "opentelemetry-sdk>=1.25.0",
 ]
 frozenlake = [
   "gymnasium",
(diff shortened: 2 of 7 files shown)


erain:
Reworked this PR from a full backend replacement into a much smaller **opt-in double-write** step, per the discussion on #1685 (the issue description has been updated with the phased plan):

- The Metrax/`jax.monitoring` backends are now completely untouched and remain the default — with the new `MetricsLoggerOptions.enable_opentelemetry` flag unset (the default), the diff is purely additive and behavior is identical to `main`.
- With the flag set, every scalar is written through **both** pipelines, so they can be compared side by side.
- Added `tunix.sft.otel_wandb.WandbMetricsExporter` so the OpenTelemetry path can feed a Weights & Biases run (W&B has no OTLP endpoint for run metrics), using the familiar `{prefix}/{mode}/{name}` chart keys.
- OpenTelemetry moved out of core dependencies into a new optional `[otel]` extra; `google-metrax` stays.
- Dropped the span/event helpers, trainer instrumentation, and logger-ownership changes from this PR — deferred to Phase 2 of #1685.

Force-pushed as a single commit; the PR description is rewritten accordingly.

wang2yn84:
Hi @erain , you should not have "Co-Author: <no****y​@anthropic.com>" in your commit. Please fix it and push the commit. The CLA issue should be gone.

erain:
(Address offline discussion "Thanks! I see it's testing with mock. I think we can run a simple script in 3p environment and see how it work.")

Good point — one clarification first: the end-to-end test in the PR is not mock-based. `WandbOfflineEndToEndTest` runs the real `wandb` client via `wandb.init(mode="offline")` (so CI needs no credentials) and verifies the run's actual transaction log; only the exporter *unit* tests use a fake run.

That said, you're right that it doesn't show the real UI result, so here is a simple script run in a plain external environment against live wandb.ai. It writes **both pipelines into a single W&B run** so the charts can be compared side by side: the legacy Metrax `WandbBackend` (`actor/train/loss`-style keys) and the new OpenTelemetry double-write path through `WandbMetricsExporter` (`actor/train/tunix.training.loss`-style keys).

Live run report : https://api.wandb.ai/links/erain-google/60767dru

Reading the run back through the W&B cloud API confirms 50 history rows (steps 1–50) with the two pipelines step-aligned and numerically identical in every row. Step 25, for example:

| legacy key | value | OpenTelemetry key | value |
| --- | --- | --- | --- |
| `actor/train/loss` | 0.96767 | `actor/train/tunix.training.loss` | 0.96767 |
| `actor/train/perplexity` | 2.63180 | `actor/train/tunix.training.perplexity` | 2.63180 |
| `actor/train/learning_rate` | 7.77821e-05 | `actor/train/tunix.training.learning_rate` | 7.77821e-05 |
| `actor/train/grad_norm` | 1.52027 | `actor/train/tunix.training.gradient.norm` | 1.52027 |
| `actor/train/rewards/score mean` | 0.5 | `actor/train/tunix.rewards.score.mean` | 0.5 |

The eval series lands under `actor/eval/...` in both pipelines as well (final `actor/eval/loss` = `actor/eval/tunix.training.loss` = 0.29087).

<details>
<summary>The script (runs with <code>pip install 'google-tunix[otel]' wandb</code> + <code>wandb login</code>)</summary>

```python
"""Demo: Tunix MetricsLogger double-write to a live Weights & Biases run."""

import numpy as np
from opentelemetry.sdk.metrics import MeterProvider
from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
import wandb

from tunix.sft import metrics_logger
from tunix.sft import otel_wandb

run = wandb.init(
    project="tunix-otel-double-write-demo",
    name="double-write-demo",
)

# New path: application-owned OpenTelemetry provider exporting to the run.
reader = PeriodicExportingMetricReader(
    otel_wandb.WandbMetricsExporter(run), export_interval_millis=5_000
)
provider = MeterProvider(metric_readers=[reader])

# Legacy path: the stock Metrax WandbBackend, unchanged.
options = metrics_logger.MetricsLoggerOptions(
    log_dir="/tmp/tunix-double-write-demo",
    backend_kwargs={
        "custom_backend": [
            lambda: metrics_logger.WandbBackend(
                project="tunix-otel-double-write-demo"
            )
        ]
    },
    enable_opentelemetry=True,
)
logger = metrics_logger.MetricsLogger(options, otel_meter_provider=provider)

# Simulate a short training run.
rng = np.random.default_rng(42)
loss = 2.5
for step in range(1, 51):
  loss = max(0.05, loss * 0.96 + rng.normal(0.0, 0.015))
  logger.log("actor", "loss", loss, metrics_logger.Mode.TRAIN, step)
  logger.log(
      "actor", "perplexity", float(np.exp(loss)), metrics_logger.Mode.TRAIN,
      step,
  )
  logger.log(
      "actor", "learning_rate", 1e-4 * 0.99**step, metrics_logger.Mode.TRAIN,
      step,
  )
  logger.log(
      "actor", "grad_norm", float(abs(rng.normal(1.5, 0.3))),
      metrics_logger.Mode.TRAIN, step,
  )
  logger.log(
      "actor", "rewards/score mean", min(1.0, 0.02 * step),
      metrics_logger.Mode.TRAIN, step,
  )
  if step % 10 == 0:
    logger.log("actor", "loss", loss * 1.08, metrics_logger.Mode.EVAL, step)
  provider.force_flush()

provider.force_flush()
run_url = run.url
provider.shutdown()  # final export while the run is still open
logger.close()  # closes the legacy backend, which finishes the run

print("\nW&B run URL:", run_url)
```

</details>

One practical note from setting this up: W&B has retired anonymous mode (the 0.28 client removed it and the server now rejects anonymous key creation), so any wandb path — legacy or OpenTelemetry — requires `wandb login` in external environments.

Happy to also put together a Colab that runs a real `PeftTrainer` loop (rather than synthetic `log()` calls) with the flag enabled, if that would be useful for the evaluation.

erain:
Screenshots

image

image

sizhit2:
Thank you @erain to add the end-2-end testing. I couldn't access your wandb runs since your wandb project is private. Could you create w&b report so I can see the actual runs?

Also, the PR cannot be merged because some TPU checks are failing. I think it's due to wandb setup. Could you take a look and fix it?

sizhit2 (approved):
Oops, I just saw your screenshot.

PR #1698, opened 2026-07-17
GRPO: pluggable per-token log-prob hook (for non-autoregressive / diffusion rollouts)

ethannnnnn:
## Summary
Adds an optional `per_token_logps_fn` hook to `GRPOLearner` / `grpo_loss_fn` so a non-autoregressive rollout (e.g. a block-diffusion / masked-diffusion policy) can supply its own trajectory log-probability instead of the built-in autoregressive next-token computation. Default behavior is unchanged.

## What's in this PR
- `GRPOLearner.__init__(..., per_token_logps_fn=None)` (`rl/grpo/grpo_learner.py`): stored and forwarded into `grpo_loss_fn` only when set.
- `grpo_loss_fn(..., per_token_logps_fn=None)`: selects the log-prob function — the injected one if provided, otherwise the existing `common.compute_per_token_logps` (autoregressive next-token shift).
- The clipped surrogate, advantage estimation, KL, and optimizer are untouched.

## Motivation
Diffusion / masked-diffusion LMs don't have a simple AR per-token log-prob; their trajectory log-prob must be computed by the model owner. This hook lets an external rollout provide it while reusing all of GRPO/DAPO's machinery (the corresponding rollout lives in the caller).

## Back-compat
Fully backward-compatible: when `per_token_logps_fn` is not provided, the AR path runs exactly as before.

## Testing (CPU)
- Default routes to `common.compute_per_token_logps` (AR), not the override.
- Override routes to the injected function, not the default.
- Identical log-probs → identical loss (surrogate untouched); KL path covered (6 tests).

### DIFF
--- a/tunix/rl/grpo/grpo_learner.py
+++ b/tunix/rl/grpo/grpo_learner.py
@@ -17,7 +17,7 @@
 from __future__ import annotations
 
 import dataclasses
-from typing import Iterable, List, Sequence, TypeVar
+from typing import Callable, Iterable, List, Sequence, TypeVar
 
 import flax
 from flax import nnx
@@ -126,6 +126,7 @@ class GRPOLearner(rl_learner.RLLearner[TGrpoConfig]):
       reward_fns: RewardFn | List[RewardFn],
       metric_fns: Sequence[MetricFn] | None = None,
       data_shuffle_seed: int | None = None,
+      per_token_logps_fn: Callable[..., jax.Array] | None = None,
   ):
     """Initializes the `GRPOTrainer`.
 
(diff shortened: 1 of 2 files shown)


tunix/rl/grpo/grpo_learner.py
```diff
@@ -132,6 +126,7 @@ def __init__(
       reward_fns: RewardFn | List[RewardFn],
       metric_fns: Sequence[MetricFn] | None = None,
       data_shuffle_seed: int | None = None,
+      per_token_logps_fn: Callable[..., jax.Array] | None = None,
```
tianshub, line 129: @ethannnnnn  thanks for adding the support, you may want to enable this directly in [agentic_grpo_learner](https://github.com/google/tunix/blob/9b24d9dfa9e9892116d59f617c4dc54f1d1a8ce2/tunix/rl/agentic/agentic_grpo_learner.py#L187)  since it's the new learner we've been using.

PR #1703, opened 2026-07-17
Fix docker build and add training dependencies (kubernetes, r2e-gym)

niting:
Add relevant changes to get DeepSWE training to work on the Tunix stack. This includes dependency additions and fixes to import failures on examples.deepswe


Dockerfile
```diff
@@ -36,6 +36,8 @@ COPY . .
 RUN pip install -e .
 
 RUN bash /app/scripts/install_tunix_vllm_requirement.sh
+RUN pip install kubernetes
+RUN pip install --no-deps git+https://github.com/r2e-gym/r2e-gym.git@0d94c4eb9431cd195c55a7ea3abd54006c9a1735
```
wang2yn84: Let's make it optional instead of default

Dockerfile
```diff
@@ -7,7 +7,7 @@ ENV TZ=Etc/UTC
 
 # Install system dependencies, including Python 3 and pip
 RUN apt-get update && \
-    apt-get install -y build-essential git python3 python3-pip && \
+    apt-get install -y build-essential git python3 python3-pip curl && \
```
wang2yn84: If you replace the curl install of uv with pip, do you still need curl here?

PR #1740, opened 2026-07-23
Fix dependency errors and deprecated JAX syntax in qlora_gemma notebook

selamw1:
**Overview**
This PR resolves missing module errors, updates deprecated JAX syntax.

**Changes Made**

- Switched pip install `tensorboardX` to `tensorboard` to fix the `ModuleNotFoundError` while `%load_ext tensorboard.`
- Changed the installation check from `dotenv` to `tensorflow` to guarantee required libraries are installed.
- Replaced the deprecated with `mesh:` context manager with with `jax.set_mesh(mesh):`.


sizhit2 (approved):
Because the change is too big, I can't see from Github. I will approve it for it to generate CL internal and review from there.

PR #1832, opened 2026-08-05
[tunix] Add model-agnostic JAX diffusion contracts

ethannnnnn:
## Motivation

Diffusion training integrations need a small, stable JAX boundary that does
not depend on MaxText, a specific model, or a training algorithm.

## Scope

This PR adds only:

- a batch-major `DiffusionTokenBatch` with target-aligned token IDs and
  explicit loss weights;
- typed raw-batch adapter and logits-function protocols;
- shape, dtype, eager-value, and scorer-output validation; and
- focused contract tests and package documentation.

It does not add an SFT or distillation objective, change gradient reduction,
prepare a rollout, alter RL behavior, or import MaxText.

## Design

Every model-input leaf has the same leading batch axis. Static model
configuration remains in the logits-function closure rather than becoming a
batch leaf. Targets refer to physical prediction positions, so model-specific
same-position or shifted-logit alignment happens before constructing the
canonical batch.

`DiffusionLogitsFn` receives an NNX model and prepared model inputs and returns
floating-point logits with shape `[batch, length, vocabulary]`. Construction
and scoring boundaries validate array structure, target alignment, and active
values. Static checks remain safe under JIT and sharding; concrete finite and
range checks run when arrays are eagerly addressable.

## Compatibility

The package is new and opt-in. Existing SFT, distillation, RL, and
autoregressive paths do not construct these contracts and remain unchanged.

## Performance and Quality

This contracts-only PR does not add a diffusion training or inference runtime,
so it has no direct performance or quality delta relative to the existing
autoregressive path. The package README now compares their execution and
quality tradeoffs and defines the evaluation protocol for implementation PRs:
matched model, data, training-compute, hardware, batch, sequence, and decoding
budgets; latency and throughput for performance; and fixed tasks and seeds for
quality. It also calls out that autoregressive perplexity and diffusion
training loss are not directly interchangeable.

## Tests

```text
python -m pytest -q tests/diffusion/types_test.py
15 passed
```

Pyink, isort, scoped Pylint, Python compilation, and `git diff --check` also
pass for the changed files.

## Follow-Ups

Denominator-aware weighted reduction will be proposed separately after this
contract lands. Diffusion SFT, prepared distillation, OPD, rollout transport,
policy scoring, and GRPO integration remain later independent reviews.

Design document:

### DIFF
--- a/tunix/diffusion/__init__.py
+++ b/tunix/diffusion/__init__.py
@@ -1,0 +1,31 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Model-agnostic JAX contracts for diffusion training."""
+
+# pylint: disable=g-importing-member
+
+from tunix.diffusion.interfaces import compute_diffusion_logits
+from tunix.diffusion.interfaces import DiffusionBatchAdapter
+from tunix.diffusion.interfaces import DiffusionLogitsFn
+from tunix.diffusion.types import DiffusionTokenBatch
+from tunix.diffusion.types import validate_diffusion_logits
+
+__all__ = [
+    "DiffusionBatchAdapter",
+    "DiffusionLogitsFn",
+    "DiffusionTokenBatch",
+    "compute_diffusion_logits",
+    "validate_diffusion_logits",
+]
--- a/tunix/diffusion/interfaces.py
+++ b/tunix/diffusion/interfaces.py
@@ -1,0 +1,60 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Callable interfaces for model-agnostic JAX diffusion integrations."""
+
+from typing import Protocol, TypeVar
+
+from flax import nnx
+from tunix.diffusion import types
+
+RawBatchT_contra = TypeVar("RawBatchT_contra", contravariant=True)
+
+
+class DiffusionBatchAdapter(Protocol[RawBatchT_contra]):
+  """Converts an external batch to Tunix's canonical diffusion batch."""
+
+  def __call__(self, batch: RawBatchT_contra, /) -> types.DiffusionTokenBatch:
+    """Prepares a raw batch as target-aligned diffusion arrays."""
+    ...
+
+
+class DiffusionLogitsFn(Protocol):
+  """Computes target-aligned token logits for a diffusion model."""
+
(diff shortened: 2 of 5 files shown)


tunix/diffusion/types.py
```diff
@@ -0,0 +1,200 @@
...
+import jax
+import jax.numpy as jnp
+import numpy as np
+
+Array: TypeAlias = jax.Array | np.ndarray
+ModelInputs: TypeAlias = Mapping[str, Any]
+
+
+def _shape_and_dtype(name: str, value: Any) -> tuple[tuple[int, ...], Any]:
+  if not isinstance(value, (jax.Array, jax.core.Tracer, np.ndarray)):
```
wang2yn84: Should consider jax.ShapeDtypeStruct if run jax.eval_shape()

tunix/diffusion/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+from tunix.diffusion.interfaces import compute_diffusion_logits
+from tunix.diffusion.interfaces import DiffusionBatchAdapter
+from tunix.diffusion.interfaces import DiffusionLogitsFn
+from tunix.diffusion.types import DiffusionTokenBatch
+
+__all__ = [
+    "DiffusionBatchAdapter",
+    "DiffusionLogitsFn",
+    "DiffusionTokenBatch",
+    "compute_diffusion_logits",
```
wang2yn84, line 29: The design do explicitly lists validate_diffusion_logits as one of the primary public API functions. Consider add here

tunix/diffusion/types.py
```diff
@@ -0,0 +1,200 @@
...
+from collections.abc import Mapping
+from typing import Any, TypeAlias
+
+import flax
+import jax
+import jax.numpy as jnp
+import numpy as np
+
+Array: TypeAlias = jax.Array | np.ndarray
+ModelInputs: TypeAlias = Mapping[str, Any]
```
wang2yn84: Consider  jaxtyping.PyTree based on the doc

tunix/diffusion/interfaces.py
```diff
@@ -0,0 +1,53 @@
...
+  """Converts an external batch to Tunix's canonical diffusion batch."""
+
+  def __call__(self, batch: RawBatchT_contra, /) -> types.DiffusionTokenBatch:
+    ...
+
+
+class DiffusionLogitsFn(Protocol):
+  """Computes target-aligned token logits for a diffusion model."""
+
+  def __call__(
```
wang2yn84, line 36: It'll be useful if you can put a docstring for __call__ function too. You already have compute_diffusion_logits, not sure what __call__ does.

tunix/diffusion/types.py
ethannnnnn: Done. `Array` and `_shape_and_dtype` now accept `jax.ShapeDtypeStruct`, and abstract values skip eager host checks. A regression test round-trips a `DiffusionTokenBatch` through `jax.eval_shape` and validates abstract logits.

tunix/diffusion/__init__.py
ethannnnnn: Done. `validate_diffusion_logits` is now imported and listed in `tunix.diffusion.__all__`; the abstract-logits test calls it through the public package surface.

tunix/diffusion/types.py
ethannnnnn: Done. `ModelInputs` is now `jaxtyping.PyTree[Array]`, the mapping-only runtime restriction is removed, and a nested tuple/dict PyTree test covers the generic contract.

tunix/diffusion/interfaces.py
ethannnnnn: Done. Both protocol `__call__` methods now have method-level docstrings. `DiffusionLogitsFn.__call__` is documented as the model-specific scorer returning unvalidated target-aligned `[batch, length, vocabulary]` logits; `compute_diffusion_logits` owns validation.

ethannnnnn:
Thanks for the suggestion. I added a performance and quality comparison with the existing autoregressive path to both the PR description and `tunix/diffusion/README.md`. Because this PR defines contracts only and does not add a diffusion training or inference runtime, it has no direct benchmark delta. The new section instead documents the execution and quality tradeoffs and the matched evaluation protocol that implementation PRs should use, without claiming unsupported gains.

wang2yn84 (approved):

PR #1833, opened 2026-08-05
[tunix] Preserve fractional weighted-gradient denominators

ethannnnnn:
## Motivation

`LossOutput` carries an unreduced numerator and its denominator so gradient
accumulation can preserve the weighting of every token or example. Tunix now
accumulates numerator gradients using the global denominator, but
`GradientAccumulator.get()` still clamps that denominator to at least one.
That under-scales valid fractional weights. The depth-one fast path also
divides directly by zero for an empty weighted batch.

## Scope

This PR changes only the remaining denominator edge cases in the generic SFT
trainer and its focused tests:

- preserve every positive fractional denominator exactly;
- return zero gradients when the accumulated denominator is zero;
- apply the same rule to the depth-one fast path; and
- cover fractional and zero denominators in both execution paths.

It does not add diffusion contracts, change metric logging or reduction,
introduce an SFT objective, or alter RL behavior.

## Design

For microbatches `i`, Tunix already computes

```text
sum_i grad(numerator_i) / sum_i denominator_i
```

This PR replaces `1 / max(sum_i denominator_i, 1)` with a zero-safe reciprocal:
positive denominators retain their exact value, while a zero total denominator
produces an all-zero gradient tree without evaluating a divide-by-zero path.
Legacy scalar losses still contribute denominator `1`, so their behavior is
unchanged.

## Compatibility

The behavior changes only when a loss function returns `utils.LossOutput`.
Plain scalar losses retain the existing gradient-accumulation semantics. The
public `LossOutput` and `WeightedMetric` data contracts are unchanged.

## Tests

```text
python -m pytest -q tests/sft/peft_trainer_test.py
```

The guarded publisher also runs changed-line Pyink, scoped error/fatal Pylint,
Python compilation, and `git diff --check` on the two changed files.

## Follow-Ups

Global reduction of weighted train and evaluation metrics is intentionally a
separate review. Diffusion training and RL integrations remain later layers.

Design document:

### DIFF
--- a/tunix/sft/peft_trainer.py
+++ b/tunix/sft/peft_trainer.py
@@ -167,6 +167,10 @@ def _calculate_global_batch_size(train_example: Any) -> int:
   )
 
 
+def _zero_safe_reciprocal(denom: jax.Array) -> jax.Array:
+  return jnp.where(denom == 0, jnp.zeros_like(denom), 1.0 / denom)
+
+
 class GradientAccumulator(nnx.Module):
   """Accumulates gradients over multiple micro-steps.
 
@@ -261,7 +265,7 @@ class GradientAccumulator(nnx.Module):
     self.denom.set_value(self.denom[...] + denom_val)
 
   def get(self):
-    scale = 1.0 / jnp.maximum(self.denom[...], jnp.asarray(1.0, jnp.float32))
+    scale = _zero_safe_reciprocal(self.denom[...])
 
     def _scale_and_cast(v, target_dtype):
       res = v[...] * scale.astype(v[...].dtype)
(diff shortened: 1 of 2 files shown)


tunix/sft/peft_trainer.py
```diff
@@ -167,6 +167,11 @@ def _calculate_global_batch_size(train_example: Any) -> int:
   )
 
 
+def _zero_safe_reciprocal(denom: jax.Array) -> jax.Array:
+  safe_denom = jnp.where(denom == 0, jnp.ones_like(denom), denom)
+  return jnp.where(denom == 0, jnp.zeros_like(denom), 1.0 / safe_denom)
```
wang2yn84: nit, no need for safe_denom, we already have conditional check here, so return jnp.where(denom == 0, jnp.zeros_like(denom), 1.0 /denom) should be fine.

ethannnnnn: Done. `_zero_safe_reciprocal` now returns the zero-guarded `jnp.where` expression directly, without the intermediate `safe_denom`. The fractional- and zero-denominator tests still cover both the depth-one and accumulator-backed paths, and all 70 `PeftTrainer` tests pass.

PR #1839, opened 2026-08-05
Fix prompt batching structure for agentic rollout generation

niting:
Ensure that when rollout_micro_batch_size = 1, the input conversation
history (list of dicts) is wrapped in a list to match the expected
batch format (list of list of dicts) of the rollout generation engine.


wang2yn84 (approved):
Thank you! Looks reasonable. Wondering what kind of check fails for you?

PR #1880, opened 2026-08-13
[tunix] Preserve optimizer-state dtypes across updates

ethannnnnn:
## Motivation

Tunix commonly uses float32 optimizer state for accuracy, while also supporting
optimizer configurations that intentionally use lower-precision moments to
reduce memory. In both cases, the dtype tree selected when the optimizer is
initialized or restored must remain stable. An Optax optimizer built with
`inject_hyperparams` can instead promote some floating moment leaves during
`update`. With bfloat16 moments this silently changes the optimizer-state
contract to float32. On the legacy gradient-accumulation path it also makes the
update and skip branches of `nnx.cond` return different dtypes, so tracing
fails.

## Scope

This PR preserves the optimizer-state dtype tree across each trainer update:

- capture optimizer-state dtypes immediately before `optimizer.update`;
- restore only floating leaves whose dtype changed during the update;
- cover the legacy trainer's depth-one and conditional accumulated paths;
- apply the same invariant to the graduating `peft_trainer_v2` fused and split
  update paths; and
- test actual jitted updates for direct and injected AdamW at accumulation
  depths one and two in both trainers.

Integer counters and floating leaves that already use their configured dtype
are unchanged. A dedicated v2 test confirms that the default float32 state
remains float32. This PR does not modify model parameter dtypes, loss or metric
reduction, gradient accumulation math, checkpoint formats, or optimizer
selection.

## Dependency

This follow-up is based on `main` after #1850. The two changes are semantically
independent, but both touch `PeftTrainer` and its tests, so this patch was
rebuilt after #1850 merged rather than replaying the old preparation commit.
The guarded publisher verifies that #1850 is present before preparing or
publishing the branch.

## Tests

```text
python -m pytest -q tests/sft/peft_trainer_test.py \
  tests/experimental/train/peft_trainer_v2_test.py \
  -k optimizer_state_dtypes_under_jit
python -m pytest -q tests/sft/peft_trainer_test.py \
  tests/experimental/train/peft_trainer_v2_test.py
```

The focused matrix has nine passing cases: direct and injected AdamW at
accumulation depths one and two in both trainers, plus the default float32 v2
state. The complete trainer suites have 114 passing tests. The guarded
publisher also runs changed-range Pyink, scoped error/fatal Pylint, Python
compilation, and `git diff --check`.

Design document:

### DIFF
--- a/tests/experimental/train/peft_trainer_v2_test.py
+++ b/tests/experimental/train/peft_trainer_v2_test.py
@@ -765,6 +765,127 @@ class PeftTrainerTest(parameterized.TestCase):
     )
 
 
+class OptimizerMemoryTest(parameterized.TestCase):
+  """Optimizer-state dtype preservation across v2 update paths."""
+
+  def _bf16_model(self):
+    return nnx.Linear(
+        in_features=4,
+        out_features=2,
+        rngs=nnx.Rngs(0),
+        param_dtype=jnp.bfloat16,
+    )
+
+  def _opt_state_dtypes(self, trainer):
+    return jax.tree_util.tree_map(
+        lambda value: value[...].dtype,
+        nnx.state(trainer.optimizer, nnx.optimizer.OptState),
+        is_leaf=lambda value: isinstance(value, nnx.Variable),
+    )
+
+  def _run_update(self, trainer, inputs, gradient_accumulation_steps):
+    if gradient_accumulation_steps == 1:
+      trainer.train_step(inputs)
+      return
+
+    for _ in range(gradient_accumulation_steps):
+      trainer.fwd_bwd(inputs)
+    trainer.update()
+
+  @parameterized.product(
+      optimizer_kind=('direct', 'injected'),
+      gradient_accumulation_steps=(1, 2),
+  )
+  def test_preserves_bf16_optimizer_state_dtypes_under_jit(
+      self, optimizer_kind, gradient_accumulation_steps
+  ):
+    """Jitted fused and split updates preserve optimizer-state dtypes."""
+    model = self._bf16_model()
+    if optimizer_kind == 'direct':
+      tx = optax.adamw(
+          lambda _: jnp.asarray(0.1, dtype=jnp.float32),
+          mu_dtype=jnp.bfloat16,
+      )
+    else:
+      tx = optax.inject_hyperparams(optax.adamw, hyperparam_dtype=jnp.float32)(
+          learning_rate=0.1
+      )
+    config = peft_trainer_v2.TrainingConfig(
+        eval_every_n_steps=100,
+        max_steps=1,
+        gradient_accumulation_steps=gradient_accumulation_steps,
+    )
+    trainer = peft_trainer_v2.PeftTrainer(model, tx, config)
+    trainer.with_loss_fn(lambda m, x, y: jnp.sum((m(x) - y) ** 2))
+
+    initial_opt_state_dtypes = self._opt_state_dtypes(trainer)
+    initial_params = jax.tree_util.tree_map(
+        lambda value: jnp.copy(value[...]),
+        nnx.state(model, nnx.Param),
+        is_leaf=lambda value: isinstance(value, nnx.Variable),
+    )
+    inputs = {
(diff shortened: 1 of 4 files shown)


wang2yn84:
Thank you for the PR! In Tunix, we tend to keep optimizer state to be f32 for best accuracy. I see what you mean with this PR, it's necessary for bf16 training. Since we are also graduating peft_trainer_v2 soon, Can you also help to apply the same patch to peft_trainer_v2 too? Thank you!

ethannnnnn:
> Thank you for the PR! In Tunix, we tend to keep optimizer state to be f32 for best accuracy. I see what you mean with this PR, it's necessary for bf16 training. Since we are also graduating peft_trainer_v2 soon, Can you also help to apply the same patch to peft_trainer_v2 too? Thank you!

Thanks! I’ve applied the same dtype-preservation logic to peft_trainer_v2, covering both fused and split update paths. The implementation preserves the optimizer-selected dtype: default f32 remains f32, while explicitly configured bf16 moments remain bf16. All 114 v1/v2 trainer tests pass.

tianshub:
@ethannnnnn thanks for the fix, curious how did you spot this problem? are you trying to use bf16 for moments? (normally we keep those in fp32 for stability)

Also do you mind capture a xprof showing the before and after step time to make sure there's no regressions on performance?

ethannnnnn:
> @ethannnnnn thanks for the fix, curious how did you spot this problem? are you trying to use bf16 for moments? (normally we keep those in fp32 for stability)
> 
> Also do you mind capture a xprof showing the before and after step time to make sure there's no regressions on performance?


Thanks! I hit this during a multi-host TPU SFT smoke test with an explicitly configured low-precision optimizer state. With injected AdamW and gradient accumulation, mu was promoted from bf16 to fp32 in the update branch while the skip branch stayed bf16, so the jitted conditional failed type checking.

This PR doesn't make bf16 moments the default — it preserves the dtype tree chosen at optimizer init/restore: fp32 stays fp32, an explicitly configured bf16 state stays bf16. I added a regression test confirming the default fp32 state is unchanged, including the PeftTrainerV2 path.

I'm capturing a paired XProf now against the exact PR parent and head on the same TPU v7x allocation, using injected AdamW with default fp32 moments. I'll follow up with before/after step times and traces. fp32 is the valid no-regression comparison, since the original failing bf16 case doesn't produce a pre-fix training step to profile.

ethannnnnn:
> @ethannnnnn thanks for the fix, curious how did you spot this problem? are you trying to use bf16 for moments? (normally we keep those in fp32 for stability)
> 
> Also do you mind capture a xprof showing the before and after step time to make sure there's no regressions on performance?

 ### XProf A/B results

  **Setup**

  - Parent: `de29aaee34fc498c96313ac312278725650f2db3`
  - PR head: `9a17c3714b9d40b03fb479107bfaa245ed6b932f`
  - Same TPU v7 `4x4x4` allocation with 128 JAX devices
  - `PeftTrainerV2` with injected AdamW
  - Default fp32 optimizer moments
  - Gradient accumulation depth 2
  - 2,147,483,648 synthetic parameters
  - 5 warm-up updates and 5 XProf-captured updates per revision
  - 20 logical-update and 20 optimizer-only timing samples per
  revision
  - Each sample reports the slowest-host latency

  **Results**

  - Logical-update mean: `39.267 ms` -> `39.138 ms` (`-0.329%`)
  - Logical-update median: `38.398 ms` -> `38.830 ms` (`+1.124%`)
  - Logical-update trimmed mean: `38.949 ms` -> `38.913 ms`
  (`-0.093%`)
  - Optimizer-update median: `16.560 ms` -> `16.544 ms` (`-0.102%`)
  - Optimizer-update trimmed mean: `16.683 ms` -> `16.704 ms`
  (`+0.123%`)

  The parent optimizer samples contained one isolated `151.565 ms`
  outlier, so I used the median and trimmed mean for the comparison.

  The robust optimizer measurements differ by at most `0.123%`, and
  the complete logical-update measurements are effectively flat. I
  found no material step-time regression on the normal fp32-moment
  path.

PR #1891, opened 2026-08-14
[tunix] Add target-aligned diffusion SFT adapter

ethannnnnn:
## Motivation

Tunix now has model-agnostic diffusion batch and logits contracts, but
`PeftTrainer` still needs a small adapter and a target-aligned supervised loss
before those contracts can be used for diffusion CFT or SFT. Autoregressive
loss helpers cannot be reused because they shift targets by one token.

## Scope

This PR adds only the diffusion SFT integration layer:

- adapt an external raw batch into `DiffusionTokenBatch`;
- call a model-specific `DiffusionLogitsFn` through the public diffusion
  contract;
- compute float32 cross entropy at the same token positions as `target_ids`;
- preserve the unreduced weighted numerator and denominator expected by
  `PeftTrainer`; and
- sanitize inactive positions so zero-weight NaN, infinity, or invalid target
  values cannot contaminate the loss or gradient.

It does not add a model implementation, tokenizer, dataset recipe,
distillation objective, rollout path, or RL trainer.

## Dependencies

This PR depends on the contracts merged in #1832 and the fractional and
zero-denominator gradient behavior merged in #1833. Both dependencies are
verified against `main` by the guarded publisher before the patch is replayed.
It does not depend on #1850 or the optional OPD track.

## Compatibility

The existing autoregressive `PeftTrainer` path is unchanged. Users opt in by
calling `configure_diffusion_sft` with their batch adapter and target-aligned
logits function. CFT and SFT use the same objective; the external adapter
chooses which target positions and weights are active.

## Tests

```text
python -m pytest -q tests/diffusion/types_test.py tests/sft/diffusion_sft_test.py
python -m pytest -q tests/sft/peft_trainer_test.py
```

The focused suite covers trainer wiring, same-position alignment, float32
metrics, zero-weight numerical safety and zero gradients, jitted gradients,
microbatch equivalence, and shape rejection. The guarded publisher also runs
full `PeftTrainer` regression tests, Pyink, scoped error/fatal Pylint, Python
compilation, and `git diff --check`.

Design document:

### DIFF
--- a/tunix/sft/diffusion.py
+++ b/tunix/sft/diffusion.py
@@ -1,0 +1,93 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Supervised fine-tuning loss for target-aligned diffusion scores."""
+
+from typing import Any, TypeVar
+
+from flax import nnx
+import jax.numpy as jnp
+import optax
+from tunix.diffusion import interfaces as diffusion_interfaces
+from tunix.diffusion import types as diffusion_types
+from tunix.sft import peft_trainer
+from tunix.sft import utils as sft_utils
+
+RawBatchT = TypeVar("RawBatchT")
+TrainerT = TypeVar("TrainerT", bound=peft_trainer.PeftTrainer)
+
+
+def diffusion_loss_fn(
+    model: nnx.Module,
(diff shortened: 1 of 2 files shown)


tunix/sft/diffusion.py
```diff
@@ -0,0 +1,93 @@
...
+  """
+
+  logits = diffusion_interfaces.compute_diffusion_logits(
+      model, batch, logits_fn
+  )
+  logits = jnp.asarray(logits, dtype=jnp.float32)
+  targets = jnp.asarray(batch.target_ids)
+  weights = jnp.asarray(batch.loss_weights, dtype=jnp.float32)
+
+  active_targets = weights != 0
```
wang2yn84: nit, better be weights > 0 according to validation?

ethannnnnn: Good catch. `DiffusionTokenBatch.validate()` requires nonnegative loss weights and defines active targets with `weights > 0`, so I've updated this check to use the same predicate. This is equivalent for valid batches and keeps zero-weight positions inactive.

wang2yn84:
It's actually committed in https://github.com/google/tunix/commit/bd8092a5dd7df677d6d665abc241c26f5a348a84. Somehow this PR is not closed properly. Close it now.

PR #1904, opened 2026-08-17
Add optional process-pool evaluation of reward functions

gutianyu-google:
Sequence-level reward functions are pure Python, evaluated serially over the whole batch on one core while the host's remaining cores idle. On a GRPO post-training benchmark (qwen3-0.6b, 2048 sequences/step, 3 reward functions on a 40+ core head node), this serial pass took **29.6 s of a 155.6 s global step** — the second-largest component after rollout.

This PR adds `AlgorithmConfig.reward_num_workers`:

- **Default `0` keeps the serial implementation and preserves current behavior bit-for-bit.**
- When `> 1`, `SequenceRewardManager` evaluates each reward function over contiguous chunks of the batch in a fork-based process pool and concatenates results in order. Per-sequence reward functions therefore produce identical results to the serial path. kwargs that are per-example columns (`len == num_prompts`, e.g. ground-truth answer lists) are sliced in lockstep with the batch; everything else passes through unchanged.

Robustness, all covered by unit tests:

- Functions that cannot run in a worker — unpicklable, or spawning subprocesses of their own (which daemonic pool workers forbid) — are detected at runtime and evaluated in the parent process from then on, where their own internal parallelism keeps working.
- Pool-creation failure, worker crashes, chunk timeouts, or a platform without the `fork` start method all fall back to the serial implementation with one warning.
- `TUNIX_REWARD_PARALLEL=0` disables the pool at runtime without a config change.

**Measured** (same benchmark, `reward_num_workers=32`): reward computation 29.6 s → ~3.5 s, global step **155.6 s → 129.0 s (−17%)**, with per-step rewards byte-identical to the serial run over all 20 steps (max abs diff 0.000000). One of the three reward functions dispatches to its own subprocess pool and ran in the parent via the fallback; widening that internal pool contributed part of the win.

**Reference**
Benchmark context: MaxText GRPO post-training on TPU v7x; reward functions are the `match_format_*` / `check_numbers` family from MaxText's RL utils.

**Checklist**

### DIFF
--- a/tests/rl/reward_manager_test.py
+++ b/tests/rl/reward_manager_test.py
@@ -13,9 +13,13 @@
 # limitations under the License.
 
 import dataclasses
+import gc
 import inspect
+import multiprocessing
+import time
 from typing import Any, List
 from unittest import mock
+
 from absl import logging
 from absl.testing import absltest
 from absl.testing import parameterized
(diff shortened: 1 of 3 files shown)


tunix/rl/reward_manager.py
```diff
@@ -101,6 +106,110 @@ def __init__(
...
+    # runtime: fns that are unpicklable, or that spawn subprocesses of their
+    # own (which the pool's daemonic workers forbid).
+    self._parent_only_fns = set()
+    self._parallel_disabled = False
+
+  def _parallel_enabled(self) -> bool:
+    return (
+        getattr(self.algo_config, "reward_num_workers", 0) > 1
+        and not self._parallel_disabled
+        and os.getenv("TUNIX_REWARD_PARALLEL", "1") != "0"
```
wang2yn84: Remove it, this env should not exist

tunix/rl/algorithm_config.py
```diff
@@ -63,6 +73,11 @@ def __post_init__(self):
           f"policy_loss_fn must be one of {valid_policy_loss_fns}."
           f" Received: {self.policy_loss_fn}"
       )
+    if self.reward_num_workers < 0:
```
wang2yn84: How about upper limit check?  Maybe print a warning if it exceeds the number of CPUs? Or allow reward_num_workers to be -1 and automatically assign with number of CPUs.

tunix/rl/reward_manager.py
```diff
@@ -101,6 +106,110 @@ def __init__(
...
+      call_kwargs: Dict[str, Any],
+  ) -> Any:
+    """Evaluates one reward fn over the batch, using workers when enabled."""
+    fn_name = getattr(reward_fn, "__name__", str(reward_fn))
+    if not self._parallel_enabled() or fn_name in self._parent_only_fns:
+      return reward_fn(prompts=prompts, completions=completions, **call_kwargs)
+
+    try:
+      pool = self._get_pool()
+    except Exception as e:  # pylint: disable=broad-except
```
wang2yn84, line 179: You already captured exception in _get_pool API, no need to capture again. Only need to keep one.

tunix/rl/reward_manager.py
```diff
@@ -101,6 +106,110 @@ def __init__(
...
+    try:
+      n_chunks = max(
+          1,
+          min(pool._processes * 2, num_prompts),  # pylint: disable=protected-access
+      )
+      bounds = [num_prompts * c // n_chunks for c in range(n_chunks + 1)]
+      jobs = []
+      for c in range(n_chunks):
+        lo, hi = bounds[c], bounds[c + 1]
+        if lo == hi:
```
wang2yn84: Only happens when num_prompts is 0, which if you want to handle should check at the very beginning and error out.

tunix/rl/reward_manager.py
```diff
@@ -101,6 +106,110 @@ def __init__(
...
+        for k, v in call_kwargs.items():
+          # kwargs that are per-example columns (one entry per prompt, e.g.
+          # ground-truth answers) are sliced in lockstep with the batch;
+          # everything else passes through unchanged.
+          if isinstance(v, (list, tuple)) and len(v) == num_prompts:
+            chunk_kwargs[k] = list(v[lo:hi])
+          elif isinstance(v, np.ndarray) and len(v) == num_prompts:
+            chunk_kwargs[k] = v[lo:hi]
+          else:
+            chunk_kwargs[k] = v
```
wang2yn84, line 215: Add a comment for this condition

tunix/rl/reward_manager.py
```diff
@@ -101,6 +106,110 @@ def __init__(
...
+          " implementation.",
+          e,
+      )
+      return reward_fn(prompts=prompts, completions=completions, **call_kwargs)
+
+    num_prompts = len(prompts)
+    try:
+      n_chunks = max(
+          1,
+          min(pool._processes * 2, num_prompts),  # pylint: disable=protected-access
```
wang2yn84: Why we don't respect the user setting of number of worker here by x2?

tests/rl/reward_manager_test.py
```diff
@@ -280,5 +282,122 @@ def test_log_metrics_non_interference_no_reward_fns(self):
...
+  p.start()
+  p.join()
+  return [7.0] * len(prompts)
+
+
+child_spawning_reward.__name__ = "child_spawning_reward"
+
+
+@absltest.skipIf(
+    "fork" not in multiprocessing.get_all_start_methods(),
```
wang2yn84, line 318: Internal use g3_multiprocessing, need some special handling for this test to run on 1p.

tunix/rl/reward_manager.py
gutianyu-google: Removed; `reward_num_workers=0` is the off switch now.

tunix/rl/algorithm_config.py
gutianyu-google: Done: -1 now means one worker per CPU, and a value above the CPU count logs a warning.

tunix/rl/reward_manager.py
gutianyu-google: Done: an empty batch is checked up front and takes the serial path, and the guard is removed.

tunix/rl/reward_manager.py
gutianyu-google: The pool size is exactly `reward_num_workers`; the 2x is only chunk count, for load balance since reward cost is skewed across completions. Added a comment saying so. Can also make it one chunk too.

tests/rl/reward_manager_test.py
gutianyu-google: The test already skips when the fork start method is unavailable. Plz lemme know the best way to proceed.

gutianyu-google:
Addressed in aa3d1a6f, see the threads for details.

tunix/rl/reward_manager.py
gutianyu-google: Done, the try/except inside `_get_pool` is gone; only `_call_reward_fn` handles pool-creation failure now.

tunix/rl/reward_manager.py
gutianyu-google: Added.

tests/rl/reward_manager_test.py
wang2yn84: The principle is we want to run these tests on g3 too. So need to update the copybara to use g3_multiprocessing instead of multiprocessing. You will need to patch a cl on g3. That's what I meant.

tunix/rl/reward_manager.py
wang2yn84: I meant to say the TUNIX_REWARD_PARALLEL env is not needed, not the reward_num_workers. Sorry about the confusion.

tunix/rl/reward_manager.py
```diff
@@ -101,6 +106,132 @@ def __init__(
...
+      call_kwargs: Dict[str, Any],
+  ) -> Any:
+    """Evaluates one reward fn over the batch, using workers when enabled."""
+    fn_name = getattr(reward_fn, "__name__", str(reward_fn))
+    if (
+        not prompts
+        or not self._parallel_enabled()
+        or fn_name in self._parent_only_fns
+    ):
+      # An empty batch has nothing to spread over workers; let the fn see
```
wang2yn84: Empty batch should be error?

wang2yn84:
Also, please squash commits into 1.

tunix/rl/reward_manager.py
gutianyu-google: Right, only the env var was removed; `reward_num_workers` is still there (0 = off, -1 = one worker per CPU).

tests/rl/reward_manager_test.py
gutianyu-google: Done internally in cl/978284674

tunix/rl/reward_manager.py
gutianyu-google: Done, it raises now.

gutianyu-google:
Squashed to one commit.

tunix/rl/algorithm_config.py
```diff
@@ -37,7 +38,17 @@ class AlgorithmConfig:
...
+  # reward functions over the batch. `0` (default) keeps the serial
+  # implementation and preserves prior behavior bit-for-bit. Values > 1
+  # chunk the (prompts, completions) batch across a fork-based process
+  # pool — reward functions are called with contiguous slices, so
+  # per-sequence reward functions produce identical results. Functions
+  # that cannot run in a worker (unpicklable, or spawning subprocesses of
+  # their own) are detected at runtime and evaluated in the parent
+  # process instead; any other failure falls back to the serial path.
+  # `-1` uses one worker per CPU.
+  reward_num_workers: int = 0
```
wang2yn84, line 53: Shall we default this to -1. This optimization seems doesn't have side effects and we should turn it on by default.

tunix/rl/reward_manager.py
```diff
@@ -27,6 +28,10 @@
 
 RewardFn = Callable[..., Any]
 
+# How long to wait for one reward-fn chunk before treating the worker pool as
+# unusable for that fn and re-evaluating it in the parent process.
+_REWARD_WORKER_TIMEOUT_SECONDS = 180
```
wang2yn84: Shall we make it configurable?

tunix/rl/algorithm_config.py
gutianyu-google: Setting -1 by default breaks the existing learner tests whose reward function depends on the whole batch (`reward_1` uses the sequence's position in the batch, and chunking changes that). So I think leaving 0 as default is the right path.

tunix/rl/reward_manager.py
gutianyu-google: Done: `reward_worker_timeout_seconds` in `AlgorithmConfig`, default 180.

tunix/rl/algorithm_config.py
wang2yn84: We probably should fix the test instead of compromise here. But that's outside of the scope of this PR. Let's land it.

wang2yn84 (approved):

PR #1927, opened 2026-08-19
Update pinned commits for vllm and tpu-inference

niting:
Updates pinned commits for `vllm` and `tpu-inference` to mutually verified versions:

- **vllm**: `f5bb701fa270f5c801f1572e1478b56f292d8dfc` (2026-08-03)
- **tpu-inference**: `a5596b27f02d1b1f1fb64c8bd4b0a73ae19b0336` (2026-08-09, matches vLLM LKG `f5bb701fa`)

Ran gsm8k on the agentic learner to confirm this works.

### DIFF
--- a/.github/workflows/tpu-nightly-regression.yml
+++ b/.github/workflows/tpu-nightly-regression.yml
@@ -44,7 +44,7 @@ jobs:
     runs-on: [linux-x86-ct6e-180-8tpu]
     environment: testing
     container:
-      image: vllm/vllm-tpu:nightly-20260731-df7d1ff-a7a204c
+      image: vllm/vllm-tpu:nightly-20260819-12eed6e-9842d70
       options: --privileged
       env:
         CLOUD_TPU_ACCELERATOR: v5e-8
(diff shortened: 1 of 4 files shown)


wang2yn84:
Hi Nitin, our rule of thumb is to pick an image from https://hub.docker.com/r/vllm/vllm-tpu/tags, then use the commits for the requirement files. That's how we setup the CI: https://github.com/google/tunix/blob/518a0fdaded4f4574062c36a45941d8fe719d20e/.github/workflows/tpu-tests.yml#L136.

niting:
> Hi Nitin, our rule of thumb is to pick an image from https://hub.docker.com/r/vllm/vllm-tpu/tags, then use the commits for the requirement files. That's how we setup the CI:
> 
> https://github.com/google/tunix/blob/518a0fdaded4f4574062c36a45941d8fe719d20e/.github/workflows/tpu-tests.yml#L136
> 
> .

vllm/vllm-tpu:nightly-20260810-a5596b2-f5bb701 is actually the image where the VLLM commit is from. I updated tpu-tests.yml and tpu-nightly-regression.yml as well - let me know if that's not needed though.

PR #1988, opened 2026-08-26
feat(worker): add execution logging, and tune gRPC keepalives

SurbhiJainUSC:
## Summary                                                                                                                                                                                                  
                                                                                                                                                                                                                
This PR adds task execution logging with error tracebacks to `RemoteExecutionServer`, and tunes gRPC keepalive parameters to maintain stable worker connections during long compilation and weight-staging windows.                                                                                                              
 
 ## Motivation & Context                                                                                                                                                                                     
1. **gRPC Channel Drops on Long Idle / Compiles**: During long JAX model compilations or multi-gigabyte weight-staging phases, gRPC channels could disconnect due to ping-interval strictness or idle timeouts. Setting `max_pings_without_data=0`, `max_ping_strikes=0`, and `min_recv_ping_interval_without_data_ms=5000` ensures channels stay healthy.
2. **Task Failure Observability**: When a task failed on `RemoteExecutionServer`, error details were wrapped in `ExecutionResponse` without server-side error logging, making remote worker crash debugging difficult.

**Checklist**
<!--- Please make sure all checkboxes are ticked before submitting this PR for review. -->

### DIFF
--- a/tunix/experimental/worker/remote_execution.py
+++ b/tunix/experimental/worker/remote_execution.py
@@ -88,8 +88,10 @@ def _grpc_options() -> List[Tuple[str, int]]:
       ("grpc.keepalive_time_ms", 20000),
       ("grpc.keepalive_timeout_ms", 10000),
       ("grpc.keepalive_permit_without_calls", 1),
-      ("grpc.http2.min_recv_ping_interval_without_data_ms", 10000),
+      ("grpc.http2.max_pings_without_data", 0),
       ("grpc.http2.max_ping_strikes", 0),
+      ("grpc.http2.min_ping_interval_without_data_ms", 5000),
+      ("grpc.http2.min_recv_ping_interval_without_data_ms", 5000),
   ]
 
 
@@ -246,7 +248,24 @@ class RemoteExecutionServer(abc.ABC):
     return request.request_id
 
   async def _run_and_enqueue(self, request: ExecutionRequest) -> None:
+    logging.debug(
+        "[RemoteExecutionServer] Starting task %s method=%s",
+        request.request_id,
+        request.method_name,
+    )
     response = await self.execute_request(request)
+    if response.error_message:
+      logging.error(
+          "[RemoteExecutionServer] Task %s failed: %s\n%s",
+          request.request_id,
+          response.error_message,
+          response.traceback,
+      )
+    else:
+      logging.debug(
+          "[RemoteExecutionServer] Task %s finished successfully",
+          request.request_id,
+      )
     await self._get_response_queue().put(response)
 
   async def poll_response(
--- a/tests/experimental/worker/remote_execution_test.py
+++ b/tests/experimental/worker/remote_execution_test.py
@@ -909,7 +909,7 @@ class RemoteExecutionTest(absltest.TestCase):
   def test_grpc_options_tolerate_client_keepalive_pings(self):
     options = dict(remote_lib._grpc_options())
     self.assertEqual(
-        options["grpc.http2.min_recv_ping_interval_without_data_ms"], 10000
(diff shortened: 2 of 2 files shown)


tunix/experimental/worker/remote_execution.py
```diff
@@ -516,6 +535,91 @@ async def poll_responses(
     """Long-polls remote server response queue for completed result."""
     pass
 
+  def info(self) -> Any:
```
jiangyangmu: ActorHandle is meant to be generic (even more generic then worker), I don't think we should add worker interfaces here.

The type safety should be handled at worker wrapper level instead of here.

s-noghabi: +1 these should not be added in this layer.

tunix/experimental/worker/remote_execution.py
```diff
@@ -88,8 +88,10 @@ def _grpc_options() -> List[Tuple[str, int]]:
       ("grpc.keepalive_time_ms", 20000),
       ("grpc.keepalive_timeout_ms", 10000),
       ("grpc.keepalive_permit_without_calls", 1),
-      ("grpc.http2.min_recv_ping_interval_without_data_ms", 10000),
+      ("grpc.http2.max_pings_without_data", 0),
       ("grpc.http2.max_ping_strikes", 0),
+      ("grpc.http2.min_ping_interval_without_data_ms", 5000),
+      ("grpc.http2.min_recv_ping_interval_without_data_ms", 5000),
```
s-noghabi, line 94: why are we changing this? did  you experiences issues with current values?

tunix/experimental/worker/remote_execution.py
SurbhiJainUSC: Thanks for the feedback! I have removed all worker lifecycle and weight-sync methods from `ActorHandle`.

tunix/experimental/worker/remote_execution.py
```diff
@@ -246,7 +248,24 @@ async def dispatch_task(self, request: ExecutionRequest) -> str:
     return request.request_id
 
   async def _run_and_enqueue(self, request: ExecutionRequest) -> None:
+    logging.debug(
+        "[RemoteExecutionServer] Starting task %s method=%s",
+        request.request_id,
+        request.method_name,
+    )
     response = await self.execute_request(request)
+    if response.error_message:
```
tianshub, line 257: our upper layer framework should already handle the extraction of error msg from failed response, do you need this logging here?

SurbhiJainUSC: The motivation is to have visibility directly in the worker pod logs when diagnosing worker-side failures rather than only in the orchestrator log.
Let me know if this makes sense, otherwise I can remove it if preferred.

tunix/experimental/worker/remote_execution.py
SurbhiJainUSC: I haven't tried with the default values - this was an agent-suggested tweak intended to make long-lived idle gRPC channels more tolerant of keepalive pings during long training/rollout steps in large model runs. If you prefer keeping the gRPC options unchanged, I am happy to revert these back to the previous defaults.

s-noghabi: It does make it more tolerant but might add more load. we can keep your changes and if at scale we see issues bump it up

s-noghabi (approved):
Thanks for making the fixes!

tunix/experimental/worker/remote_execution.py
jiangyangmu: In error case, can we still use logging.debug for now ?

Consider the volumn of requests, I can imagine it's easy to generate tons of noisy logs (even inside a single worker node) and hides really useful signals.

In general I think this should be turned on by a switch, instead of always-on.

PR #1989, opened 2026-08-26
infra: add model-specific Pathways template with scratch volume, memory limits, and cached kube context

SurbhiJainUSC:
## Summary
              
This PR improves GKE JobSet stability for large-scale distributed training jobs by adding an ephemeral scratch disk for crash core dumps, setting explicit memory requests/limits for the Pathways user container, and making GKE context initialization faster and source-safe.
 
## Motivation & Problem Statement
1. **Boot Disk Evictions from Crash Dumps**: When training 30B+ MoE models or using native TPU weight transfer, a process crash or segmentation fault can produce a 50GB+ core dump. Without a dedicated volume mount, core dumps write to the node's local boot disk (~44GB allocatable on TPU worker nodes), triggering node-wide ephemeral-storage pressure and evicting all pods.
2. **Transient Host-Memory OOMKills**: Without explicit memory guarantees (`requests.memory`), the user container has no QoS protection. Under transient host memory spikes (e.g. during D2H host-staging of ~80GB of bf16 parameters on 30B+ MoE models), the kubelet would OOM-kill the container even when the node had hundreds of gigabytes of free physical capacity.
3. **Interactive Shell Termination in Context Script**: Running `source enter_kube_context.sh` with `exit 1` terminated the developer's interactive terminal session upon credential failure.
 
### 1. jobset.pathways.yaml
* **Ephemeral Scratch Volume (400Gi)**: Added a `premium-rwo` generic ephemeral volume claim template mounted at `/scratch`. Sized to comfortably absorb full-RSS memory dumps.
* **Core Dump Redirection**: Configured `/proc/sys/kernel/core_pattern` inside the privileged container to redirect core files to `/scratch/core.%e.%p.%t`.
* **Memory Guarantees**: Set `requests.memory: 200G` and `limits.memory: 420G` on `${USER_CONTAINER}` to protect the workload during heavy parameter staging and resharding.
  
### 2. enter_kube_context.sh
* **Fast Context Check**: Avoids redundant `gcloud container clusters get-credentials` network calls if the context is already present in `kubectl config get-contexts`.
* **Source-Safe Error Handling**: Uses `return 1 2>/dev/null || exit 1` so that sourcing the script safely returns without terminating the user's active terminal.       

**Checklist**

### DIFF
--- a/tunix/experimental/examples/math_gsm8k_dist/enter_kube_context.sh
+++ b/tunix/experimental/examples/math_gsm8k_dist/enter_kube_context.sh
@@ -8,6 +8,12 @@ ZONE=${ZONE:-us-central1-a}
 CLUSTER=${CLUSTER:-trellis-demo-0810}
 
 export KUBECONFIG="${KUBECONFIG:-$HOME/.kube/config.$PROJECT.$REGION.$CLUSTER}"
-gcloud container clusters get-credentials $CLUSTER --region=$REGION --project=$PROJECT --dns-endpoint || { echo "gcloud get-credentials failed"; return; }
-kubectl config use-context "gke_${PROJECT}_${REGION}_${CLUSTER}" || { echo "kubectl use-context failed"; return; }
-kubectl config set-context --current --namespace=default || { echo "kubectl set-context failed"; return; }
+CONTEXT_NAME="gke_${PROJECT}_${REGION}_${CLUSTER}"
+
+if kubectl config get-contexts "$CONTEXT_NAME" &>/dev/null; then
+  kubectl config use-context "$CONTEXT_NAME" >/dev/null || true
+else
+  gcloud container clusters get-credentials $CLUSTER --region=$REGION --project=$PROJECT --dns-endpoint || { echo "gcloud get-credentials failed" >&2; return 1 2>/dev/null || exit 1; }
+  kubectl config use-context "$CONTEXT_NAME" >/dev/null || { echo "kubectl use-context failed" >&2; return 1 2>/dev/null || exit 1; }
+fi
+kubectl config set-context --current --namespace=default >/dev/null || true
--- a/tunix/experimental/distributed/deployment/yamls/jobset.pathways.qwen3.5-397b.yaml
+++ b/tunix/experimental/distributed/deployment/yamls/jobset.pathways.qwen3.5-397b.yaml
@@ -1,0 +1,320 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: ${JOBSET_NAME}
+  namespace: default
+spec:
+  coordinator:
+    replicatedJob: proc
+  failurePolicy:
+    restartStrategy: Recreate
+  network:
+    enableDNSHostnames: true
+    publishNotReadyAddresses: true
+  replicatedJobs:
+  - name: proc
+    replicas: ${REPLICAS}
+    template:
+      metadata:
+        annotations: {}
+      spec:
+        backoffLimit: 0
+        completionMode: Indexed
+        completions: 1
+        parallelism: 1
+        template:
+          spec:
+            dnsPolicy: ClusterFirstWithHostNet
+            hostNetwork: true
+            restartPolicy: Never
+            terminationGracePeriodSeconds: 10
+            nodeSelector:
+              cloud.google.com/gke-tpu-accelerator: ${TPU_TYPE}
+              cloud.google.com/gke-tpu-topology: ${TPU_TOPOLOGY}
+            affinity:
+              podAffinity:
+                requiredDuringSchedulingIgnoredDuringExecution:
+                # place on one of the pathways-worker nodes
+                - topologyKey: kubernetes.io/hostname
+                  labelSelector:
+                    matchExpressions:
+                    - key: jobset.sigs.k8s.io/jobset-name
+                      operator: In
+                      values:
+                      - ${JOBSET_NAME}
+                    - key: jobset.sigs.k8s.io/replicatedjob-name
+                      operator: In
+                      values:
+                      - pw-node
+            volumes:
+            # Ephemeral scratch volume for large core dumps to prevent node
+            # eviction from ephemeral-storage pressure on small boot disks.
+            - name: scratch-disk
+              ephemeral:
+                volumeClaimTemplate:
+                  spec:
+                    accessModes: ["ReadWriteOnce"]
+                    storageClassName: premium-rwo
+                    resources:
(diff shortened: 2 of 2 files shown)


tunix/experimental/distributed/deployment/yamls/jobset.pathways.yaml
jiangyangmu, line 1: Can you fork into a new template for the changes, e.g. jobset.pathways.trellis.yaml ? so the base template remains minimum.

The scratch-disk is not always necessary for dev workflow.
The memory requirement is not small and will make this template unusable for small testing clusters.

tunix/experimental/examples/math_gsm8k_dist/enter_kube_context.sh
jiangyangmu, line 1: What are the improvements of changes here ?

jiangyangmu: The "Fast Context Check" part makes sense.

The "Source-Safe Error Handling" part is probably referring to old version which uses "exit 1", now it's replaced with "return".

tunix/experimental/distributed/deployment/yamls/jobset.pathways.yaml
```diff
@@ -121,10 +141,21 @@ spec:
...
+              # Without an explicit request, this container has no memory
+              # guarantee and kubelet can OOM-kill it under any transient
+              # pressure even when the node has ample free capacity --
+              # confirmed the hard way: OOMKilled here during Raiden's D2H
+              # host-staging of a ~30B param MoE model with padded MoE MLP
+              # weights (~80GB of host-resident bf16 params), on a node that
+              # `kubectl describe node` showed had 300+GB free at the time.
+              resources:
+                requests:
+                  memory: 200G
```
s-noghabi: these values seem quite high. do we want to hardcode with such big values? this will make it hard to run on smaller setups.

tunix/experimental/distributed/deployment/yamls/jobset.pathways.yaml
```diff
@@ -137,6 +168,14 @@ spec:
                   _sigterm() (kill -SIGTERM $$! 2>/dev/null)
                   trap _sigterm SIGTERM
 
+                  # Best-effort: redirect crash core dumps onto the scratch-disk
+                  # volume instead of the node's own small boot disk. core_pattern
+                  # is a host-wide (not per-container) sysctl, so this only takes
+                  # effect when the container can write /proc/sys/kernel -- true
+                  # here since the container runs privileged -- and is harmless
+                  # (STARTUP_COMMAND still proceeds) if it can't.
+                  echo "/scratch/core.%e.%p.%t" > /proc/sys/kernel/core_pattern 2>/dev/null || true
```
s-noghabi: Isn't /proc/sys/kernel/core_pattern  a host-kernel-wide sysctl and not namespaced per-container? wouldn't this cause issues?

tunix/experimental/distributed/deployment/yamls/jobset.pathways.yaml
tianshub: +1 you can also consider putting model name in the specific file so it's easy to correlate the setting. e.g. jobset.pathways.qwen3.5-397b.yaml

tunix/experimental/examples/math_gsm8k_dist/enter_kube_context.sh
```diff
@@ -7,7 +7,12 @@ REGION=${REGION:-us-central1}
...
 
-export KUBECONFIG="${KUBECONFIG:-$HOME/.kube/config.$PROJECT.$REGION.$CLUSTER}"
-gcloud container clusters get-credentials $CLUSTER --region=$REGION --project=$PROJECT --dns-endpoint || { echo "gcloud get-credentials failed"; return; }
-kubectl config use-context "gke_${PROJECT}_${REGION}_${CLUSTER}" || { echo "kubectl use-context failed"; return; }
-kubectl config set-context --current --namespace=default || { echo "kubectl set-context failed"; return; }
+CONTEXT_NAME="gke_${PROJECT}_${REGION}_${CLUSTER}"
+if kubectl config get-contexts "$CONTEXT_NAME" &>/dev/null; then
+  kubectl config use-context "$CONTEXT_NAME" >/dev/null || true
+else
+  export KUBECONFIG="${KUBECONFIG:-$HOME/.kube/config.$PROJECT.$REGION.$CLUSTER}"
```
s-noghabi: shouldn't this be above  CONTEXT_NAME? the if above statement checks against this

tunix/experimental/examples/math_gsm8k_dist/enter_kube_context.sh
```diff
@@ -7,7 +7,12 @@ REGION=${REGION:-us-central1}
...
+  gcloud container clusters get-credentials $CLUSTER --region=$REGION --project=$PROJECT --dns-endpoint &>/dev/null || { echo "gcloud get-credentials failed" >&2; return 1 2>/dev/null || exit 1; }
```
s-noghabi: why are we suppressing the get_cridentials errors by using &>/dev/null? seems more useful to know the details

tunix/experimental/examples/math_gsm8k_dist/enter_kube_context.sh
SurbhiJainUSC: In main:
`kubectl config use-context "gke_${PROJECT}_${REGION}_${CLUSTER}" || { echo "kubectl use-context failed"; return; }`
echo is the last executed command, and echo always exits with success(0). A bare `return;` returns the status of the last command, meaning a failed kubectl command returned success.

However, in this PR:
`kubectl config use-context "$CONTEXT_NAME" >/dev/null || { echo "kubectl use-context failed" >&2; return 1 2>/dev/null || exit 1; }`
`return 1` guarantees a proper non-zero failure code (1).

tunix/experimental/examples/math_gsm8k_dist/enter_kube_context.sh
SurbhiJainUSC: Great catch, moved `export KUBECONFIG` above `CONTEXT_NAME` and the `if` statement so `kubectl config get-contexts` evaluates against the dedicated `$KUBECONFIG` file rather than default `~/.kube/config`.

tunix/experimental/examples/math_gsm8k_dist/enter_kube_context.sh
SurbhiJainUSC: Good point! Removed `&>/dev/null` so error messages are printed to the console when credentials need to be fetched.

tunix/experimental/distributed/deployment/yamls/jobset.pathways.yaml
SurbhiJainUSC: Done

tunix/experimental/distributed/deployment/yamls/jobset.pathways.yaml
SurbhiJainUSC: This memory requirement is for Qwen3.5-397B specifically. I have created a model-specific template containing these memory allocations for 397B workload.

tunix/experimental/distributed/deployment/yamls/jobset.pathways.yaml
SurbhiJainUSC: Yes, you are right. I have removed the core_pattern modification and instead set `ulimit -c 0` inside the container entrypoint to prevent large crash dumps from filling the boot disk without affecting the host.

tunix/experimental/examples/math_gsm8k_dist/enter_kube_context.sh
jiangyangmu: Make sense, thanks for the improvement!

jiangyangmu (approved):

PR #2013, opened 2026-08-27
Add MaxText and Raiden optional dependencies to Dockerfile

SurbhiJainUSC:
### Summary
  
This PR adds modular build support for MaxText and Raiden dependencies in Dockerfile via conditional build arguments (INSTALL_MAXTEXT and INSTALL_RAIDEN), allowing reproducible container builds for distributed RL workloads without bloating or breaking the standard base image.

### Key Changes
1. Dockerfile:
   - Installs uv directly in the virtual environment (/opt/venv).
   - Adds ARG `INSTALL_MAXTEXT=false` to conditionally install MaxText/Flax training dependencies from requirements/maxtext_requirements.txt.
   - Adds `ARG INSTALL_RAIDEN=false` to conditionally install Raiden weight synchronization wheels (/app/raiden_wheels/*.whl) or `tpu-raiden-jax` from Google Artifact Registry.
2. requirements/maxtext_requirements.txt (maxtext_requirements.txt):
   - Defines standalone dependencies required for running MaxText trainer backends (e.g. maxtext, flax, optax).
3. requirements/special_requirements.txt (special_requirements.txt):
   - Pins compatible package versions for distributed RL runtimes.
4. tunix/oss/utils.py (utils.py):
   - Refines OSS utility helper imports and formatting.

### Tests
Nightly regression tests: https://github.com/google/tunix/actions/runs/33435227159
  
**Checklist**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ ENV PATH="/opt/venv/bin:$PATH"
 # Upgrade pip
 RUN pip install --upgrade pip
 
+RUN pip install uv
 RUN pip install git+https://github.com/ayaka14732/jax-smi.git
 # If you encounter a checkpoint issue, try using following old version of pathways-utils.
 # RUN pip install git+https://github.com/AI-Hypercomputer/pathways-utils.git@b72729bb152b7b3426299405950b3af300d765a9#egg=pathwaysutils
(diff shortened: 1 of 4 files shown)


tunix/oss/utils.py
jiangyangmu, line 1: This will change user workflow, can you revert it ?

Dockerfile
```diff
@@ -37,6 +38,26 @@ RUN pip install -e .
...
+
+# Build argument to conditionally install Raiden weight sync dependencies
+ARG INSTALL_RAIDEN=false
+
+# Install Raiden specific dependencies conditionally
+RUN if [ "$INSTALL_RAIDEN" = "true" ]; then \
+      if [ -d "/app/raiden_wheels" ] && ls /app/raiden_wheels/*.whl 1>/dev/null 2>&1; then \
+        pip install --force-reinstall --no-deps /app/raiden_wheels/*.whl; \
+      else \
+        pip install tpu-raiden-jax --extra-index-url https://us-python.pkg.dev/cloud-tpu-inference-test/tpu-raiden/simple/; \
```
jiangyangmu, line 58: does this work w/o `pip install keyrings.google-artifactregistry-auth` ?

tunix/oss/utils.py
SurbhiJainUSC: The reason for this change is that `hf.login()` opens an interactive shell whenever `HF_TOKEN` is unset. In non-interactive environments such as GKE, this causes the process to hang waiting for user input or fail.

Dockerfile
SurbhiJainUSC: good catch, updated.

tunix/oss/utils.py
jiangyangmu: In GKE we should set HF_TOKEN env from "secret" in this case. Which should be fixed in YAML, not here.

requirements/special_requirements.txt
```diff
@@ -2,5 +2,4 @@
 # --find-links https://storage.googleapis.com/jax-releases/libtpu_releases.html
 # --pre
 
-tpu-inference @ git+https://github.com/vllm-project/tpu-inference.git@12eed6e661a1ffacbce9d107a9894c1a471e7479
```
jiangyangmu, line 5: Bump this version should be careful, we need to run package test and nightly regression test before bump.

Do you depend on changes in the updated version ?

SurbhiJainUSC: Yes, we need these weight sync changes in tpu-inference: https://github.com/vllm-project/tpu-inference/commit/4a8951606026d5c0307554bc4ed93f4690c4640a

wang2yn84: When you update tpu-inference, please also update vllm, and the vllm nightly image tag in https://github.com/google/tunix/tree/main/.github/workflows. So the correct way of doing this is identify which vllm docker image you'd like to use, then apply the same commit to these requirement files.

PR #2046, opened 2026-08-31
Carry rollout MoE routing from the vLLM sampler to the maxtext trainer

khatwanimohit:
Lets an RL run replay the MoE expert routing a rollout actually used instead of re-running the router during training. Both ends already existed -- vLLM and tpu-inference capture routing (`enable_return_routed_experts` -> `CompletionOutput.routed_experts`, `[length, num_layers, top_k]`) and MaxText consumes it (`forced_routed_experts`) -- so this is the middle.

Sampler side:
- `VllmConfig.return_routed_experts` sets vLLM's engine arg; `VllmSampler.detokenize` collects the arrays and `SamplerOutput` carries them.
- The in-process adapter populates `SamplingResponse.routed_experts`, and warns when a request asks for routing the engine was not built to capture, since capture is engine-level and the alternative is silently handing the trainer None -- indistinguishable from a dense model.

Pipeline side:
- `routed_experts` on `TokenSegment`, `TrajectoryItem` and `RLTrainerPayload`, plus `UNSET_ROUTED_EXPERT = -1` for slots the trainer must not replay.
- The GRPO adapter pads short captures with the sentinel rather than expert 0.
- The padded assembler lays routing out over `[prompt | completion]` the same way the token ids are padded: prompts right-aligned keeping the tail (as `_left_pad` does), completions left-aligned. Getting this wrong detaches every replayed expert from its token -- silent, not a crash.

Demo side (math_gsm8k_dist), all opt-in:
- trainer node: `--trainer_engine=maxtext` swaps PeftTrainer for MaxText's engine wired for replay, importing MaxText lazily and skipping the tunix actor model, which only knows the demo's dense architectures. `--maxtext_load_parameters_path` points it at the checkpoint; without one it warns, since random trainer weights make replayed routing meaningless.
- rollout node: `--return_routed_experts`, and `--load_rollout_weights` to load the real checkpoint instead of vLLM's dummy weights. With both sides on the same weights a first step is meaningful with no weight transfer at all.
- `launcher.sh` and `k8s_launcher.sh`: `ROUTER_REPLAY=1` sets the above; slices and meshes are now configurable rather than hardcoded.
- Dockerfile: opt-in `INSTALL_MAXTEXT` build arg, which also drops triton (GPU-only, and its native module aborts the process when MaxText imports on a TPU host).

Tests pin layout rather than mere presence, since the realistic failure is misalignment. The loss assertions are differential -- two different replays must disagree -- because a "loss is finite" check passes just as happily when the routing is dropped. Each was verified to fail against a mutation that disables what it covers.

**Checklist**

### DIFF
--- a/tunix/experimental/common/datatypes.py
+++ b/tunix/experimental/common/datatypes.py
@@ -37,6 +37,10 @@ Step = agent_types.Step
 TrajectoryStatus = agent_types.TrajectoryStatus
 Role = common_datatypes.Role
 
+# Marks a router-replay slot the trainer must not replay, so the model falls
+# back to its own gate there. Matches what MaxText's replay path expects.
+UNSET_ROUTED_EXPERT = -1
+
 
 # TODO(tunix-dev): Unify this extended TrajectoryItem back into
 # agent_types.TrajectoryItem so that all agentic workflows share the same strict
@@ -44,6 +48,7 @@ Role = common_datatypes.Role
 @dataclasses.dataclass(kw_only=True)
 class TrajectoryItem:
   """Extended TrajectoryItem for Orchestrator with token arrays."""
+
   prompt_id: str = ""
   group_index: int = 0
   start_step: int = 0
@@ -51,6 +56,9 @@ class TrajectoryItem:
   prompt_tokens: np.ndarray | None = None
   completion_tokens: np.ndarray | None = None
   action_mask: np.ndarray | None = None
+  # `[len(prompt_tokens) + len(completion_tokens), num_layers, top_k]` expert
+  # ids from the rollout, for replaying its routing during training.
+  routed_experts: np.ndarray | None = None
   policy_version: int = 0
   metadata: dict[str, Any] = dataclasses.field(default_factory=dict)
 
@@ -272,12 +280,17 @@ class TokenSegment:
     loss_mask: Array of ints, 1 where the token is model-emitted (trainable).
     logps: Array of per-token log-probabilities under the sampling distribution,
       or None for spans the model did not emit (e.g. env tokens).
+    routed_experts: `[len(tokens), num_layers, top_k]` MoE expert ids this span
+      was routed through, so training can replay the routing the rollout
+      actually used. None for dense models, spans the model did not emit, or
+      when the sampler was not asked to capture routing.
   """
 
   source: str
   tokens: np.ndarray
   loss_mask: np.ndarray
   logps: np.ndarray | None = None
+  routed_experts: np.ndarray | None = None
 
   def __post_init__(self):
     if self.loss_mask.shape != self.tokens.shape:
@@ -289,6 +302,19 @@ class TokenSegment:
       raise ValueError(
           f"logps shape {self.logps.shape} != tokens shape {self.tokens.shape}"
       )
+    if self.routed_experts is not None:
+      # The trailing axes are [num_layers, top_k] and are model-dependent, so
+      # only the rank and the leading (per-token) axis are checked.
+      if self.routed_experts.ndim != 3:
+        raise ValueError(
+            "routed_experts must be [length, num_layers, top_k]; got shape"
+            f" {self.routed_experts.shape}"
+        )
+      if self.routed_experts.shape[0] != self.tokens.shape[0]:
+        raise ValueError(
+            f"routed_experts shape {self.routed_experts.shape} does not cover"
+            f" tokens shape {self.tokens.shape} along axis 0"
+        )
 
 
 @dataclasses.dataclass(kw_only=True)
@@ -577,6 +603,11 @@ class RLTrainerPayload(TrainerPayload):
     ref_per_token_logps: Optional [B, C] reference model log-probabilities.
     old_per_token_logps: Optional [B, C] behavior policy log-probabilities.
     sampler_is_weights: Optional [B, C] importance sampling weights.
+    routed_experts: Optional `[B, T, num_layers, top_k]` MoE expert ids
+      captured during rollout. When set, a training engine that supports router
+      replay forces these experts instead of re-running its own gate, so the
+      training forward pass matches the routing the rollout actually used. `-1`
+      marks a padded or unused slot.
     returns: Optional [B, C] value baseline returns (for PPO / Critic).
     old_values: Optional [B, C] critic value estimates (for PPO / Critic).
     metadata: Extra payload metadata dictionary.
@@ -594,6 +625,7 @@ class RLTrainerPayload(TrainerPayload):
   ref_per_token_logps: ArrayLike | None = None
   old_per_token_logps: ArrayLike | None = None
   sampler_is_weights: ArrayLike | None = None
+  routed_experts: ArrayLike | None = None
   returns: ArrayLike | None = None
   old_values: ArrayLike | None = None
   metadata: dict[str, Any] = dataclasses.field(default_factory=dict)
--- a/tunix/experimental/orchestrator/batch_assembly.py
+++ b/tunix/experimental/orchestrator/batch_assembly.py
@@ -54,6 +54,51 @@ def _left_pad(
   return out, mask
 
 
+def _routed_experts_aligned(
+    routed: np.ndarray,
+    prompt_len: int,
+    completion_len: int,
+    max_prompt_length: int,
+    max_response_length: int,
+) -> np.ndarray:
+  """Lays one row of routing out over the padded `[prompt | completion]`.
+
+  Mirrors how the token ids themselves are padded -- prompts right-aligned in
+  the prompt window (keeping the tail, as `_left_pad` does) and completions
+  left-aligned in the response window -- so replayed routing stays attached to
+  the token it was captured for. Everything else is left unset.
+
+  Args:
+    routed: `[prompt_len + completion_len, num_layers, top_k]` for one
+      generation. Only axis 0 (the per-token axis) is ever sliced below; the
+      trailing `[num_layers, top_k]` axes are carried through untouched.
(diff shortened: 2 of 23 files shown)


tests/experimental/worker/router_replay_trainer_worker_test.py
```diff
@@ -0,0 +1,301 @@
...
+      learning_rate=1e-2,
+      warmup_steps_fraction=0.0,
+  )
+
+
+def _payload(forced):
+  """Wraps a forced-routing array in a payload the replay loss fn accepts."""
+  forced[:, -PAD_LEN:] = datatypes.UNSET_ROUTED_EXPERT
+  token_ids, token_mask = _tokens()
+  return maxtext_engine.RouterReplayTrainerPayload(
```
tianshub: Since the field is added to [RLTrainerPayload](https://github.com/google/tunix/blob/795635a28430c03de2911a5c595db59748e3e2bf/tunix/experimental/common/datatypes.py#L535), why do you still need to use RouterReplayTrainerPayload here? The actual data class that passed to maxtext should be RLTrainerPayload no?

tunix/experimental/common/datatypes.py
```diff
@@ -289,6 +302,16 @@ def __post_init__(self):
       raise ValueError(
           f"logps shape {self.logps.shape} != tokens shape {self.tokens.shape}"
       )
+    # Only the leading (per-token) axis has to line up; the trailing axes are
+    # [num_layers, top_k] and are model-dependent.
+    if (
+        self.routed_experts is not None
+        and self.routed_experts.shape[0] != self.tokens.shape[0]
```
tianshub: you can add a check on the ndim=3 as well

tunix/experimental/examples/math_gsm8k_dist/k8s_launcher.sh
```diff
@@ -46,6 +46,17 @@ export WEIGHT_SYNC_MODE=${WEIGHT_SYNC_MODE:-none}
 # MaxText trainer configuration: only consulted when TRAINER_BACKEND=maxtext
 export MAXTEXT_MODEL_NAME=${MAXTEXT_MODEL_NAME:-qwen3-1.7b}
 export MAXTEXT_CKPT=${MAXTEXT_CKPT:-}
+# Router replay: the rollout records the MoE experts it routed through and the
+# MaxText trainer replays them instead of re-running its own gate. MoE models
+# only; requires TRAINER_BACKEND=maxtext.
+#
+# This also loads the real checkpoint into the rollout instead of vLLM's dummy
```
tianshub: put the comment after L58?

tunix/experimental/examples/math_gsm8k_dist/launcher.sh
```diff
@@ -55,6 +55,17 @@ WANDB_RUN_NAME=${WANDB_RUN_NAME:-}
 WANDB_API_KEY=${WANDB_API_KEY:-}
 SAMPLER=${SAMPLER:-inprocess_vllm}
 WEIGHT_SYNC_MODE=${WEIGHT_SYNC_MODE:-none}
+# Router replay: the rollout records the MoE experts it routed through and the
```
tianshub: we can delete this comment since launch script already explained it

tunix/experimental/examples/math_gsm8k_dist/launcher.sh
```diff
@@ -55,6 +55,17 @@ WANDB_RUN_NAME=${WANDB_RUN_NAME:-}
...
+# MaxText trainer replays them instead of re-running its own gate. MoE models
+# only; both must be on for replay to happen.
+ROUTER_REPLAY=${ROUTER_REPLAY:-0}
+TRAINER_BACKEND=${TRAINER_BACKEND:-tunix}
+MAXTEXT_MODEL_NAME=${MAXTEXT_MODEL_NAME:-qwen3.5-35b-a3b}
+MAXTEXT_CKPT=${MAXTEXT_CKPT:-}
+# Load the real checkpoint into vLLM instead of its dummy random weights, so
```
tianshub: ditto

tunix/experimental/examples/math_gsm8k_dist/run_trainer_node.py
```diff
@@ -275,6 +285,69 @@ def close(self) -> None:
...
+  `forced_routed_experts`.
+  """
+  from maxtext.training_engine import maxtext_engine  # pylint: disable=g-import-not-at-top
+
+  def gen_model_input_fn(payload: Any) -> dict[str, Any]:
+    token_ids, token_mask = _concat_prompt_completion(payload)
+    # Reuse MaxText's adapter for positions/targets/segmentation rather than
+    # re-deriving them; it already handles left-padded rows and the roll(-1)
+    # wrap-around at the sequence end.
+    shim = maxtext_engine.RouterReplayTrainerPayload(
```
tianshub: this redirection shouldn't be necessary right? the router_replay_gen_model_input_fn can directly consume RLTrainerPayload and extract the fields

tunix/experimental/examples/math_gsm8k_dist/run_trainer_node.py
```diff
@@ -275,6 +285,69 @@ def close(self) -> None:
       self._trainer.close()
 
 
+def _mask_or_ones(mask: Any, ids: np.ndarray) -> np.ndarray:
+  """`mask` as an array, or an all-ones mask matching `ids` when absent."""
+  return np.ones_like(ids) if mask is None else np.asarray(mask)
+
+
+def _concat_prompt_completion(payload: Any) -> tuple[np.ndarray, np.ndarray]:
```
tianshub: we should reduce these kind of functions from the script as much as possible and just bake it into the framework, let's put it in the router_replay_gen_model_input_fn. Also I doubt if you need to have separate gen_model_input function for this, wouldn't it be easier to have one function that can deal with optional field?

tunix/experimental/orchestrator/batch_assembly.py
```diff
@@ -54,6 +54,36 @@ def _left_pad(
   return out, mask
 
 
+def _routed_experts_aligned(
+    routed: np.ndarray,
```
tianshub, line 58: add comment for shape annotation

tunix/experimental/orchestrator/batch_assembly.py
```diff
@@ -54,6 +54,36 @@ def _left_pad(
...
+) -> np.ndarray:
+  """Lays `[P + C, num_layers, top_k]` routing out over the padded `[P | C]`.
+
+  Mirrors how the token ids themselves are padded -- prompts right-aligned in
+  the prompt window (keeping the tail, as `_left_pad` does) and completions
+  left-aligned in the response window -- so replayed routing stays attached to
+  the token it was captured for. Everything else is left unset.
+  """
+  routed = np.asarray(routed, dtype=np.int32)
+  prompt_part = routed[:prompt_len][-max_prompt_length:]
```
tianshub: I'm a bit confused on the routed shape, isn't dim0 suppose to be P + C? why are we slicing on dim0 and dim1?

tianshub (changes requested):

tests/experimental/worker/router_replay_trainer_worker_test.py
khatwanimohit: `RLTrainerPayload` is the only payload involved now

tunix/experimental/common/datatypes.py
khatwanimohit: Done. `TokenSegment.__post_init__` now rejects a non-3D `routed_experts` with the expected `[length, num_layers, top_k]` in the message

tunix/experimental/examples/math_gsm8k_dist/k8s_launcher.sh
khatwanimohit: done

tunix/experimental/examples/math_gsm8k_dist/launcher.sh
khatwanimohit: Deleted

tunix/experimental/examples/math_gsm8k_dist/launcher.sh
khatwanimohit: Deleted

tunix/experimental/examples/math_gsm8k_dist/run_trainer_node.py
khatwanimohit: done

tunix/experimental/examples/math_gsm8k_dist/run_trainer_node.py
khatwanimohit: agreed

tunix/experimental/orchestrator/batch_assembly.py
khatwanimohit: made it clearer now

tunix/experimental/orchestrator/batch_assembly.py
khatwanimohit: done

tunix/rl/common.py
```diff
@@ -337,6 +410,7 @@ def compute_per_token_logps(
     segment_positions: jax.Array | None = None,
     temperature: float = 1.0,
     chunk_size: int = 0,
+    forced_routed_experts: jax.Array | None = None,
```
tianshub: should this just be routed_experts so it's consistent with others?

tianshub (approved):
Thank you for the update, looks much cleaner.

tunix/rl/common.py
khatwanimohit: changed to routed_experts

PR #2051, opened 2026-08-31
Support DeepSWE on MaxText models in Tunix

niting:
Primarily adds a script train_maxtext_nb.py that uses MaxText models on a DeepSWE setup. The DeepSWE setup uses agentic sandboxes which maintains warmpools for faster training execution. Modifies Dockerfile accordingly to install necessary dependencies, adds a flag to swe_env.py to allow limiting warmpool replicas. Makes various robustness fixes for error logging etc.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,13 +32,21 @@ RUN pip install wandb
 # Set the working directory
 WORKDIR /app
 
-# Copy the project files to the image
-COPY . .
+# Copy scripts and requirements first to leverage Docker cache
+COPY scripts/install_tunix_vllm_requirement.sh scripts/
+COPY requirements/ requirements/
 
-# Install the project in editable mode
-RUN pip install -e .
+RUN bash scripts/install_tunix_vllm_requirement.sh
 
-RUN bash /app/scripts/install_tunix_vllm_requirement.sh
+# Copy pyproject.toml and README.md to install dependencies first
+COPY pyproject.toml README.md /app/
+RUN mkdir /app/tunix && touch /app/tunix/__init__.py
+RUN uv pip install .
+
+# Install SFT/MaxText dependencies (unconditional)
+RUN uv pip install --upgrade flax && \
+    uv pip install torchax aqtp tokamax math_verify drjax && \
+    uv pip install --no-deps git+https://github.com/google/maxtext.git
 
 # Build argument to conditionally install MaxText dependencies
 ARG INSTALL_MAXTEXT=false
@@ -67,11 +75,19 @@ ARG INSTALL_DEEPSWE_DEPS=false
 
 # Install DeepSWE specific dependencies and apply runtime patches conditionally
 RUN if [ "$INSTALL_DEEPSWE_DEPS" = "true" ]; then \
-      pip install kubernetes gym swebench==3.0.2 && \
-      pip install --no-deps git+https://github.com/r2e-gym/r2e-gym.git@0d94c4eb9431cd195c55a7ea3abd54006c9a1735 && \
+      uv pip install kubernetes gym swebench==3.0.2 && \
+      uv pip install --no-deps git+https://github.com/kubernetes-sigs/agent-sandbox.git#subdirectory=clients/python/agentic-sandbox-client && \
+      uv pip install --no-deps git+https://github.com/kubernetes-sigs/agent-sandbox.git#subdirectory=examples/agent-sandbox-rl && \
+      uv pip install --no-deps git+https://github.com/r2e-gym/r2e-gym.git@0d94c4eb9431cd195c55a7ea3abd54006c9a1735 && \
       sed -i 's/create_repo, upload_folder, HfFolder/create_repo, upload_folder/' /opt/venv/lib/python3.12/site-packages/r2egym/agenthub/utils/utils.py && \
       sed -i 's/self.commit = ParsedCommit(\*\*json.loads(self.commit_json))/self.commit = ParsedCommit(\*\*(json.loads(self.commit_json) if isinstance(self.commit_json, str) else self.commit_json))/' /opt/venv/lib/python3.12/site-packages/r2egym/agenthub/runtime/docker.py; \
     fi
+
+# Copy the rest of the project files
+COPY . .
+
+# Install Tunix in editable mode
+RUN uv pip install --no-deps -e .
 
 # Set the default command to bash
 CMD ["bash"]
--- a/examples/deepswe/swe_env.py
+++ b/examples/deepswe/swe_env.py
@@ -94,6 +94,7 @@ def _init_global_fleet(
     max_concurrency: int = 128,
     num_generations: int = 8,
     batch_size: int = 8,
+    max_warmpool_replicas: int | None = None,
 ) -> Any:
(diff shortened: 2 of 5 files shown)


examples/deepswe/swe_env.py
```diff
@@ -228,13 +234,19 @@ def _extract_images(self, batch: Any) -> list[str]:
   def _warm_batch(self, batch: Any, wait: bool = False):
     images = self._extract_images(batch)
     if images and self.fleet:
+      target_replicas = (
+          self.max_warmpool_size
```
tianshub: nit: this can be `self.max_warmpool_size or self.num_generations`

examples/deepswe/train_maxtext_nb.py
```diff
@@ -0,0 +1,1295 @@
...
+    help="Logging level for the script and relevant libraries.",
+)
+
+args, _ = parser.parse_known_args()
+
+MODEL_VERSION = args.model_version
+NODE_SELECTOR_VAL = args.node_selector_val
+
+
+# Monkeypatch r2egym DockerRuntime to dynamically configure Kubernetes nodeSelector.
```
tianshub: put this this in a standalone r2e_gym_helper.py file? so we can reuse it later

tunix/rl/agentic/pipeline/rollout_orchestrator.py
```diff
@@ -177,14 +177,14 @@ async def _runner(
       if isinstance(e, ExceptionGroup):
         for sub_e in e.exceptions:
           logging.error(
-              "Fatal error in runner for pair %d: %s",
-              env.extra_kwargs["pair_index"],
+              "Fatal error in runner for pair %s: %s",
+              env.extra_kwargs.get("pair_index", "unknown"),
```
tianshub: when will pair_index be None?

Dockerfile
```diff
@@ -25,28 +25,45 @@ RUN pip install git+https://github.com/ayaka14732/jax-smi.git
...
-# Install the project in editable mode
-RUN pip install -e .
+# Copy pyproject.toml and README.md to install dependencies first
+COPY pyproject.toml README.md /app/
+RUN mkdir /app/tunix && touch /app/tunix/__init__.py
+RUN uv pip install .
 
-RUN bash /app/scripts/install_tunix_vllm_requirement.sh
+# Install SFT/MaxText dependencies (unconditional)
+RUN pip install --upgrade flax && \
```
tianshub: uv pip install upgrade?

examples/deepswe/swe_env.py
niting: Done.

examples/deepswe/train_maxtext_nb.py
niting: Done.

niting: I also realized I didn't need the agentic sandbox patch. Only the patch for overriding the node selector in r2egym.

tunix/rl/agentic/pipeline/rollout_orchestrator.py
niting: I removed this entirely. It's possible this was happening in certain conditoins that are not common.

Dockerfile
niting: Done.

tianshub (approved):

examples/deepswe/swe_env.py
```diff
@@ -138,7 +139,7 @@ def _init_global_fleet(
         ],
         max_concurrent=effective_max_concurrent,
         window_size=batch_size,
-        max_warmpool_size=num_generations,
+        max_warmpool_size=max_warmpool_size if max_warmpool_size is not None else num_generations,
```
wang2yn84: Is it better to use max_concurrency here?

examples/deepswe/train_maxtext_nb.py
```diff
@@ -0,0 +1,1174 @@
+# %%
```
wang2yn84, line 15: For new file, please have the copyright section.

examples/deepswe/train_maxtext_nb.py
```diff
@@ -0,0 +1,1174 @@
...
+# [WIP] Reproduction of [DeepSWE](https://www.together.ai/blog/deepswe)
```
wang2yn84, line 16: I'm not completely objecting to have a new training script, but switching to the MaxText model should not change that much? Is it possible to use the same script? Any other reason why you want a dedicated training script.

Dockerfile
```diff
@@ -32,13 +32,21 @@ RUN pip install wandb
...
+# Copy scripts and requirements first to leverage Docker cache
+COPY scripts/install_tunix_vllm_requirement.sh scripts/
+COPY requirements/ requirements/
+
+RUN bash scripts/install_tunix_vllm_requirement.sh
 
-# Install the project in editable mode
-RUN pip install -e .
+# Copy pyproject.toml and README.md to install dependencies first
+COPY pyproject.toml README.md /app/
```
wang2yn84, line 42: Interesting, any reason to install deps firrst?

wang2yn84:
Hi Nitin, please squash the commits into one before submission.

examples/deepswe/swe_env.py
niting: That will likely not be right? max_concurrency for my script is 256, but I want two replicas per stream so I set max_warmpool_size to 2 so that I have 512 sandboxes with 256 on standby at a time.

examples/deepswe/train_maxtext_nb.py
niting: I was doing that earlier but started running into way too many if and else's throughout the script which muddled the control flow. That's the only reason for using a separate script. We might be able to delete this eventually, but want to keep this around for now while we work on newer recipes.

Dockerfile
niting: That's just because of how docker caches its layers. These dependencies don't change over time but the subsequent layers do. When these are installed first, docker leverages caching better and the images seem to build faster. Can revert this change if you think that might not be the case and I am skipping cache for some other reason.

examples/deepswe/train_maxtext_nb.py
niting: Done.

Dockerfile
wang2yn84: Thanks for the explanation, that makes sense to me.

examples/deepswe/train_maxtext_nb.py
wang2yn84: Let's land this first and see how it goes. It's a bit misleading for users when we have multiple script for the same purpose.

examples/deepswe/swe_env.py
wang2yn84: I see, so the max_warmpool_size is per image. The naming is a bit misleading. nit, maybe max_warmpool_per_stream or per_image?

niting:
This was merged separately: https://github.com/google/tunix/blob/main/examples/deepswe/train_maxtext_nb.py, due to some shenanigans with our internal (Google) importer. Closing this PR out.

PR #2054, opened 2026-08-31
Bootstrap rollout policy from target state

wang2yn84:
This PRs plumbs through the RL system transfer target state metadata from sampler -> rollout -> orchestrator -> trainer. We also need to sync weights once at step 0 to align the trainer and rollout policy.

**Reference**

**Colab Notebook**

**Checklist**

- [ ] I have followed [Contribution Guidelines](https://github.com/google/tunix/blob/main/docs/contributing.md).

### DIFF
--- a/tunix/experimental/orchestrator/distributed_rl_engine.py
+++ b/tunix/experimental/orchestrator/distributed_rl_engine.py
@@ -551,10 +551,42 @@ class DistributedRLEngine(rl_engine_interface.AbstractRLEngine):
         raise ValueError(f"No worker registered for role {role}")
       return await self._invoke_worker(worker, "get_metrics", **kwargs)
 
+  async def prepare_rollout_policy(
+      self,
+      role: datatypes.Role = datatypes.Role.ACTOR,
+      sync_weights: bool = True,
+      policy_version: int | None = None,
+  ) -> int | None:
+    """Bootstraps the rollout-visible policy state before step 0."""
+    trainer = self._trainer_workers.get(role)
+    if trainer is None:
+      raise ValueError(f"No trainer worker registered for role {role}")
+
+    if self._rollout_workers:
+      rollout = self._rollout_workers[0]
+      try:
+        target_state = await self._invoke_worker(rollout, "get_target_state")
+        await self._invoke_worker(
+            trainer, "set_target_state", target_state=target_state
+        )
+      except RuntimeError as exc:
+        if "AttributeError" not in str(exc):
+          raise
+
+    if not sync_weights:
+      return None
+    target_policy_version = (
+        self._policy_version if policy_version is None else policy_version
+    )
+    return await self.sync_weights(
+        role=role, policy_version=target_policy_version
+    )
+
   async def sync_weights(  # pyrefly: ignore[bad-override]
       self,
       role: datatypes.Role = datatypes.Role.ACTOR,
       target_roles: Sequence[datatypes.Role] | None = None,
+      policy_version: int | None = None,
   ) -> int:
     """Runs one weight sync round through the coordinator."""
     del role, target_roles
@@ -563,12 +595,13 @@ class DistributedRLEngine(rl_engine_interface.AbstractRLEngine):
           "sync_weights needs a coordinator; construct the engine with"
           " weight_sync_coordinator."
       )
+    next_policy_version = self._policy_version + 1 if policy_version is None else policy_version
     logging.info(
-        "Synchronizing weights (advancing to policy_version=%d)...",
-        self._policy_version + 1,
+        "Synchronizing weights (target policy_version=%d)...",
+        next_policy_version,
     )
     result = await self._weight_sync_coordinator.sync(
-        policy_version=self._policy_version + 1
+        policy_version=next_policy_version
     )
     self._policy_version = result.policy_version
     logging.info(
--- a/tunix/experimental/orchestrator/rl_engine_interface.py
+++ b/tunix/experimental/orchestrator/rl_engine_interface.py
@@ -135,10 +135,20 @@ class AbstractRLEngine(Protocol):
     """Retrieves step metrics from the worker for the specified role."""
     ...
 
+  async def prepare_rollout_policy(
+      self,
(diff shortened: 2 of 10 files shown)


tunix/experimental/worker/rollout_worker.py
```diff
@@ -114,6 +114,16 @@ def info(self) -> datatypes.WorkerInfo:
...
 
   def initialize(self) -> datatypes.Response:
+    if self.state == WorkerState.READY:
+      return datatypes.Response(
+        metadata={
+          "worker_id": self.worker_id,
+          "state": self.state.value,
+          "policy_version": self._policy_version,
+          "initialized": True,
+          "already_ready": True,
```
tianshub: lol shall we rename already_ready to be just ready?

tunix/experimental/worker/rollout_worker.py
```diff
@@ -477,7 +493,14 @@ async def weight_sync(self, sync_request: Any = None, **kwargs) -> Any:
     metadata = kwargs.pop("metadata", None)
     request = sync_request if sync_request is not None else metadata
     result = await self.manager.weight_sync(request, **kwargs)
-    self._policy_version += 1
+    if isinstance(result, int):
+      self._policy_version = result
+    else:
```
tianshub, line 507: in what situation we will hit this branch? shouldn't weight sync always return a valid version?

tunix/experimental/orchestrator/rl_program.py
```diff
@@ -732,6 +732,16 @@ async def run_async(
     self.engine = engine
     logging.info("Starting StandardRLProgram concurrent stages...")
 
+    if self.sync_weights:
+      prepare_rollout_policy = getattr(engine, "prepare_rollout_policy", None)
+      if callable(prepare_rollout_policy):
```
tianshub: can we avoid this branch since the rl_engine interface is updated, we always expect this function right?

tunix/experimental/orchestrator/rl_engine_interface.py
```diff
@@ -135,10 +135,20 @@ async def get_metrics(
     """Retrieves step metrics from the worker for the specified role."""
     ...
 
+  async def prepare_rollout_policy(
```
tianshub, line 138: should this be called something like setup_weight_sync?

tunix/experimental/orchestrator/distributed_rl_engine.py
```diff
@@ -516,10 +516,36 @@ async def get_metrics(
...
+      self,
+      role: datatypes.Role = datatypes.Role.ACTOR,
+      sync_weights: bool = True,
+  ) -> int | None:
+    """Bootstraps the rollout-visible policy state before step 0."""
+    trainer = self._trainer_workers.get(role)
+    if trainer is None:
+      raise ValueError(f"No trainer worker registered for role {role}")
+
+    if self._rollout_workers:
```
tianshub, line 565: this suppose to wait for rollout workers to be ready?

tunix/experimental/worker/rollout_worker.py
wang2yn84: In our protocol, the weight_sync return value is (str, none, any), so it's not guaranteed to be version

wang2yn84: And we control the weight sync version here: https://github.com/google/tunix/blob/7e83930703760e5158a4f867783e127b2b655807/tunix/experimental/orchestrator/distributed_rl_engine.py#L571. So we don't really need the weight sync to return the new policy version.

tunix/experimental/orchestrator/rl_engine_interface.py
wang2yn84: This API right now does 2 things: 1. Get the target state metadata from rollout and transfer to trainer 2. Does weight sync once. So it's not just setup weight sync but actually sync weight once. Prepare_rollout_policy feels more accurate?

tunix/experimental/orchestrator/distributed_rl_engine.py
wang2yn84: Hmm, the call is synchronous so should be fine, but the bigger problem is there is no lock to the rollout worker initialization function. There could be racing conditions. Added a lock so that it won't be init twice.

tunix/experimental/orchestrator/rl_program.py
wang2yn84: Yes, done!

tunix/experimental/worker/rollout_worker.py
wang2yn84: Yea, trainer did it first lol. Changed to "ready" in both rollout and trainer worker.

tunix/experimental/orchestrator/distributed_rl_engine.py
tianshub: but wouldn't rollout worker stay in initialization state until everything is ready? I think we can keep on querying the state of the rollout worker until it's ready?

tunix/experimental/orchestrator/rl_engine_interface.py
tianshub: prepare_rollout_policy sounds like setting up the rollout worker, but I'm fine with either way. Another option can be setup_weight_sync_group.

tunix/experimental/worker/rollout_worker.py
tianshub: how come weight sync can return any? it will be good if we can avoid if-else branches in weight sync since this mechanism is already fairly complex

tianshub (approved):

tunix/experimental/worker/rollout_worker.py
wang2yn84: Will simplify and fix in the follow up PR.

tunix/experimental/orchestrator/distributed_rl_engine.py
s-noghabi: yes, it will stay in init() until ready:

https://github.com/google/tunix/blob/main/tunix/experimental/orchestrator/lifecycle.py#L68

PR #2061, opened 2026-09-01
decouple force_on_policy_ratio from step0 reforward.

entrpn:
Measured on qwen3-4b (OpenMathInstruct-2, agentic rollout):
  - clipfrac       0.0006 -> 0.0000
  - log_ratio_abs  0.0157 -> 0.0000
  - step time      984s   -> 742s (~24%)

Gradient magnitude is essentially unchanged; clipping was only firing on
~0.06% of tokens, all of it kernel noise rather than policy drift.

Also in this change:
- log_sampler_trainer_agreement: opt-in flag that spends one trainer
  forward pass to keep the sampler/trainer agreement metrics alive.
  Without it, enabling force_on_policy_ratio silently drops them.
- force_on_policy_ratio + off_policy_steps > 0 now warns instead of
  raising. The MLPerf reference pairs the two deliberately.
- Compare completion_tokens[-1] against EOS when counting clipped
  completions; the mask was being compared against a vocab id, so
  clip_ratio over-counted.
- Buffer metrics once per group instead of once per env/reward-time
  sub-key. Previously duplicated every metric, or dropped all of them
  when no timing keys were present.

Matches the MLPerf Qwen3.5-397B GRPO reference recipe.

**Reference**

Same implementation as nemo_rl for force_on_policy_ratio, found here: https://github.com/NVIDIA-NeMo/RL/blob/main/nemo_rl/algorithms/loss/loss_functions.py#L520C9-L525C36

**Checklist**

### DIFF
--- a/tests/rl/agentic/agentic_grpo_learner_test.py
+++ b/tests/rl/agentic/agentic_grpo_learner_test.py
@@ -109,10 +109,13 @@ def _mock_generate(
     else:
       prompt_tokens.append(tokenizer.encode(" ".join(m["content"] for m in p)))
   max_p_len = max(len(pt) for pt in prompt_tokens)
-  padded_prompts = np.array([
-      np.pad(pt, (max(0, max_p_len - len(pt)), 0), constant_values=0)
-      for pt in prompt_tokens
-  ], dtype=np.int32)
+  padded_prompts = np.array(
+      [
+          np.pad(pt, (max(0, max_p_len - len(pt)), 0), constant_values=0)
+          for pt in prompt_tokens
+      ],
+      dtype=np.int32,
+  )
   return base_rollout.RolloutOutput(
       text=text,
       tokens=tokens,
@@ -585,9 +588,7 @@ class AgenticGrpoLearnerTest(parameterized.TestCase):
         chat_parser=MockChatParser(),
     )
 
-    train_ds = _dummy_dataset(
-        MySource(data=["1", "2"], repeat=1), batch_size=2
-    )
+    train_ds = _dummy_dataset(MySource(data=["1", "2"], repeat=1), batch_size=2)
 
     with (
         mock.patch.object(
@@ -639,9 +640,7 @@ class AgenticGrpoLearnerTest(parameterized.TestCase):
       def __init__(self, *, rngs: nnx.Rngs):
         self.lm_head = 1
 
-      def __call__(
-          self, inputs, positions, cache, attention_mask, **kwargs
-      ):
+      def __call__(self, inputs, positions, cache, attention_mask, **kwargs):
         del kwargs
         return (
             jnp.full(
@@ -703,9 +702,7 @@ class AgenticGrpoLearnerTest(parameterized.TestCase):
       def __init__(self, *, rngs: nnx.Rngs):
         self.lm_head = 1
 
-      def __call__(
-          self, inputs, positions, cache, attention_mask, **kwargs
-      ):
+      def __call__(self, inputs, positions, cache, attention_mask, **kwargs):
         del kwargs
         return (
             jnp.full(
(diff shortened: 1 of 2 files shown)


tests/rl/agentic/agentic_grpo_learner_test.py
```diff
@@ -2375,6 +2375,185 @@ def _patch_process_results(
...
+      train_example = results[0]
+
+      # 1. Asserts 0 extra trainer forward passes!
+      mock_get_actor_logps.assert_not_called()
+
+      # 2. Asserts old_per_token_logps is None (forces ratio to 1.0 via stop_gradient in loss)
+      self.assertIsNone(train_example.old_per_token_logps)
+
+
+  def test_compute_packed_logps_force_on_policy_skips_recompute(self):
```
tianshub: I guess we can remove this test since logps recompute has nothing to do with seq packing, so we save some testing overhead.

tunix/rl/agentic/agentic_grpo_learner.py
```diff
@@ -124,6 +124,7 @@ class GRPOConfig(agentic_rl_learner.AgenticRLConfig):
   # significantly with the rollout sampler. Without this correction, importance
   # ratios computed directly against the sampler's logp can spike on outlier
   # tokens, producing large-variance gradient updates.
+  force_on_policy_ratio: bool = False
```
tianshub: this should be moved before L115 with a comment. e.g. An optimized version of recomputing actor logps when num_iteration equals to 1, instead of recomputing, we will directly use detached logps from actor model.

tunix/rl/agentic/agentic_grpo_learner.py
```diff
@@ -646,7 +661,11 @@ def _process_results(
 
     rollout_per_token_logps = None
     trainer_per_token_logps = None
-    if self.algo_config.use_rollout_logps and padded_old_logprobs:
+    if self.algo_config.force_on_policy_ratio:
+      # Bypass both rollout logprobs and Step-0 Re-Forward passes.
+      # Settting this to None instructs grpo_loss_fn to use stop_gradient
```
tianshub: setting, also `setting this to None instructs grpo_loss_fn to use stop_gradient` isn't accurate, the loss fn also depends on the knob value, we can just remove this comment

tunix/rl/agentic/agentic_grpo_learner.py
```diff
@@ -680,6 +699,7 @@ def _process_results(
     elif is_packed:
       old_per_token_logps = None
     else:
+      # Step-0 Re-Forward path (use_rollout_logps=False, force_on_policy_ratio=False)
```
tianshub: this is not step-0 only, so we should remove the term

tianshub (changes requested):

tunix/rl/algo_core.py
```diff
@@ -430,7 +430,9 @@ def grpo_loss_fn(
   # TODO(tsbao): We should handle token level advantages.
   advantages = jnp.astype(train_example.advantages, jnp.float32)
 
-  if train_example.old_per_token_logps is None:
+  force_on_policy = getattr(algo_config, "force_on_policy_ratio", False)
+
+  if force_on_policy or train_example.old_per_token_logps is None:
```
tianshub: actually do you need to check on force_on_policy? since old_per_token_logps won't be populated with force_on_policy

tunix/rl/agentic/agentic_grpo_learner.py
lc5211: Can we add docstring for force_on_policy_ratio?

tunix/rl/algo_core.py
lc5211: +1, old_per_token_logps is already None with force_on_policy.

tests/rl/agentic/agentic_grpo_learner_test.py
entrpn: done.

tunix/rl/agentic/agentic_grpo_learner.py
entrpn: done.

tunix/rl/agentic/agentic_grpo_learner.py
entrpn: done.

tunix/rl/agentic/agentic_grpo_learner.py
entrpn: done.

tunix/rl/agentic/agentic_grpo_learner.py
entrpn: done.

tunix/rl/algo_core.py
entrpn: good call, I removed it.

entrpn:
@lc5211 @tianshub resolved your comments, PTAL.

tunix/rl/agentic/agentic_grpo_learner.py
```diff
@@ -83,6 +83,17 @@ class GRPOConfig(agentic_rl_learner.AgenticRLConfig):
     max_concurrency: Maximum number of concurrent rollout engines.
     off_policy_steps: Number of off-policy steps can be accepted before a policy
       update.
+    use_rollout_logps: Use the rollout engine's log-probabilities as
+      old_per_token_logps. False makes the trainer recompute them.
+    force_on_policy_ratio: When num_iterations == 1, use
+      stop_gradients(current_logp) as old_per_token_logps instead of recomputing
+      or using rollout logps. Pins the surrogate ratio to 1.0, so clipping never
+      fires and sampler-vs-trainer numerical noise is removed from the ratio.
+    log_sampler_trainer_agreement: Optinally spend one extra trainer forward
```
tianshub: Optinally -> Optionally

tunix/rl/agentic/agentic_grpo_learner.py
```diff
@@ -112,6 +123,12 @@ class GRPOConfig(agentic_rl_learner.AgenticRLConfig):
       False  # Whether to mask out degenerate groups with all-0 advantages.
   )
   use_rollout_logps: bool = True
+  # Pin the surrogate ratio to 1.0 (old_logp := stop_gradient(current_logp)).
+  # Valid for single-iteration on-policy training only.
+  force_on_policy_ratio: bool = False
+  # Costs one trainer forward pass; keeps the sampler/trainer agreement metrics
+  # alive when force_on_policy_ratio would otherwise leave nothing to compare.
+  log_sampler_trainer_agreement: bool = False
```
tianshub, line 131: @lc5211 are there any concerns if we disable this? probably no, but just to be aware this is another knob if we want to see the logged metrics

tunix/rl/agentic/agentic_grpo_learner.py
```diff
@@ -573,7 +622,7 @@ def _process_results(
       )
       if (
           len(completion_tokens) >= max_response_length
-          and completion_mask[-1] != eos_value
+          and completion_tokens[-1] != eos_value
```
tianshub, line 625: ah wow, is this a bug before?

tunix/rl/agentic/agentic_grpo_learner.py
```diff
@@ -845,11 +911,11 @@ def _process_results(
             f"{prefix}/{sub_key}/max": (np.max(flat_vals), np.max),
             f"{prefix}/{sub_key}/min": (np.min(flat_vals), np.min),
         })
-        self.rl_engine.buffer_metrics_async(
-            metrics_to_log,  # pyrefly: ignore[bad-argument-type]
-            mode=mode,
-            step=expected_step,  # pyrefly: ignore[bad-argument-type]
-        )
+      self.rl_engine.buffer_metrics_async(
```
tianshub, line 914: why is this changed?

tunix/rl/agentic/agentic_grpo_learner.py
entrpn: yes this was there already. I added the fix.

tunix/rl/agentic/agentic_grpo_learner.py
entrpn: if timedicts was empty, we would lose the upstream metrics if we keep it inside the if statement. Also it just keeps running the update on flat_vals as the for loop updates it which is unnecessary. I can revert if, if you prefer, it doesn't affect the rest of the PR.

tunix/rl/agentic/agentic_grpo_learner.py
entrpn: fixed.

tunix/rl/agentic/agentic_grpo_learner.py
lc5211: It's used in combination with force_on_policy_ratio. 
```not self.algo_config.force_on_policy_ratio
            or self.algo_config.log_sampler_trainer_agreement``` in this implementation so we are good. maybe let's add comment in line 486 just for the context purpose on when we enforce training logps recomputation.

tunix/rl/agentic/agentic_grpo_learner.py
tianshub: gotcha

tianshub (approved):

PR #2073, opened 2026-09-01
Add OpenHands agent-sandbox to train_maxtext_nb.py for Qwen3.5-35B DeepSWE training

susanbao:
Resolves #

### Description

This PR adds support for the OpenHands agent scaffold and sandbox integration in DeepSWE training (`examples/deepswe/`).

Rather than treating OpenHands as merely an alternative prompt on top of the R2E gym environment or running in a generic shared container, this integration supports OpenHands container sandboxes running directly on per-task benchmark images (e.g., R2E task images):
1. **Init Container Binary Injection**: OpenHands server binaries (`openhands-agent-server`) are injected into the task pod via an `initContainer` (`oh-server`) using a shared `emptyDir` volume (`/oh`), allowing the main container to run the unmodified per-task benchmark environment with the repository pre-installed under `/testbed`.
2. **Dual Adapter Binding & Ground-Truth Evaluation**: The environment binds both `agent_sandbox_rl`'s OpenHands workspace client (`make_handle_workspace`) for agent interaction (`execute_bash`, `str_replace_editor`, `submit`) and `make_fleet_repo_env` for evaluation, delegating grading directly to `FleetRepoEnv.compute_reward` (`self.final_reward_fn = self.env.compute_reward`) for authoritative benchmark reward calculation.
3. **Modular OpenHands Utilities**: Workspace configuration (`setup_openhands_workspace`) and action dispatch (`step_openhands`) are cleanly isolated in `openhands_utils.py`.
4. **Pipelined Batch Prewarming & Buffering**: `PrewarmDatasetIterator` manages in-flight batch windows to prevent premature pod teardown while overlapping rollouts are still active.

*(Note: Experimental JAX sharding and checkpoint manager modifications previously bundled in this PR have been separated into a dedicated PR to maintain surgical scope.)*

---

#### 1. OpenHands Agent Scaffold (`examples/deepswe/swe_agent.py`)
- **Changes**:
  - Added `"openhands"` as an accepted scaffold option in `SWEAgent.__init__` validation (`["r2egym", "sweagent", "openhands"]`).
  - Extracted prompt template definitions and helper accessors into `template.py`, while re-exporting all prompt constants (`SWE_SYSTEM_PROMPT`, `SWE_SYSTEM_PROMPT_FN_CALL`, `SWEAGENT_SYSTEM_PROMPT`, `OPENHANDS_SYSTEM_PROMPT`, `SWE_USER_PROMPT`, `SWE_USER_PROMPT_FN_CALL`, `SWEAGENT_USER_PROMPT`) and accessors (`get_system_prompt`, `get_user_prompt_template`) for backward compatibility.
  - Dynamically resolves `system_prompt` and `user_prompt_template` based on the configured scaffold.
  - Added token budget awareness in `update_from_env`: triggers at `TOKEN_WARNING_THRESHOLD = 28000`, urging the model to stop exploration and emit a final `<function=submit>` or `<function=finish>` call.
  - Streamlined dependency handling: removed fallback action shims and strictly imports `Action as SWEAction` from `r2egym.agenthub.action`.
- **Reason**:
  - Decouples prompt maintenance from agent orchestration logic, avoiding code bloat in `swe_agent.py` while preserving existing public symbols.
  - Enables prompt specialization for OpenHands (`execute_bash`, `str_replace_editor`, `submit`) and guards against token limit truncations during rollout.

#### 2. Centralized Templates & Pod Specifications (`examples/deepswe/template.py`)
- **Changes**:
  - Created a dedicated `template.py` module defining system and user prompts across `r2egym`, `sweagent`, and `openhands` scaffolds.
  - Defined `OPENHANDS_SYSTEM_PROMPT` containing full tool specifications for `execute_bash`, `str_replace_editor`, and `submit`, along with XML-style function call conventions and structured problem-solving workflows.
  - Added `DEFAULT_OPENHANDS_KEEPALIVE_CMD` with adaptive fallback: verifies execution permissions on `/oh/openhands-agent-server`, establishes bidirectional symlinks between `/workspace` and `/testbed`, configures git `safe.directory '*'`, changes directory to `/testbed`, and starts `openhands-agent-server --host 0.0.0.0 --port 8000`.
  - Implemented `get_openhands_pod_template(node_selector)`:
    - Constructs `agent_sandbox_rl.TemplateSpec` incorporating the keepalive command.
    - Configures an `initContainers` named `oh-server` (using `OPENHANDS_SERVER_IMAGE`, default `gcr.io/cloud-tpu-multipod-dev/sanbao/openhands-agent-server:1.44.1`) to copy server binaries into an `emptyDir` volume mounted at `/oh`.
    - Mounts `/oh` into the main task container, opens port 8000, and attaches an HTTP readiness probe on `/health` (period: 2s, failureThreshold: 150).
    - Specifies resource requests (`SANDBOX_CPU`, `SANDBOX_MEM`) and limits (`SANDBOX_CPU_LIMIT`, `SANDBOX_MEM_LIMIT`).
    - Configures session API key authentication (`OH_SESSION_API_KEYS_0` via `SANDBOX_SESSION_KEY`) and supports custom keepalive overrides via `AGENT_SERVER_COMMAND`.
  - Implemented `get_template(scaffold, node_selector)` returning the OpenHands pod spec for `scaffold == "openhands"` or `None` for other scaffolds.
- **Reason**:
  - Eliminates the need for monolithic, pre-baked OpenHands task images by dynamically injecting the server into standard benchmark containers via Kubernetes init containers.
  - Centralizes prompt templates and sandbox pod specifications in a clean, reusable module.

#### 3. OpenHands Utilities & Step Execution (`examples/deepswe/openhands_utils.py`)
- **Changes**:
  - Created a new `openhands_utils.py` module isolating OpenHands workspace setup and step execution logic.
  - **Workspace Setup (`setup_openhands_workspace`)**:
    - Configures git identity (`openhands@agent.sandbox`) and registers `/testbed` and `/workspace` in git `safe.directory`.
    - Creates bidirectional symlinks (`/testbed <-> /workspace`) idempotently so tasks and commands work transparently regardless of directory conventions.
    - Avoids costly repo cloning since task benchmark containers already ship the pre-cloned repository at the target commit under `/testbed`.
  - **Action Dispatch (`step_openhands`)**:
    - Handles `finish` and `submit` as terminal actions (`done=True`).
    - Translates `str_replace_editor` calls to `file_editor` and delegates directly to `env.env.step()` to leverage R2E's built-in file editor implementation.
    - Executes `execute_bash` commands via `env.workspace.execute_command(f"(cd /testbed 2>/dev/null || cd /workspace) && {cmd}")`, capturing combined stdout and stderr on non-zero exit codes.
    - Validates tool names and parameters, returning clear error observations for invalid calls.
- **Reason**:
  - Keeps `swe_env.py` lightweight and maintainable by modularizing OpenHands-specific execution pathways.
  - Translating `str_replace_editor` to `file_editor` allows OpenHands agents to edit files using standard tool calls while reusing verified R2E editor mechanics.

#### 4. OpenHands Agent-Sandbox & Fleet Integration (`examples/deepswe/swe_env.py`)
- **Changes**:
  - **Execution on Task Images**:
    - Runs OpenHands sandbox pods directly on per-task benchmark images instead of forcing a generic shared server image.
    - `_normalize_tasks_for_fleet` preserves task-specific Docker images and ds metadata across all scaffolds.
  - **Dual Adapter Binding & Grading Delegation**:
    - Split `_initial_observation` into `_init_agent_sandbox_env` and `_init_local_repo_env`.
    - In `_init_agent_sandbox_env` for OpenHands:
      - Instantiates `agent_sandbox_rl.adapters.openhands.make_handle_workspace` for remote pod interaction.
      - Simultaneously binds `agent_sandbox_rl.adapters.r2egym.make_fleet_repo_env` to the same pod handle for evaluation.
      - Binds `self.final_reward_fn = self.env.compute_reward`, ensuring ground-truth benchmark reward computation.
      - Calls `openhands_utils.setup_openhands_workspace` to initialize git configuration and path symlinks.
  - **Fleet Initialization & Sizing (`_init_global_fleet`)**:
    - Plumbs `scaffold` and attaches `template_mod.get_template(scaffold, node_sel)` to the fleet configuration.
    - Sized warm pools per task (`warm_per_task=True`, `max_warmpool_size=num_generations`) with `template_name_prefix="oh-img-"`.
    - Added configurable image rewriting via `_get_image_rewrite_fn` and `IMAGE_REWRITE_PREFIX`.
    - Registered `_teardown_global_fleet` via `atexit` to clean up sandbox pods on exit.
  - **Pipelined Prewarming with In-Flight Buffering (`PrewarmDatasetIterator`)**:
    - Replaced single previous-batch tracking with an in-flight batch buffer (`in_flight_batches`, `buffer_size = 2`) to ensure images are only unwarmed after completely finishing execution across overlapping rollout windows.
    - Applies `image_rewrite` uniformly to extracted images.
  - **Acquire Retries & Cleanup**:
    - Added 5 acquire retries with linear backoff (5s, 10s, 15s, 20s, 25s) on `fleet.acquire(task)` to tolerate transient Kubernetes scheduling/network blips.
    - Added cleanup for both `self.workspace` and `self.env` in `SWEEnv.close()`.
- **Reason**:
  - Resolves the structural zero-reward issue caused by generic shared server images by running real benchmark task images and delegating evaluation to the official `FleetRepoEnv.compute_reward`.
  - In-flight buffering prevents active pods from being torn down prematurely during pipelined rollout transitions.

#### 5. DeepSWE Training Script (`examples/deepswe/train_maxtext_nb.py`)
- **Changes**:
  - Added `--scaffold` command-line argument with choices `["r2egym", "sweagent", "openhands"]` (defaulting to `"r2egym"`).
  - Forwarded `scaffold=args.scaffold` to `_init_global_fleet`, `PrewarmDatasetIterator`, `SWEAgent` (`agent_kwargs`), and `SWEEnv` (`env_kwargs`).
  - Wrapped `agentic_grpo_learner.train(...)` in a `try ... finally` block to guarantee `swe_env._teardown_global_fleet()` executes on completion or exception.
  - Cleaned up obsolete `allow_split_physical_axes` from trainer and sampler configs.
- **Reason**:
  - Exposes scaffold selection directly via CLI while preserving backward compatibility with `r2egym` as default.
  - Guarantees cluster sandbox pods and warm pools are freed even if training encounters an unhandled exception or early termination.

#### 6. Comprehensive Unit Test Suite (`tests/examples/swe_env_test.py` & `tests/examples/template_test.py`)
- **Changes**:
  - **`tests/examples/template_test.py`**:
    - Tests presence of prompt constants and accessors across all scaffolds (`test_prompt_constants_exist`, `test_get_system_prompt`, `test_get_user_prompt_template`).
    - Verifies default OpenHands `TemplateSpec` construction including `initContainers`, volumes, ports, probes, and resource limits/requests (`test_get_openhands_pod_template_default`).
    - Verifies custom environment variable overrides (`test_get_openhands_pod_template_with_overrides`).
    - Verifies `get_template` returns OpenHands spec or `None` appropriately (`test_get_template_scaffolds`).
  - **`tests/examples/swe_env_test.py`**:
    - Verifies task image extraction and warm pool sizing per task (`test_extract_images_always_uses_task_image`, `test_warm_batch_per_task_image_sizing`).
    - Verifies task normalization and image rewriting (`test_normalize_tasks_keeps_task_image`, `test_extract_images_with_image_rewrite`, `test_load_tasks_image_rewrite_applies_to_all_scaffolds`).
    - Verifies OpenHands workspace setup git config and symlinks (`test_setup_openhands_workspace_symlink_and_safe_directory`).
    - Verifies dual adapter binding (`make_handle_workspace` and `make_fleet_repo_env`) and error handling on bind failure (`test_openhands_initial_observation_binds_workspace_and_env`, `test_fleet_repo_env_bind_failure_raises`).
    - Verifies step execution for `submit`, `execute_bash`, and `str_replace_editor` delegation and reward calculation (`test_openhands_step_impl_submit`, `test_openhands_step_impl_execute_bash`, `test_openhands_step_impl_str_replace_editor_and_reward`, `test_openhands_utils_step_openhands_direct`).
    - Verifies acquire retry loop on transient errors (`test_acquire_retry_on_transient_error`).
- **Reason**:
  - Provides complete test coverage for OpenHands agent-server injection, workspace setup, dual adapter interaction, tool delegation, and robust error recovery without requiring live Kubernetes clusters.

#### 7. Scope Isolation (Reverted SFT Sharding/Checkpoint Manager Changes)
- **Changes**:
  - Reverted intermediate commits touching `tunix/sft/sharding_utils.py` and `tunix/sft/checkpoint_manager.py`.
- **Reason**:
  - Keeps this PR surgical and strictly focused on DeepSWE OpenHands scaffold support, moving unrelated JAX sharding adjustments into a dedicated pull request.

---

### Checklist

### DIFF
--- a/examples/math_gsm8k/qwen3_grpo_demo.py
+++ b/examples/math_gsm8k/qwen3_grpo_demo.py
@@ -34,10 +34,8 @@ import gc
 import logging
 import math
 import os
-import re
 import sys
 import time
-from typing import Any
 
 from absl import logging as absl_logging
 
@@ -70,7 +68,6 @@ from flax import nnx
 import jax
 from jax import numpy as jnp
 from jax.sharding import Mesh
-import numpy as np
 import optax
 from orbax import checkpoint as ocp
 import tensorflow_datasets as tfds
@@ -133,6 +130,7 @@ from tunix.rl import rl_cluster as rl_engine_lib
 from tunix.rl import utils as rl_utils
 from tunix.rl.agentic.agentic_grpo_learner import GRPOConfig, GRPOLearner
 from tunix.rl.agentic.parser.chat_template_parser import parser as chat_parser_lib
+from tunix.utils import gsm8k_vtc
 from tunix.rl.rollout import base_rollout
 from tunix.sft import metrics_logger
 from tunix.sft import utils as sft_utils
@@ -237,15 +235,22 @@ if ENABLE_CHECKPOINTING:
 
 show_hbm_usage = sft_utils.show_hbm_usage
 
-VTC_PROMPT_TEMPLATE = """Solve the following math problem.
-First, put your detailed step-by-step reasoning process inside <reasoning>...</reasoning> tags.
-Then, put your final numerical answer inside <answer>\\boxed{{}}</answer> tags. Do not put anything else in the answer tags.
-
-Problem: {}
-<reasoning>
-"""
-
-_metric_call_idx = 0
+# The GSM8K VTC recipe (prompt template, reward, metrics, raw parser) lives in
+# tunix.utils.gsm8k_vtc and is shared with the distributed experimental example
+# so the two stay in parity by construction.
+VTC_PROMPT_TEMPLATE = gsm8k_vtc.VTC_PROMPT_TEMPLATE
+_as_text = gsm8k_vtc.as_text
+extract_hash_answer = gsm8k_vtc.extract_hash_answer
+build_prompt = gsm8k_vtc.build_prompt
+_normalize_example_value = gsm8k_vtc.normalize_example_value
+normalize_single_example = gsm8k_vtc.normalize_single_example
+extract_boxed_answer = gsm8k_vtc.extract_boxed_answer
+is_vtc_format_correct = gsm8k_vtc.is_vtc_format_correct
+normalize_answer = gsm8k_vtc.normalize_answer
+_vtc_completion_outcome = gsm8k_vtc.vtc_completion_outcome
+vtc_env_reward = gsm8k_vtc.vtc_env_reward
+vtc_metric_fn = gsm8k_vtc.vtc_metric_fn
+VTCRawTextParser = gsm8k_vtc.VTCRawTextParser
 
 
 # ====== Shared Mesh ======
@@ -272,20 +277,6 @@ print(f"shared_mesh.devices.shape={shared_mesh.devices.shape}")
 
 
 # ====== Data ======
-def _as_text(value: Any) -> str:
-  return value if isinstance(value, str) else value.decode("utf-8")
-
-
-def extract_hash_answer(text: str) -> str | None:
-  if "####" not in text:
-    return None
-  return text.split("####", 1)[1].strip()
-
-
-def build_prompt(question: str) -> str:
-  return VTC_PROMPT_TEMPLATE.format(question)
-
-
 def build_gsm8k_dataset(
     *,
     split: str,
@@ -334,155 +325,7 @@ def create_datasets() -> tuple[grain.MapDataset, grain.MapDataset]:
   return train_dataset, eval_dataset
 
 
-def _normalize_example_value(value: Any) -> Any:
-  if isinstance(value, np.ndarray):
-    flat = value.reshape(-1).tolist()
-    if len(flat) == 1:
-      return _normalize_example_value(flat[0])
-    return [_normalize_example_value(v) for v in flat]
-  if isinstance(value, np.bytes_):
-    return value.tobytes().decode("utf-8")
-  if isinstance(value, bytes):
-    return value.decode("utf-8")
-  return value
-
-
-def normalize_single_example(example: dict[str, Any]) -> dict[str, Any]:
-  return {
-      key: _normalize_example_value(value) for key, value in example.items()
-  }
-
-
 # ====== Reward + Metrics ======
-def extract_boxed_answer(text: str) -> str | None:
-  answer_blocks = re.findall(r"<answer>(.*?)</answer>", text, re.DOTALL)
-  content = answer_blocks[-1] if answer_blocks else text
-
-  boxed = []
-  stack = []
-  for i, ch in enumerate(content):
-    if ch == "{":
-      stack.append(i)
-    elif ch == "}":
-      if not stack:
-        continue
-      open_idx = stack.pop()
-      if content[:open_idx].endswith(r"\boxed"):
-        boxed.append(content[open_idx + 1 : i].strip())
-  if boxed:
-    return boxed[-1]
-
-  fallback = re.search(r"\\boxed\s*\{?\s*([a-zA-Z0-9\.,\-]+)\s*\}?", content)
-  if fallback:
-    return fallback.group(1).strip()
-  return None
-
-
-def is_vtc_format_correct(text: str) -> bool:
-  has_reasoning = text.count("</reasoning>") == 1
-  has_answer = text.count("<answer>") == 1 and text.count("</answer>") == 1
-  reasoning_end = text.find("</reasoning>")
-  answer_open = text.find("<answer>")
-  answer_close = text.find("</answer>")
-  return (
-      has_reasoning
-      and has_answer
-      and reasoning_end != -1
-      and answer_open != -1
-      and answer_close != -1
-      and reasoning_end < answer_open < answer_close
-  )
-
-
-def normalize_answer(text: str | None) -> str | None:
-  if text is None:
-    return None
-  return str(text).replace(",", "").strip()
-
-
-def _vtc_completion_outcome(
-    completion: str, gold: Any
-) -> tuple[float, bool, bool, bool]:
-  format_ok = is_vtc_format_correct(completion)
-  pred = normalize_answer(extract_boxed_answer(completion))
-  true = normalize_answer(_normalize_example_value(gold))
-  answer_ok = pred is not None and true is not None and pred == true
-  extracted_ok = pred is not None
-
-  if format_ok and answer_ok:
-    score = 1.0
-  elif format_ok and not answer_ok:
-    score = 0.1
-  elif not format_ok and answer_ok:
-    score = 0.5
-  else:
-    score = 0.0
-  return score, format_ok, answer_ok, extracted_ok
-
-
-def vtc_env_reward(task, action):
-  gold = task.get("answer")
-  completion = action.action if hasattr(action, "action") else action
-  score, _, _, _ = _vtc_completion_outcome(completion, gold)
-  return score
-
-
-def vtc_metric_fn(prompts, completions, rewards, advantages, answer, **kwargs):
-  del prompts, completions, advantages, answer, kwargs
-  global _metric_call_idx
-  _metric_call_idx += 1
-
-  rewards = np.asarray(rewards, dtype=np.float32)
-  solve_all = bool(np.all(rewards > 0.1))
-  solve_none = bool(np.all(np.isclose(rewards, 0.0)))
-  solve_partial = (not solve_all) and (not solve_none)
-  solve_ratio = float(np.mean(rewards > 0.1))
-  reward_mean = float(rewards.mean())
-  reward_max = float(rewards.max())
-
-  absl_logging.info(
-      "[rollout-metric] call=%d n=%d solve_ratio=%.3f reward_mean=%.3f"
-      " reward_max=%.3f solve_all=%d solve_none=%d",
-      _metric_call_idx,
-      len(rewards),
-      solve_ratio,
-      reward_mean,
-      reward_max,
-      int(solve_all),
-      int(solve_none),
-  )
-  return {
-      "rewards/solve_all": (1 if solve_all else 0, np.mean),
-      "rewards/solve_none": (1 if solve_none else 0, np.mean),
-      "rewards/solve_partial": (1 if solve_partial else 0, np.mean),
-      "rewards/solve_ratio": (solve_ratio, np.mean),
-  }
-
-
-# ====== Tokenizer / Model ======
-class VTCRawTextParser:
-  """Raw-text prompt parser matching NeMo's vtc_raw_text_processor style."""
-
-  def parse(
-      self,
-      messages,
-      add_generation_prompt: bool = False,
-      is_first_msg: bool = False,
-  ) -> str:
-    del add_generation_prompt, is_first_msg
-    parts = []
-    for message in messages:
-      role = message.get("role")
-      content = message.get("content", "")
-      if role == "system" and content:
-        parts.append(content)
-      elif role == "user":
-        parts.append(content)
-      elif role == "assistant" and content:
-        parts.append(content)
-    return "\n".join(parts)
-
-
 class VTCGRPOLearner(GRPOLearner):
   """Demo-local learner that normalizes TFDS string payloads to Python str."""
 
--- a/tests/experimental/examples/math_gsm8k_dist/gsm8k_test.py
+++ b/tests/experimental/examples/math_gsm8k_dist/gsm8k_test.py
@@ -56,48 +56,56 @@ class GSM8KTest(absltest.TestCase):
     self.assertTrue(info["correct"])
     self.assertTrue(info["format_correct"])
 
-  def test_format_parity_with_non_experimental_recipe(self):
-    """is_gsm8k_format_correct must match examples/math_gsm8k's predicate.
+  def test_format_parity_with_qwen3_grpo_demo(self):
+    """is_gsm8k_format_correct must match qwen3_grpo_demo's is_vtc_format_correct.
 
-    Reasoning is satisfied by <reasoning>..</reasoning> OR native <think>..
-    </think>; the answer by \\boxed OR <answer>..</answer>.
+    Exactly one </reasoning>, exactly one <answer>..</answer>, in that order.
+    The opening <reasoning> tag is not required: the prompt template opens it,
+    so a completion that follows the prompt only carries the closing tag.
     """
-    # Qwen-native shape: <think> reasoning + \boxed answer, no custom tags.
+    # A completion that follows the prompt (closing tag only) is well-formed.
     self.assertTrue(
         gsm8k.is_gsm8k_format_correct(
-            "<think>2+2=4</think>The answer is \\boxed{4}"
+            "2+2=4\n</reasoning>\n<answer>\\boxed{4}</answer>"
         )
     )
-    # Explicit reasoning tags + <answer> block, no \boxed.
+    # Re-emitting the opening tag is also fine.
     self.assertTrue(
         gsm8k.is_gsm8k_format_correct(
-            "<reasoning>work</reasoning><answer>4</answer>"
+            "<reasoning>work</reasoning><answer>\\boxed{4}</answer>"
         )
     )
-    # Answer present but no reasoning of either kind -> not format-correct.
-    self.assertFalse(gsm8k.is_gsm8k_format_correct("The answer is \\boxed{4}"))
-    # Reasoning present but no answer of either kind -> not format-correct.
+    # Tags out of order, duplicated, or missing are not well-formed.
     self.assertFalse(
-        gsm8k.is_gsm8k_format_correct("<think>2+2=4</think> so it is four")
+        gsm8k.is_gsm8k_format_correct("<answer>\\boxed{4}</answer></reasoning>")
+    )
+    self.assertFalse(
+        gsm8k.is_gsm8k_format_correct(
+            "</reasoning></reasoning><answer>\\boxed{4}</answer>"
+        )
+    )
+    self.assertFalse(gsm8k.is_gsm8k_format_correct("The answer is \\boxed{4}"))
+    # Qwen-native <think> is not the recipe's format.
+    self.assertFalse(
+        gsm8k.is_gsm8k_format_correct("<think>2+2=4</think>\\boxed{4}")
     )
 
-  def test_extract_boxed_answer_falls_back_to_answer_block(self):
-    # No \boxed anywhere, but a non-empty <answer> block: use its contents.
-    self.assertEqual(
-        gsm8k.extract_boxed_answer("<answer>42</answer>"), "42"
-    )
-    # \boxed still wins when present.
+  def test_extract_boxed_answer_requires_boxed(self):
+    # Parity with examples/math_gsm8k/qwen3_grpo_demo.py: an <answer> block
+    # without \boxed{} is not an answer.
+    self.assertIsNone(gsm8k.extract_boxed_answer("<answer>42</answer>"))
     self.assertEqual(
         gsm8k.extract_boxed_answer("<answer>\\boxed{7}</answer>"), "7"
     )
 
-  def test_think_and_boxed_scores_as_correct(self):
-    # The real Qwen thinking-mode shape end to end: format + answer -> 1.0.
+  def test_think_and_boxed_scores_as_answer_only(self):
+    # Qwen thinking-mode shape: right answer, but not the recipe's format,
+    # so 0.5.
     reward, info = gsm8k.score_gsm8k_completion(
         "<think>2+2=4</think>\\boxed{4}", "4"
     )
-    self.assertEqual(reward, 1.0)
-    self.assertTrue(info["format_correct"])
+    self.assertEqual(reward, 0.5)
+    self.assertFalse(info["format_correct"])
     self.assertTrue(info["answer_correct"])
 
   def test_scores_formatted_wrong_answer_with_format_reward(self):
--- a/tests/utils/gsm8k_vtc_test.py
+++ b/tests/utils/gsm8k_vtc_test.py
@@ -1,0 +1,131 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for the shared GSM8K VTC recipe."""
+
+import types
+
+from absl.testing import absltest
+import numpy as np
+from tunix.utils import gsm8k_vtc
+
+
+class Gsm8kVtcTest(absltest.TestCase):
+
+  def test_prompt_opens_reasoning_block(self):
+    prompt = gsm8k_vtc.build_prompt("How many clips?")
+    self.assertIn("Problem: How many clips?", prompt)
+    self.assertTrue(prompt.endswith("<reasoning>\n"))
+    self.assertIn("<answer>\\boxed{}</answer>", prompt)
+
+  def test_extract_hash_answer(self):
+    self.assertEqual(gsm8k_vtc.extract_hash_answer("work #### 72"), "72")
+    self.assertIsNone(gsm8k_vtc.extract_hash_answer("no delimiter"))
+
+  def test_extract_boxed_answer(self):
+    self.assertEqual(
+        gsm8k_vtc.extract_boxed_answer("<answer>\\boxed{7}</answer>"), "7"
+    )
+    # Nested braces and the last block win.
+    self.assertEqual(
+        gsm8k_vtc.extract_boxed_answer(
+            "<answer>\\boxed{1}</answer><answer>\\boxed{\\frac{3}{4}}</answer>"
+        ),
+        "\\frac{3}{4}",
+    )
+    # Loose \boxed outside any answer block still counts.
+    self.assertEqual(gsm8k_vtc.extract_boxed_answer("so \\boxed 12 done"), "12")
+    # An answer block without \boxed is not an answer.
+    self.assertIsNone(gsm8k_vtc.extract_boxed_answer("<answer>42</answer>"))
+
+  def test_format_requires_closing_tag_and_ordered_answer_only(self):
+    # A completion that follows the prompt (closing tag only) is well-formed.
+    self.assertTrue(
+        gsm8k_vtc.is_vtc_format_correct(
+            "2+2=4\n</reasoning>\n<answer>\\boxed{4}</answer>"
+        )
+    )
+    self.assertTrue(
+        gsm8k_vtc.is_vtc_format_correct(
+            "<reasoning>work</reasoning><answer>\\boxed{4}</answer>"
+        )
+    )
+    self.assertFalse(
+        gsm8k_vtc.is_vtc_format_correct(
+            "<answer>\\boxed{4}</answer></reasoning>"
+        )
+    )
+    self.assertFalse(
+        gsm8k_vtc.is_vtc_format_correct(
+            "</reasoning></reasoning><answer>\\boxed{4}</answer>"
+        )
+    )
+    self.assertFalse(gsm8k_vtc.is_vtc_format_correct("\\boxed{4}"))
+    self.assertFalse(
+        gsm8k_vtc.is_vtc_format_correct("<think>2+2=4</think>\\boxed{4}")
+    )
+
+  def test_completion_outcome_tiers(self):
+    good = "2+2=4</reasoning><answer>\\boxed{4}</answer>"
+    self.assertEqual(gsm8k_vtc.vtc_completion_outcome(good, "4")[0], 1.0)
+    self.assertEqual(gsm8k_vtc.vtc_completion_outcome(good, "5")[0], 0.1)
+    self.assertEqual(
+        gsm8k_vtc.vtc_completion_outcome("\\boxed{4}", "4")[0], 0.5
+    )
+    self.assertEqual(gsm8k_vtc.vtc_completion_outcome("four", "4")[0], 0.0)
+    # Gold answers arrive as numpy / bytes from TFDS and with thousands commas.
+    self.assertEqual(
+        gsm8k_vtc.vtc_completion_outcome(
+            "</reasoning><answer>\\boxed{1,000}</answer>", np.bytes_(b"1000")
+        )[0],
+        1.0,
+    )
+
+  def test_env_reward_reads_action_and_gold_aliases(self):
+    action = types.SimpleNamespace(
+        action="</reasoning><answer>\\boxed{9}</answer>"
+    )
+    self.assertEqual(gsm8k_vtc.vtc_env_reward({"answer": "9"}, action), 1.0)
+    self.assertEqual(
+        gsm8k_vtc.vtc_env_reward({"gold_answer": "9"}, "\\boxed{9}"), 0.5
+    )
+
+  def test_metric_fn(self):
+    metrics = gsm8k_vtc.vtc_metric_fn(
+        None, None, [1.0, 0.5, 0.1, 0.0], None, None
+    )
+    self.assertEqual(metrics["rewards/solve_ratio"][0], 0.5)
+    self.assertEqual(metrics["rewards/solve_all"][0], 0)
+    self.assertEqual(metrics["rewards/solve_none"][0], 0)
+    self.assertEqual(metrics["rewards/solve_partial"][0], 1)
+
+  def test_raw_text_parser_joins_verbatim(self):
+    parser = gsm8k_vtc.VTCRawTextParser()
+    text = parser.parse(
+        [
+            {"role": "system", "content": ""},
+            {"role": "user", "content": gsm8k_vtc.build_prompt("Q?")},
+        ],
+        add_generation_prompt=True,
+        is_first_msg=True,
+    )
+    self.assertEqual(text, gsm8k_vtc.build_prompt("Q?"))
+    self.assertTrue(text.endswith("<reasoning>\n"))
+    tokens, extra = parser.update_assistant_end_tokens(np.array([1, 2]))
+    self.assertEqual(extra, 0)
+    np.testing.assert_array_equal(tokens, [1, 2])
+
+
+if __name__ == "__main__":
+  absltest.main()
--- a/tunix/experimental/examples/common/run_rollout_node.py
+++ b/tunix/experimental/examples/common/run_rollout_node.py
@@ -65,9 +65,25 @@ def _import_vllm_sampler():
 
 
 def _chat_parser_for(
-    model_id: str, tokenizer: Any, *, enable_thinking: bool = False
+    model_id: str,
+    tokenizer: Any,
+    mode: str = "auto",
+    *,
+    enable_thinking: bool = False,
 ):
-  """Selects the chat template parser by model family."""
+  """Selects the chat parser: `raw` text, or the model family's template.
+
+  Args:
+    model_id: Model name used to pick the family-specific template parser.
+    tokenizer: Tokenizer handed to the parser.
+    mode: Either `auto` (model family's template) or `raw` (no template).
+    enable_thinking: Whether the template parser opens a thinking block.
+
+  Returns:
+    The chat parser to use for this rollout worker.
+  """
+  if mode == "raw":
+    return chat_parser_lib.RawTextParser(tokenizer)
   name = model_id.lower()
   for family, parser_cls in CHAT_PARSERS.items():
     if family in name:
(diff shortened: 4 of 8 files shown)


andytwigg:
@susanbao can you rebase this against main please?

susanbao:
> @susanbao can you rebase this against main please?

Done!

examples/deepswe/swe_env.py
```diff
@@ -79,19 +80,28 @@ def _patched_start_container(
...
+      os.getenv("AGENT_SERVER_IMAGE") if scaffold == "openhands" else None
+  )
   normalized = []
   for item in tasks:
     if isinstance(item, Task):
       normalized.append(item)
     elif isinstance(item, dict):
-      img = item.get("docker_image") or item.get("image", "default")
+      img = (
+          agent_server_override
```
tomergee: With AGENT_SERVER_IMAGE set (L101), every task maps to the same image, so the fleet plans a single pool sized to num_generations (L201, L317) while the learner runs max_concurrency claims against it. Only the first 8 bind warm; the rest wait on the controller to replenish, which is a cold pod start per claim hidden behind the 900s ready timeout. Your own run config used --max_concurrency 64 with --max_warmpool_size 1.

On top of that, _setup_openhands_workspace git-clones the upstream repo from GitHub on every episode (L656). The R2E-Gym images already have the repo at the right commit with the env built in /testbed, so this throws away the whole warm-pool win.

Suggest keeping per-task images: build derived images FROM the R2E image with the agent-server added (the OpenHands Dockerfile supports BASE_IMAGE + the binary-minimal target, which installs to the same /usr/local/bin path you already run), map them via fleet.load_tasks(image_rewrite=...), set working_dir="/testbed", and drop the clone. That keeps per-image pools of G replicas and the prewarm iterator working as designed. If you keep a shared-image mode, the pool needs to be sized to max_concurrency, not G (see _pool_size in examples/run_openhands_fleet.py).

examples/deepswe/swe_env.py
```diff
@@ -486,28 +580,271 @@ def _initial_observation(self) -> Any:
...
+      else:
+        logging.info(
+            "[SWEEnv] Successfully set up repository %s in /workspace", repo
+        )
+    except Exception as e:
+      logging.warning(
+          "[SWEEnv] Failed to set up repository in /workspace: %s", e
+      )
+
+  def compute_reward(self, verbose: bool = False) -> float:
```
tomergee: For R2E-Gym rows this always returns 0.0. The dataset has no test_patch, eval_script, or test_command, and the generic agent-server image has no /run_tests.sh, so the last branch runs `exit 1` (L762). GRPO advantages will be all zero.

Even on a proper R2E image, exit code is not the label: R2E-Gym's _calculate_reward_r2e parses per-test statuses and compares them to expected_output_json (and the SWE-bench variants check FAIL_TO_PASS / PASS_TO_PASS). Easiest fix is to reuse that logic on the same pod, e.g. bind make_fleet_repo_env(handle) alongside the workspace just for compute_reward, or port the parse-and-compare onto workspace.execute_command.

Small related note: the container runs as the non-root `openhands` user, so the `ln -s /workspace /testbed` (L651) and the `/eval.sh` write (L726) fail silently and the /testbed paths never take effect.

examples/deepswe/swe_env.py
```diff
@@ -230,6 +282,10 @@ def __init__(
       pass
 
   def _extract_images(self, batch: Any) -> list[str]:
+    agent_server_override = os.getenv("AGENT_SERVER_IMAGE")
```
tomergee: This applies AGENT_SERVER_IMAGE unconditionally, but _normalize_tasks_for_fleet (L92) only applies it when scaffold == "openhands". With the env var set and the r2egym scaffold, the tasks are planned on their docker_image but the iterator warms the agent-server image instead. warm_images logs "not in the plan, skipped" and every real task image ends up on-demand. The two paths need the same gate.

tomergee (commented):
Took a pass focused on the agent-sandbox / warm-pool side of this. Main points inline.

examples/deepswe/swe_env.py
```diff
@@ -144,25 +155,66 @@ def _init_global_fleet(
     effective_max_concurrent = max(
         max_concurrency, batch_size * num_generations * 2
     )
-    fleet_cfg = FleetConfig(
-        clusters=[
+
+    template = None
+    if scaffold == "openhands":
+      session_key = os.getenv("SANDBOX_SESSION_KEY", "")
+      template = TemplateSpec(
```
tomergee: No resources set, so pods get the SDK defaults: 250m / 512Mi requests and no limits. That is under-provisioned for running a repo test suite for reward and unbounded for the model-generated commands. The reference template uses 500m / 1Gi requests with limits; pass a ResourceSpec here and set limits via extra_pod_spec.

examples/deepswe/swe_env.py
susanbao: Addressed: Gated `AGENT_SERVER_IMAGE` on `scaffold == 'openhands'` in `PrewarmDatasetIterator._extract_images`, and passed `scaffold=args.scaffold` into `PrewarmDatasetIterator`.

examples/deepswe/swe_env.py
susanbao: Addressed: Added `ResourceSpec(cpu='500m', memory='1Gi')` (configurable via env vars) to `TemplateSpec` and added resource limits (`cpu: 2`, `memory: 4Gi`) in `extra_pod_spec`.

examples/deepswe/swe_env.py
susanbao: Addressed:
1. Supported per-task images as the default (when `AGENT_SERVER_IMAGE` is unset) with optional `image_rewrite` forwarding in `fleet.load_tasks` (and `IMAGE_REWRITE_PREFIX` env support).
2. For shared-image mode (`AGENT_SERVER_IMAGE` set), sized the warm pool to `effective_max_concurrent` rather than `num_generations`.
3. Set `working_dir='/testbed'` for the OpenHands workspace.
4. Dropped redundant git clone in `_setup_openhands_workspace` when `/testbed` already exists (which R2E / SWE-bench images have pre-baked).

examples/deepswe/swe_env.py
susanbao: Addressed:
1. Bound `make_fleet_repo_env(self.handle)` as `self.env` alongside `self.workspace` when acquiring the pod, so `compute_reward` calls R2E-Gym's full grading and test-status comparison against `expected_output_json` on the same pod.
2. Removed the non-functional `ln -s /workspace /testbed` symlink attempt in favor of using `working_dir='/testbed'` directly, and added `git config --global --add safe.directory`.
3. Fixed non-root permission issue in fallback script execution by writing to `/tmp/eval.sh` and prioritizing `/testbed/run_tests.sh`.

tunix/rl/agentic/trajectory/trajectory_collect_engine.py
```diff
@@ -197,35 +197,45 @@ async def collect(self, mode: str = "Conversation") -> Any:
     Returns:
         Trajectory | dict | list: Depending on mode.
     """  # fmt: skip
-    await self._reset()
+    masked_out = False
+    try:
```
tianshub: why are we making this change?

tunix/sft/checkpoint_manager.py
```diff
@@ -58,13 +59,8 @@ def _replicate_if_pspec_uses_unknown_mesh_axis(
...
   if pspec is None:
     return jax.sharding.PartitionSpec()
-  for axis_name in pspec:
-    if axis_name is None:
-      continue
-    axis_names = axis_name if isinstance(axis_name, tuple) else (axis_name,)
-    if any(name is not None and name not in mesh.shape for name in axis_names):
-      return jax.sharding.PartitionSpec()
-  return pspec
+  cleaned = sharding_utils.deduplicate_mesh_axes(pspec, mesh)
```
tianshub: is this related to openhand env?

examples/deepswe/swe_env.py
```diff
@@ -144,30 +158,88 @@ def _init_global_fleet(
     effective_max_concurrent = max(
         max_concurrency, batch_size * num_generations * 2
     )
-    fleet_cfg = FleetConfig(
-        clusters=[
+
+    template = None
```
tianshub: can we put these template to a separate file? the env and agent class are getting over long now.

tianshub (changes requested):
please make sure only changes relevant to the environment is included in the PR

tunix/sft/checkpoint_manager.py
susanbao: No. It is for the preventing DuplicateSpecError which appeared during my exp. I can remove this in the final submission.

examples/deepswe/swe_env.py
susanbao: Great suggestion!

tunix/rl/agentic/trajectory/trajectory_collect_engine.py
susanbao: Have removed this change.

examples/deepswe/swe_env.py
tomergee: Confirmed, the iterator now gates on scaffold and the trainer passes it through. Resolving.

examples/deepswe/swe_env.py
tomergee: Looks good, requests plus limits with env overrides. Resolved

examples/deepswe/swe_env.py
tomergee: Thanks, binding make_fleet_repo_env on the same handle is the right fix, and I checked that its close() is a no-op so there is no double release. One caveat: this only grades correctly when the pod is running a real R2E image, which is the open question in the thread above (per-task images can't currently boot the agent-server, and shared mode has no /run_tests.sh). Resolving this one and keeping that part in the other thread.

examples/deepswe/swe_env.py
tomergee: Thanks for the changes. Checked the new head, this one isn't there yet:

1. Per-task mode (the new default) can't start: plain R2E images don't have tini or the agent-server, so the pods exit and the pool never goes Ready. Shared mode is still the only thing that runs.
2. In shared mode the clone still happens on every episode, because the generic image has no /testbed so the guard never skips it.
3. Shared-mode pool sizing now warms max(max_concurrency, batch*G*2) pods, which is 512 for your 64-concurrency run. Use max_concurrency.

Question: is it intentional that this doesn't use the OpenHands agent? It's the OpenHands prompt plus bash/submit, and the workspace is only used to run bash.

If so, the simplest fix is to drop the agent-server and the shared image entirely: keep the OpenHands prompt and action parsing, and run bash through the same FleetRepoEnv the r2egym scaffold already uses. Per-task R2E images, warm pools, and R2E grading then all work with no other changes.

Also, the new catch-all in trajectory_collect_engine turns any rollout error into a reward-0 trajectory that still goes into training. Please pull that out of this PR.

examples/deepswe/swe_env.py
susanbao: Have moved prompts and templates to the new single file.

tomergee:
Suggestion for landing this: make openhands a prompt and tool set on top of the environment that already works, instead of a parallel implementation. In `_initial_observation`, have the openhands branch do what the r2egym branch does: acquire the handle, then `make_fleet_repo_env(self.handle, command_files=SWEAGENT_COMMAND_FILES)`, and drop the workspace and `_setup_openhands_workspace`. In `_step_impl` remove the openhands-only block and let it fall through to `self.env.step`; the R2E env already ends the episode on `submit`. Go back to `self.final_reward_fn = self.env.compute_reward` and delete the custom `compute_reward`. In `swe_agent.py` keep `OPENHANDS_SYSTEM_PROMPT` but list the tools that are installed (`str_replace_editor`, `execute_bash`, `submit`). In `_init_global_fleet` remove the agent-server TemplateSpec, the `oh-img-` prefix and all `AGENT_SERVER_IMAGE` handling, so the fleet uses per-task R2E images with G replicas like it does for r2egym. In the yaml drop `AGENT_SERVER_IMAGE` and `SANDBOX_SESSION_KEY` and don't pass `--max_warmpool_size 1`. Finally, move the try/except in `trajectory_collect_engine` and the microbatch skip in the learner to a separate PR. That fixes reward, removes the per-episode clone, gives the model a file editor, and keeps the warm pools working as measured, with the openhands scaffold down to a prompt and a tool list. Running the real OpenHands agent-server in the pod is a separate follow-up that needs R2E images with the server baked in.

examples/deepswe/swe_env.py
tomergee: I see this was resolved, but the new head has the same shape: shared agent-server image, clone on every episode (the /testbed/.git check is never true on that image), and bash-only tools. f9785648 also removed the FleetRepoEnv binding again, so reward is back to the exit-code fallback. I posted a consolidated suggestion on the PR; can we settle that before closing this one? Thanks for pulling the trajectory-engine changes out.

tomergee:
There are two ways to run the OpenHands workspace on real task environments. Either way the shared agent-server image goes away: it has no repo, no tests, and no way to grade, so reward is zero and every episode clones from GitHub.

**Path A: bake the agent-server into the task images**
Build a derived image per R2E task image: start `FROM` the R2E image and copy in the agent-server binary (the OpenHands Dockerfile supports this directly with `BASE_IMAGE` and the `binary-minimal` target). Push them under a prefix and point `IMAGE_REWRITE_PREFIX` at it. The PR's per-task mode then works as written.
- fastest pod start, nothing happens at runtime
- self-contained images, same approach OpenHands uses for its own SWE-bench eval
- needs a build pipeline for thousands of images, plus registry storage
- every agent-server version bump means rebuilding the whole set
- not something to do inside this PR

**Path B: keep the R2E images, inject the server at pod start**
Add an init container to the pod template that copies the agent-server binary from the official image into an emptyDir volume. Mount that volume into the R2E container and point the keepalive command at the copied binary. Roughly:

```yaml
initContainers:
- name: oh-server
  image: ghcr.io/openhands/agent-server:1.44.1-python
  command: ["cp", "/usr/local/bin/openhands-agent-server", "/oh/"]
  volumeMounts: [{name: oh, mountPath: /oh}]
volumes:
- {name: oh, emptyDir: {}}
# task container: volumeMounts: [{name: oh, mountPath: /oh}]
# keepalive: ["/oh/openhands-agent-server", "--host", "0.0.0.0", "--port", "8000"]
```

This goes through `extra_pod_spec`, which the PR already uses, so it is about twenty lines in `get_openhands_pod_template` in `template.py`. No image builds.
- works with the existing R2E images, per-task pools of G replicas unchanged
- version bump is one tag in the template
- runs as root inside the R2E container, so `/testbed` is writable and the clone and symlink code can be deleted
- a few seconds per pod start for the copy, hidden by the warm pool
- one more image pulled per node, cached after the first pull
- the binary needs a glibc-compatible base; R2E images are Debian, should be fine, but it needs one smoke test

Either way, two things still need doing: bind `make_fleet_repo_env` alongside the workspace so reward comes from R2E grading (that was in an earlier push and got removed), and give the model a file editor in the prompt and tool set, not bash alone.

susanbao:
> Suggestion for landing this: make openhands a prompt and tool set on top of the environment that already works, instead of a parallel implementation. In `_initial_observation`, have the openhands branch do what the r2egym branch does: acquire the handle, then `make_fleet_repo_env(self.handle, command_files=SWEAGENT_COMMAND_FILES)`, and drop the workspace and `_setup_openhands_workspace`. In `_step_impl` remove the openhands-only block and let it fall through to `self.env.step`; the R2E env already ends the episode on `submit`. Go back to `self.final_reward_fn = self.env.compute_reward` and delete the custom `compute_reward`. In `swe_agent.py` keep `OPENHANDS_SYSTEM_PROMPT` but list the tools that are installed (`str_replace_editor`, `execute_bash`, `submit`). In `_init_global_fleet` remove the agent-server TemplateSpec, the `oh-img-` prefix and all `AGENT_SERVER_IMAGE` handling, so the fleet uses per-task R2E images with G replicas like it does for r2egym. In the yaml drop `AGENT_SERVER_IMAGE` and `SANDBOX_SESSION_KEY` and don't pass `--max_warmpool_size 1`. Finally, move the try/except in `trajectory_collect_engine` and the microbatch skip in the learner to a separate PR. That fixes reward, removes the per-episode clone, gives the model a file editor, and keeps the warm pools working as measured, with the openhands scaffold down to a prompt and a tool list. Running the real OpenHands agent-server in the pod is a separate follow-up that needs R2E images with the server baked in.

We completely agree with your suggestion. Layering OpenHands as a prompt and tool set on top of the existing R2E fleet environment is much simpler, avoids unnecessary pod overhead and runtime clones, and ensures grading is natively handled by R2E. We've implemented this in commit 62b2a278:

- R2E Fleet Integration: In _initial_observation, OpenHands now uses make_fleet_repo_env(self.handle, command_files=SWEAGENT_COMMAND_FILES) (same as sweagent). We completely dropped self.workspace and _setup_openhands_workspace (no per-episode git cloning or symlinking).
- Native Reward Grading: Reverted back to self.final_reward_fn = self.env.compute_reward and deleted the custom in-sandbox reward grading method.
- Execution & Termination: In _step_impl, actions now fall through directly to self.env.step, where the R2E gym environment executes tools and naturally terminates episodes on submit.
- Prompt & Tool Set: Updated OPENHANDS_SYSTEM_PROMPT to include str_replace_editor (view, create, str_replace, insert) alongside execute_bash and submit.
- Fleet & Warm-Pool Sizing: Removed AGENT_SERVER_IMAGE handling, is_shared_image sizing, and the oh-img- prefix in _init_global_fleet, _normalize_tasks_for_fleet, and PrewarmDatasetIterator. The fleet now uses standard per-task R2E images with $G$ replicas.
- Config & Launchers: Dropped AGENT_SERVER_IMAGE / SANDBOX_SESSION_KEY and aligned --max_warmpool_size with the generation replica count.

Please take another look when you get a chance!

susanbao:
> Suggestion for landing this: make openhands a prompt and tool set on top of the environment that already works, instead of a parallel implementation. In `_initial_observation`, have the openhands branch do what the r2egym branch does: acquire the handle, then `make_fleet_repo_env(self.handle, command_files=SWEAGENT_COMMAND_FILES)`, and drop the workspace and `_setup_openhands_workspace`. In `_step_impl` remove the openhands-only block and let it fall through to `self.env.step`; the R2E env already ends the episode on `submit`. Go back to `self.final_reward_fn = self.env.compute_reward` and delete the custom `compute_reward`. In `swe_agent.py` keep `OPENHANDS_SYSTEM_PROMPT` but list the tools that are installed (`str_replace_editor`, `execute_bash`, `submit`). In `_init_global_fleet` remove the agent-server TemplateSpec, the `oh-img-` prefix and all `AGENT_SERVER_IMAGE` handling, so the fleet uses per-task R2E images with G replicas like it does for r2egym. In the yaml drop `AGENT_SERVER_IMAGE` and `SANDBOX_SESSION_KEY` and don't pass `--max_warmpool_size 1`. Finally, move the try/except in `trajectory_collect_engine` and the microbatch skip in the learner to a separate PR. That fixes reward, removes the per-episode clone, gives the model a file editor, and keeps the warm pools working as measured, with the openhands scaffold down to a prompt and a tool list. Running the real OpenHands agent-server in the pod is a separate follow-up that needs R2E images with the server baked in.

While that approach simplifies the PR and fixes grading, it means we are not actually using OpenHands as the sandbox at all—the pods run standard R2E containers with zero OpenHands server components, no agent-server binary, and no OpenHands workspace client. I do not think this is OK.

susanbao:
> please make sure only changes relevant to the environment is included in the PR

@tianshub Thank you for your review. I have solved all your comments, please review again.

tomergee:
The current head is byte-identical to yesterday's revert, so same hitches on performance:
Shared agent-server image, clone on every episode, bash-only tools, exit-code reward. 

If the shared image reward is structurally zero, so every episode runs the full 30 turns and step time just goes to the ceiling (Getting a real reward requires the R2E task image under the agent-server)

The one thing worth trying in the meantime is raising max_concurrency from 64, since 256 rollouts per step at 64 concurrent is four serial waves through a sampler that is probably not saturated. Everything else, the oversized shared pool, the clone, the 60-second command timeout, is secondary vs the reward....

examples/deepswe/swe_env.py
```diff
@@ -341,7 +432,25 @@ def _teardown_global_fleet() -> None:
   r2egym = cast(Any, None)
   EnvArgs = cast(Any, None)
   RepoEnv = cast(Any, None)
-  Action = cast(Any, None)
+  Action = None
+
+
+class _ActionFallback:
```
tianshub: why do we need this? shouldn't it just fail if r2e-gym is not available?

examples/deepswe/swe_env.py
```diff
@@ -79,26 +86,57 @@ def _patched_start_container(
...
+
+
+def _normalize_tasks_for_fleet(
+    tasks: Any, scaffold: str = "r2egym"
+) -> list[Any]:
   """Normalize heterogeneous dataset entries into Task objects for SandboxFleet."""
+  TaskCls = None
   try:
     from agent_sandbox_rl import Task  # pytype: disable=import-error
+    if isinstance(Task, type):
```
tianshub: do we need these kind of fallbacks? it looks like we are spending a lot of energy handling import errors, can we not do that since the recipe should be runnable with all deps installed

examples/deepswe/swe_env.py
```diff
@@ -201,12 +266,18 @@ def __init__(
       num_generations: int = 8,
       batch_size: int = 8,
       max_warmpool_replicas: int | None = None,
+      scaffold: str = "r2egym",
+      image_rewrite: Any | None = None,
```
tianshub, line 230: what is this image_rewrite function?

examples/deepswe/swe_env.py
```diff
@@ -486,28 +650,285 @@ def _initial_observation(self) -> Any:
         )
         if self.scaffold == "r2egym":
           self.env.add_commands(R2EGYM_COMMAND_FILES)
-        else:
+        elif self.scaffold == "sweagent":
           self.env.add_commands(SWEAGENT_COMMAND_FILES)
     else:
-      self.env.reset()
+      if self.env is not None:
```
tianshub: the defensive if statement shouldn't be needed.

examples/deepswe/swe_env.py
```diff
@@ -442,34 +554,86 @@ def __init__(
...
+            self.entry.get(
+                "instance_id", self.entry.get("docker_image", "default")
+            )
         )
-        self.handle = fleet.acquire(task)
-        # TODO(wuhao): Revisit command_files once other harnesses (such as OpenHands) are supported.
-        cmd_files = r2egym_command_files()
-        self.env = make_fleet_repo_env(self.handle, command_files=cmd_files)
+        task = None
+        if hasattr(fleet, "tasks") and fleet.tasks:
```
tianshub: is tasks an optional field? how can we avoid such if statements?

examples/deepswe/swe_env.py
```diff
@@ -486,28 +650,285 @@ def _initial_observation(self) -> Any:
...
+      return str(
+          self.entry.get("problem_statement")
+          or self.entry.get("instruction")
+          or ""
+      )
+
     # Polls docker runtime to get task instruction.
     return self.env.get_task_instruction()  # pytype: disable=attribute-error
 
+  def _setup_openhands_workspace(self) -> None:
```
tianshub: can this be moved to an openhands specific file?

examples/deepswe/swe_env.py
```diff
@@ -2,10 +2,17 @@
 import json
```
tianshub, line 2: if openhand env is significantly different form swe_env (looks like it), we can consider just have a separate env file and not change the original swe_env.py

tianshub (changes requested):

tests/examples/template_test.py
```diff
@@ -0,0 +1,137 @@
...
+  """Verify get_template delegates to openhands or returns None for other scaffolds."""
+  mock_as = _setup_mock_agent_sandbox()
+  with mock.patch.dict(sys.modules, {"agent_sandbox_rl": mock_as}), \
+       mock.patch.dict(os.environ, {}, clear=True):
+    assert template.get_template("openhands") is not None
+    assert template.get_template("r2egym") is None
+    assert template.get_template("sweagent") is None
+
+
+def test_swe_agent_reexports():
```
tianshub: this test looks pretty redundant

examples/deepswe/swe_env.py
susanbao: You are completely right. _ActionFallback was originally added so that OpenHands rollouts and isolated tests could parse actions even when r2egym was not installed in the environment. However, for DeepSWE recipes, r2egym is a standard dependency (and required for reward grading). Having mock fallbacks adds dead weight, masks missing dependencies, and complicates the codebase.

I have removed _ActionFallback and let execution fail early if r2egym is not installed.

examples/deepswe/swe_env.py
susanbao: Agreed. These try/except blocks and the in-tree TaskCls fallback dataclass were added so unit tests could execute without agent_sandbox_rl installed, but it adds unnecessary defensive code to the recipe. Since the recipe requires all dependencies to be installed in the runtime environment, we should not spend energy catching import errors.

I have removed the fallback dataclass and import Task directly from agent_sandbox_rl (and mock it cleanly in test files where needed).

examples/deepswe/swe_env.py
susanbao: image_rewrite is a callback hook accepted by SandboxFleet.load_tasks in agent_sandbox_rl.

When running large distributed rollouts on GCP/GKE clusters, pulling hundreds of task images from public Docker Hub causes Docker Hub rate-limiting (429 Too Many Requests) or slow inter-region pulls. image_rewrite (or the IMAGE_REWRITE_PREFIX env var) rewrites the image path (e.g. from docker.io/swebench/... to an internal Artifact Registry / GCR mirror such as us-central1-docker.pkg.dev/<project>/...) so cluster nodes pull from the fast internal mirror without requiring edits to the dataset records.

examples/deepswe/swe_env.py
susanbao: Agreed. This check was leftover from an earlier iteration where OpenHands mode only initialized self.workspace and left self.env = None (which threw AttributeError on reset). Under the current design where the environment handle is always bound for evaluation, self.env is guaranteed to exist.

I have removed the defensive if check and call self.env.reset() directly.

examples/deepswe/swe_env.py
susanbao: In SandboxFleet, tasks is populated when fleet.load_tasks(...) is called. This check was defensively written.

examples/deepswe/swe_env.py
susanbao: I considered having a separate file, but openhands and r2egym share most of the heavy lifting in this file:

- Shared Fleet & Pipeline: Both scaffolds rely on the exact same global fleet management (_init_global_fleet, _teardown_global_fleet), pod acquisition/retries (fleet.acquire(task)), and PrewarmDatasetIterator (which manages sliding-window pre-warming of task images across K8s nodes during training).
- Shared Images & Grading: With the initContainer injection approach, both scaffolds now run on the exact same per-task R2E container images, and both use FleetRepoEnv.compute_reward for ground-truth grading.

If we split into a separate openhands_env.py, we would either:

- Duplicate hundreds of lines of fleet orchestration and dataset iteration, leading to code drift; or
- Need a much larger refactoring PR to extract shared logic into a separate fleet_utils.py / swe_utils.py.

tests/examples/template_test.py
susanbao: Have removed this test.

examples/deepswe/swe_env.py
susanbao: Have moved to the openhands specific file.

susanbao:
> The current head is byte-identical to yesterday's revert, so same hitches on performance: Shared agent-server image, clone on every episode, bash-only tools, exit-code reward.
> 
> If the shared image reward is structurally zero, so every episode runs the full 30 turns and step time just goes to the ceiling (Getting a real reward requires the R2E task image under the agent-server)
> 
> The one thing worth trying in the meantime is raising max_concurrency from 64, since 256 rollouts per step at 64 concurrent is four serial waves through a sampler that is probably not saturated. Everything else, the oversized shared pool, the clone, the 60-second command timeout, is secondary vs the reward....

@tomergee Thank you for your suggestion. I have solved the reward issue by directly use "self.final_reward_fn = self.env.compute_reward". Please review again.

examples/deepswe/swe_env.py
```diff
@@ -3,9 +3,18 @@
 import logging
 import os
 import threading
+import time
 from typing import Any, Optional, cast
 import numpy as np
 
+try:
```
tianshub: you don't need this try except block right? you should just use absolute import path

examples/deepswe/swe_env.py
```diff
@@ -79,16 +88,25 @@ def _patched_start_container(
     logging.debug("[SandboxFleet] r2egym in-memory patch note: %s", e)
 
 
-def _normalize_tasks_for_fleet(tasks: Any) -> list[Any]:
+def _get_image_rewrite_fn(image_rewrite: Any | None = None) -> Any | None:
```
tianshub: these can be moved to a separate file e.g. agent_sandbox_utils.py (since it's GKE only stuff, better to put it outside of core recipe logic). same for other relevant functions.

examples/deepswe/swe_env.py
```diff
@@ -201,15 +236,22 @@ def __init__(
...
+    self.scaffold = scaffold
     self.fleet = fleet or _get_global_fleet()
+    self.image_rewrite = _get_image_rewrite_fn(
+        image_rewrite or getattr(self.fleet, "_image_rewrite_fn", None)
+    )
     self.current_batch = None
     self.next_batch = None
-    self.prev_batch_images: list[str] = []
+    self.in_flight_batches: list[list[str]] = []
+    self.buffer_size = 2
```
tianshub: do you want to hardcode the buffer_size? (also buffer size isn't a very meaningful name, this is used for warm pool right? maybe pick a name more accurate)

examples/deepswe/swe_env.py
```diff
@@ -441,59 +493,120 @@ def __init__(
...
+      if os.getenv("ROUTER_URL"):
+        ws_kwargs["router_url"] = os.getenv("ROUTER_URL")
+      if os.getenv("ROUTER_AUTH_TOKEN"):
+        ws_kwargs["router_auth_token"] = os.getenv("ROUTER_AUTH_TOKEN")
+      ws_kwargs["working_dir"] = os.getenv(
+          "OPENHANDS_WORKING_DIR", "/testbed"
+      )
+      self.workspace = make_handle_workspace(self.handle, **ws_kwargs)
+
+    from agent_sandbox_rl.adapters.r2egym import (  # pytype: disable=import-error
```
tianshub: why do we want to import here instead of module scope?

tianshub (approved):
thanks for taking the feedback Sanbao.

A final check, have we validated the change works exactly as before for r2e_gym path (basically no regression on previous recipe config)?

examples/deepswe/swe_env.py
s-noghabi: The name `buffer_size` is vauge (buffer of what)? can you make it more specific, e.g., `self.in_flight_buffer_size`. 

also, let's not hardcode to 2. at least make it a constant defined.

examples/deepswe/swe_env.py
s-noghabi, line 1: have you tested the other two `r2egym` and `sweagent` also work after these changes?

s-noghabi (changes requested):

examples/deepswe/swe_env.py
susanbao: agent_sandbox_rl is an optional dependency specifically for GKE agent sandboxes (use_agent_sandbox=True) and isn't installed in the standard environment or CI. I have moved this line to the beginning of function.

examples/deepswe/swe_env.py
susanbao: Renamed buffer_size to self.max_in_flight_batches = 2 to accurately reflect that it bounds the number of in-flight batches retained in the warm pool. Keeping it hardcoded to 2 for now to match the lookahead pre-warming window.

examples/deepswe/swe_env.py
susanbao: Have done!

examples/deepswe/swe_env.py
susanbao: Thank you for your suggestion. Have done!

examples/deepswe/swe_env.py
susanbao: The code changes are strictly isolated (all OpenHands logic is guarded behind self.scaffold == "openhands" and self.workspace is not None, while r2egym and sweagent follow their original paths through self.env.step and make_fleet_repo_env).

I will kick off a quick smoke test with --scaffold=r2egym on the cluster to verify end-to-end execution after getting scripts and test results from Nitin, and will update this thread as soon as the run completes.

examples/deepswe/swe_env.py
susanbao: Thank you for your suggestion. Have used the constant _MAX_IN_FLIGHT_BATCHES = 2.

susanbao:
> thanks for taking the feedback Sanbao.
> 
> A final check, have we validated the change works exactly as before for r2e_gym path (basically no regression on previous recipe config)?

The code changes are strictly isolated (all OpenHands logic is guarded behind self.scaffold == "openhands" and self.workspace is not None, while r2egym and sweagent follow their original paths through self.env.step and make_fleet_repo_env).

I will kick off a quick smoke test with --scaffold=r2egym on the cluster to verify end-to-end execution after getting scripts and test results from Nitin, and will update this thread as soon as the run completes.

s-noghabi (approved):

PR #2086, opened 2026-09-02
Make the per-step gc.collect() in RLEngine configurable

gutianyu-google:
`RLEngine` runs a full host garbage collection twice on every training step, unconditionally: after the weight sync in `sync_weights` and after the actor log-prob pass in `_compute_actor_logps`. That is the right thing when models are offloaded to CPU (host references must be released promptly), but on a colocated setup with `offload_to_cpu=False` a full `gc.collect()` over a large Python heap is a pure stall inside the step.

**Measured**: on a GRPO post-training benchmark (Qwen3-0.6B, 2048 sequences/step, colocated trainer + vLLM rollout on TPU v7x, no CPU offload), `perf/train/weight_sync_time` rose from 4.7 s to 6.0 s (+1.2 s/step, consistent across steps) when the two collections were introduced; gating them recovered it, with rewards and completions unchanged.

This PR adds `ClusterConfig.per_step_gc_collect`:

- **Default `True` keeps current behavior exactly.**
- `False` skips the two per-step collections. The `offload_to_cpu`-guarded collections and the one-time collection at engine init are untouched.

**Reference**
The two collections were added in e4fc1dbf ("garbage collect after update_param, for both offloading case and weight sync").

**Checklist**

### DIFF
--- a/tunix/common/configs.py
+++ b/tunix/common/configs.py
@@ -390,6 +390,14 @@ class ClusterConfig:
       Alternatively, if a subclass of `BaseRollout` is provided, it will be used
       as the rollout engine.
     offload_to_cpu: Whether to offload models to CPU at each step..
+    gc_collect_after_weight_sync: Whether to run `gc.collect()` after each
+      weight sync. A weight sync creates another copy of the weights on HBM,
+      and that copy is only freed once Python's garbage collector releases
+      it; if the collector does not run, the copies accumulate and the run
+      can OOM over time. Collecting right after the sync frees the copy
+      promptly. A full collection costs about one second per step on a
+      colocated setup, so it can be disabled for performance, provided the
+      memory headroom is there.
     training_config: RL training config.
     rollout_config: Rollout config. It may be different for different modes,
       e.g. TRAIN vs EVAL.
@@ -407,6 +415,7 @@ class ClusterConfig:
   role_to_logical_axis_rule: dict[Role, flax.typing.LogicalRules] | None = None
   rollout_engine: str | type["base_rollout.BaseRollout"] = "vanilla"
   offload_to_cpu: bool = False
+  gc_collect_after_weight_sync: bool = True
 
   training_config: RLTrainingConfig
   rollout_config: dict[Mode, RolloutConfig] | RolloutConfig
--- a/tunix/rl/rl_cluster.py
+++ b/tunix/rl/rl_cluster.py
@@ -1156,8 +1156,11 @@ class RLEngine:
       actor_per_token_logps = jnp.concatenate(outs, axis=0)
       if not anchor_on_device:
         del anchor_policy_state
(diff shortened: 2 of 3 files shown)


tunix/rl/rl_cluster.py
```diff
@@ -1156,7 +1156,8 @@ def get_actor_per_token_logps(
       actor_per_token_logps = jnp.concatenate(outs, axis=0)
       if not anchor_on_device:
         del anchor_policy_state
-      gc.collect()
+      if self.cluster_config.per_step_gc_collect:
+        gc.collect()
```
wang2yn84: Let's move this gc collect into the if branch below, then you don't need the per_step_gc_collect guard here.

tunix/rl/rl_cluster.py
```diff
@@ -1179,7 +1180,8 @@ def sync_weights(self):
       )
       src_filtered_params = nnx.state(self.actor_trainer.model, filter_types)
       self.rollout.update_params(src_filtered_params, filter_types)
-      gc.collect()
+      if self.cluster_config.per_step_gc_collect:
```
wang2yn84: This gc collect is specifically for weight sync, consider rename to gc_collect_after_weight_sync?

tunix/rl/rl_cluster.py
gutianyu-google: Done in fcc35ad4, it sits inside the `offload_to_cpu` branch now, no knob.

tunix/rl/rl_cluster.py
gutianyu-google: Done, renamed to `gc_collect_after_weight_sync`.

wang2yn84:
Please squash commits to 1.

gutianyu-google:
Done, thanks!

tunix/common/configs.py
```diff
@@ -390,6 +390,11 @@ class ClusterConfig:
       Alternatively, if a subclass of `BaseRollout` is provided, it will be used
       as the rollout engine.
     offload_to_cpu: Whether to offload models to CPU at each step..
+    gc_collect_after_weight_sync: Whether to run `gc.collect()` after each
+      weight sync. This promptly releases the host-side references the sync
+      leaves behind, which matters when host memory is tight. On a colocated
+      setup a full host garbage collection on every step is a pure stall
+      (about one second per step observed), so it can be disabled.
```
wang2yn84: The statement is not that accurate. The reason why we have gc.collect after each weight sync is because weight sync creates another copy of weights on HBM. As time goes by, you might run into OOM issue if python gc never triggers. It's still ok if user wants to disable this for better performance but just needs to be aware of the potential issues. Let's make the description more accurate.

gutianyu-google: Done thanks!

PR #2088, opened 2026-09-02
Disable the profiler on an empty profiling window instead of raising

gutianyu-google:
`Profiler.__init__` raised `ValueError` when the first profile step was not before the last one. That window is derived from the trainer's `initial_step` and `max_step` plus the profiler options, and in RL setups the `PeftTrainer` is constructed with `initial_step == max_step`, so passing any `profiler_options` at all aborted the run at trainer construction, well after the models had loaded onto the accelerators.

An empty profiling window is a misconfiguration of profiling, not of the training run. This PR warns with the derived step numbers and disables profiling for that instance, matching how the profiler already behaves when no options are given or on a non-primary process. Valid windows are unaffected.

The existing `test_invalid_step_numbers` cases are kept and now assert the warning, the disabled state, and that `maybe_activate` does not start a trace.

**Checklist**

### DIFF
--- a/tests/sft/profiler_test.py
+++ b/tests/sft/profiler_test.py
@@ -289,20 +289,33 @@ class ProfilerTest(parameterized.TestCase):
           profiler_steps=0,
       ),
   )
-  def test_invalid_step_numbers(
-      self, max_step, initial_step, skip_first_n_steps, profiler_steps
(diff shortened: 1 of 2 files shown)


wang2yn84 (approved):
Thank you!

PR #2094, opened 2026-09-02
[Tunix] Support DeepSWE evaluation pipeline with MaxText and vLLM on TPU

susanbao:
Resolves #

### Description

This PR enhances the DeepSWE evaluation pipeline (`examples/deepswe/eval_deepswe.py`) to support both HuggingFace and MaxText models using vLLM on TPU clusters:

- **DeepSWE Evaluation Pipeline (`examples/deepswe/eval_deepswe.py`)**:
  - Adds CLI argument parsing (`--model_version`, `--model_source`, `--model_absolute_path`, `--mesh_fsdp`, `--mesh_tp`, `--scan_layers`, etc.) with fallback to environment variables.
  - Supports loading MaxText models (e.g. Qwen3.5-35B-A3B) directly into `VllmSampler` or via `AutoModel`.
  - Integrates 3D MRoPE position normalization patch for Qwen3-Omni-Moe thinker models in MaxText.
  - Dynamically configures Kubernetes `nodeSelector` via `patch_kubernetes_runtime` for GKE nodepools.
  - Implements Pass@k metric computation (`Pass@1`, `Pass@4`, `Pass@5`) using an unbiased combinatorial estimator across instance groups.
  - Adds direct GCS upload support for evaluation results using `google-cloud-storage` with `gcloud storage cp` fallback.
  - Adds garbage collection (`gc.collect()`) after model calls for memory management.
- **AutoModel (`tunix/models/automodel.py`)**:
  - Whitelists `checkpoint_storage_concurrent_gb` in `AutoModel.from_pretrained` kwargs for Orbax checkpoint restore configuration.
- **Trajectory Engine (`tunix/rl/agentic/trajectory/trajectory_collect_engine.py`)**:
  - Safely extracts prompt tokens from either `left_padded_prompt_tokens` or `padded_prompt_tokens` on `SamplerOutput` to prevent prompt desync.

**Reference**

**Colab Notebook**

**Checklist**

### DIFF
--- a/examples/deepswe/eval_deepswe.py
+++ b/examples/deepswe/eval_deepswe.py
@@ -1,81 +1,70 @@
 #!/usr/bin/env python
 """DeepSWE evaluation with deepscaler-style task-level parallelism.
 
-This script intentionally does not modify the existing eval entrypoint.
-It runs one full SWE trajectory per task and uses RolloutOrchestrator to
-parallelize whole tasks, similar to how deepscaler training/eval relies on
-the framework orchestrator rather than a custom outer runner.
-
-Environment Variables:
-  DATASET_NAME: The HuggingFace dataset name (default:
-  "R2E-Gym/SWE-Bench-Verified").
-  DATASET_SPLIT: The split of the dataset to evaluate (default: "test").
-  DATASET_CACHE: Path to cache the downloaded dataset (default:
-  "/scratch/dataset_cache").
-  MODEL_VERSION: The model identifier on HuggingFace to evaluate (default:
-  "Qwen/Qwen3-32B").
-  MAX_STEPS: Maximum number of agent steps per task/trajectory (default: 30).
-  MAX_MODEL_LEN: Maximum context length of the LLM (default: 32768).
-  MAX_RESPONSE_LENGTH / MAX_GENERATION_STEPS: Maximum tokens the model can
-  generate in a single response (default: 8192).
-  MAX_CONCURRENT: Maximum number of concurrent tasks/trajectories to run in
-  parallel (default: 256).
-  TIMEOUT: Timeout in seconds for a single trajectory evaluation (default: 600).
-  TASKS_LIMIT: Maximum number of tasks to evaluate. 0 means all tasks (default:
-  0).
-  MAX_CONTEXT_LIMIT: Limit on the number of context tokens before terminating
-  the episode (default: MAX_MODEL_LEN - 256).
-  ENABLE_GUARD: Set to "true" to enable action guard via GuardedSWEEnv (default:
-  "false").
-  ROLLOUT_ENGINE: The underlying LLM engine to use. Choices are 'vanilla',
-  'vllm', or 'sglang_jax' (default: "vllm").
-  VLLM_HBM_UTILIZATION: HBM utilization ratio for vLLM engine (default: 0.4).
-  VLLM_INIT_RANDOM_WEIGHTS: Set to "true" to init vLLM with random weights and
-  sync later (default: "true").
-  VLLM_SERVER_MODE: Set to "true" to run vLLM in server mode (default: "true").
-  VLLM_MAX_NUM_SEQS: Maximum number of concurrent sequences in vLLM (default:
-  128).
-  VLLM_MAX_BATCHED_TOKENS: Maximum batched tokens for vLLM (default: 165888).
-  SGLANG_MEM_FRACTION_STATIC: Static memory fraction for SGLang (default: 0.4).
-  SGLANG_INIT_RANDOM_WEIGHTS: Set to "true" to init SGLang with random weights
-  and sync later (default: "true").
-  SGLANG_MAX_RUNNING_REQUESTS: Maximum running requests for SGLang (default: 1).
-  OUTPUT_DIR: Directory where the evaluation results JSONL file will be saved
-  (default: "eval_results").
-  JAX_PLATFORMS: If set to "proxy", initializes Pathways utils.
-
-Usage:
-  # Full evaluation with default settings:
-  #   - Qwen/Qwen3-32B
-  #   - vLLM sampler
-  #   - MAX_CONCURRENT=256
-  #   - ENABLE_GUARD=false
-  #   - full evaluation split
-  python3 examples/deepswe/eval_deepswe.py
+This script runs SWE evaluation trajectories and uses RolloutOrchestrator to
+parallelize tasks across a TPU cluster. It supports both HuggingFace and MaxText
+models (e.g. Qwen3.5-35B-A3B) and configurable JAX/vLLM sharding meshes.
 """
 
+import argparse
 import asyncio
 import collections
+import gc
 import json
 import logging
+import math
 import os
 import sys
 import threading
 import time
 
+# Path Setup before JAX
+workdir = os.getcwd()
+tunix_root = os.path.join(workdir, "tunix")
+pathways_root = os.path.join(workdir, "pathways-utils")
+r2egym_root = os.path.join(workdir, "r2egym")
+
+for root in [
+    workdir,
+    pathways_root,
+    r2egym_root,
+    "/usr/github/rllm",
+    "/usr/github/pathways-utils",
+    "/app",
+]:
+  if os.path.exists(root) and root not in sys.path:
+    sys.path.insert(0, root)
+
+if "proxy" in os.getenv("JAX_PLATFORMS", ""):
+  try:
+    import pathwaysutils
+
+    pathwaysutils.initialize()
+    print("Pathways initialized successfully before JAX import.")
+  except Exception as e:
+    print(f"Failed to initialize pathwaysutils: {e}")
+
 from datasets import load_dataset
-from guarded_swe_env import GuardedSWEEnv
-from huggingface_hub import snapshot_download
 import jax
 import jax.numpy as jnp
 from jax.sharding import Mesh
 from kubernetes import client
 from kubernetes import config as k8s_config
 import numpy as np
-from swe_agent import SWEAgent
-from swe_env import SWEEnv
 from transformers import AutoTokenizer
+
+try:
+  from guarded_swe_env import GuardedSWEEnv
+  from swe_agent import SWEAgent
+  from swe_env import SWEEnv
+except ImportError:
+  from examples.deepswe.guarded_swe_env import GuardedSWEEnv  # pytype: disable=import-error
+  from examples.deepswe.swe_agent import SWEAgent  # pytype: disable=import-error
+  from examples.deepswe.swe_env import SWEEnv  # pytype: disable=import-error
+
+from huggingface_hub import snapshot_download
 from tunix.generate import tokenizer_adapter as tok_adapter
+from tunix.models.automodel import AutoModel, ModelSource
 from tunix.models.qwen3 import model as model_lib
 from tunix.models.qwen3 import params as params_lib
 from tunix.rl.agentic import utils as agentic_utils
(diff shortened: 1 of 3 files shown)


examples/deepswe/eval_deepswe.py
```diff
@@ -87,90 +76,336 @@
...
-    split=DATASET_SPLIT,
-    cache_dir=DATASET_CACHE,
-    num_proc=32,
-)
+try:
+  dataset = load_dataset(
+      DATASET_NAME,
+      split=DATASET_SPLIT,
+      cache_dir=DATASET_CACHE,
+      num_proc=32,
```
RissyRan: It seems like you were facing issue with this `num_proc=32`? shall we pass in a config and let users decide, instead of hardcode?

examples/deepswe/eval_deepswe.py
```diff
@@ -87,90 +76,336 @@
...
+
+  try:
+    import jax.numpy as jnp
+    from maxtext.layers.embeddings import Qwen3OmniMoeThinkerTextRotaryEmbedding
+
+    _orig_qwen_mrope_call = Qwen3OmniMoeThinkerTextRotaryEmbedding.__call__
+
+    def _patched_qwen_mrope_call(self, inputs, position, *args, **kwargs):
+      if position is not None:
+        # Case 1: vLLM (3, N, 1) -> (N, 1, 3)
```
RissyRan, line 419: Could you help point the source for cases to be handled for user readability?  i.e. why those cases exist

```
# Case 1: vLLM (3, N, 1) -> (N, 1, 3)
# Case 2: vLLM (3, N) -> (N, 1, 3)
# Case 3: (N,) -> (N, 1, 3)
# Case 4: (B, S) -> (B, S, 3)
# Case 5: (B, S, 1) -> (B, S, 3)
```

I didn't see the `else` branch, is this expected?

RissyRan: How do we test them?

examples/deepswe/eval_deepswe.py
```diff
@@ -187,142 +422,276 @@
...
+  if MODEL_SOURCE == "maxtext":
+    additional_config = {
+        "enable_continue_decode": True,
+        "maxtext_config": {
+            "model_name": MODEL_VERSION.lower().split("/")[-1],
+            "model_call_mode": "inference",
+            "load_parameters_path": MODEL_PATH,
+            "scan_layers": SCAN_LAYERS,
+            "enable_dp_attention": False,
+            "allow_split_physical_axes": True,
```
RissyRan: FYI: this usually gives huge perf hit if happens

For those hardcode configs, could we pass in from user cmd?

RissyRan (commented):
Thanks for the change! A few comments.

examples/deepswe/eval_deepswe.py
susanbao: Thanks for pointing this out! I will extract this logic into a dedicated helper function with clear docstrings explaining each case:

- Why Cases 1 & 2 exist (vLLM MRoPE format): For Qwen thinker/MRoPE models, vLLM provides input_positions with the 3 MRoPE coordinates (temporal, height, width) as the leading dimension: (3, num_tokens) or (3, num_tokens, 1). MaxText's inputs tensor has shape (num_tokens, 1, num_heads, head_dim). If not transposed, computing rotary embeddings yields (3, num_tokens, 1, head_dim), which causes JAX broadcasting failure (N, 1, 16, 64) * (3, N, 1, 64). Transposing to (N, 1, 3) aligns axis 0 to the token sequence dimension.
- Why Cases 3–5 exist (Standard 1D/2D text positions): When text positions come in as (N,), (B, S), or (B, S, 1), MaxText asserts position.shape[-1] == 3. These cases duplicate 1D text positions across the 3 MRoPE coordinate axes.
- Why there is no else branch: If position is already (..., 3) (e.g. (B, S, 3) or (N, 1, 3)), it is already in MaxText's native format and passes through to the original call untouched. I will add an explicit comment and fallback log to make this behavior obvious.

End-to-End Verification: In our cluster tests on TPU v5p-32 running Qwen3.5-35B-A3B and Qwen3-32B, this patch was exercised live across SWE-Bench tasks. Without it, vLLM crashes on the very first forward step with ValueError: Position IDs must be 2D (batch, seq) or 3D (batch, seq, 3) or broadcasting mismatch (8192, 1, 16, 64) vs (3, 8192, 1, 64).

examples/deepswe/eval_deepswe.py
susanbao: Yes, exactly. In containerized environments (e.g., GKE TPU pods with restricted /dev/shm or fewer CPU cores allocated to the driver container), load_dataset with num_proc=32 throws worker process/semaphore errors upon startup. The fallback was added to avoid crashing the job.

Thank you for the suggestion. I have added --dataset_num_proc to argparse.

examples/deepswe/eval_deepswe.py
susanbao: Thank you for the suggestion. I have added these as CLI arguments.

examples/deepswe/eval_deepswe.py
```diff
@@ -87,90 +76,410 @@
...
+  except ImportError as e:
+    logger.warning("Could not import maxtext_vllm_adapter: %s", e)
+
+  try:
+    import jax.numpy as jnp
+    from maxtext.layers.embeddings import Qwen3OmniMoeThinkerTextRotaryEmbedding
+
+    _orig_qwen_mrope_call = Qwen3OmniMoeThinkerTextRotaryEmbedding.__call__
+
+    def _patched_qwen_mrope_call(self, inputs, position, *args, **kwargs):
```
s-noghabi, line 417: can this be placed in a helper function, ideally not in this file? maybe maxtex_utils.py? it degrades the readability

examples/deepswe/eval_deepswe.py
```diff
@@ -187,142 +496,278 @@
...
-os.environ.setdefault("NODE_SELECTOR_VAL", "deepswe-cpu-pool")
+os.environ.setdefault("NODE_SELECTOR_VAL", NODE_SELECTOR_VAL)
 
-k8s_config.load_kube_config()
-k8s_client = client.CoreV1Api()
-k8s_client.list_namespace(timeout_seconds=5)
-logger.info("Kubernetes connection verified.")
 
-# ========================== Model ==========================
+def patch_kubernetes_runtime():
```
s-noghabi, line 502: is this patch_kubernetes_runtime() a change that would be needed both in eval and main deepswe? should we move this a to helper outside eval? maybe r2e_gym_helper.py?

examples/deepswe/eval_deepswe.py
```diff
@@ -187,142 +496,278 @@
...
-    "Using mesh shape fsdp=%d tp=%d (pure TP eval, total_devices=%d,"
-    " used_devices=%d)",
+    "Using mesh shape fsdp=%d tp=%d (total_devices=%d, used_devices=%d)",
     mesh.shape["fsdp"],
     mesh.shape["tp"],
     len(devices),
-    TP_SIZE,
-)
-
-if MODEL_VERSION == "Qwen/Qwen3-4B-Instruct-2507":
```
s-noghabi, line 228: why remove these checks?

examples/deepswe/eval_deepswe.py
```diff
@@ -187,142 +496,278 @@
...
-  model_config = model_lib.ModelConfig.qwen3_32b()
-else:
-  raise ValueError(f"Unsupported MODEL_VERSION: {MODEL_VERSION}")
-
-logger.info("Loading model weights from %s ...", MODEL_PATH)
-model = params_lib.create_model_from_safe_tensors(
-    MODEL_PATH, model_config, mesh, dtype=jnp.float32
+    total_mesh_devices,
 )
-sft_utils.show_hbm_usage()
```
s-noghabi, line 239: ditto, why remove these logging and info?

examples/deepswe/eval_deepswe.py
```diff
@@ -187,142 +496,278 @@
...
 
 # ========================== Sampler ==========================
 
 logger.info("Creating sampler with engine=%s ...", ROLLOUT_ENGINE)
 
-if ROLLOUT_ENGINE == "vanilla":
```
s-noghabi, line 245: why remove vanilla? we use this in our team

examples/deepswe/eval_deepswe.py
```diff
@@ -187,142 +496,278 @@
...
+            "weight_dtype": WEIGHT_DTYPE,
+            "prefuse_moe_weights": PREFUSE_MOE_WEIGHTS,
+            "attention": MAXTEXT_ATTENTION,
+            "checkpoint_storage_concurrent_gb": (
+                CHECKPOINT_STORAGE_CONCURRENT_GB
+            ),
+        },
+    }
+
+  mapping_config = mappings.MappingConfig()
```
s-noghabi, line 643: where are you building this? why drop the ```mapping_obj=None,
      model=model,
      backend="vllm_jax",
```

File: examples/deepswe/eval_deepswe.py [hunk 10]
```diff
@@ -456,14 +908,6 @@ class _EvalLoggingEnvMixin:
   """Adds phase-level reset/step logs for eval debugging."""
 
   def reset(self):
-    """Resets the environment and logs the timing.
```
s-noghabi, line 459: why remove?

examples/deepswe/eval_deepswe.py
```diff
@@ -478,7 +922,6 @@ def reset(self):
     return obs, info
 
   def step(self, action):
-    """Steps the environment and logs the action and timing."""
```
s-noghabi, line 481: ditto

examples/deepswe/eval_deepswe.py
```diff
@@ -531,14 +974,15 @@ def pairs_generator():
...
-  os.makedirs(OUTPUT_DIR, exist_ok=True)
+  if not OUTPUT_DIR.startswith("gs://"):
+    os.makedirs(OUTPUT_DIR, exist_ok=True)
 
   orchestrator = RolloutOrchestrator(
       engine_cls=EvalTrajectoryCollectEngine,
       engine_kwargs=dict(
           model_call=model_call,
           timeout=TIMEOUT,
-          max_context_limit=MAX_CONTEXT_LIMIT,
```
s-noghabi, line 541: removing this and replacing with  max_response_length is not the same thing. can you explain the rationale?

examples/deepswe/eval_deepswe.py
```diff
@@ -610,14 +1054,17 @@ async def run_evaluation():
...
-  """Computes and logs evaluation metrics such as Pass@1 and average reward.
+def _estimate_pass_at_k(n: int, c: int, k: int):
+  """Unbiased estimator for pass@k given n samples and c correct."""
+  if n < k:
+    return None
+  if n - c < k:
+    return 1.0
+  return 1.0 - math.comb(n - c, k) / math.comb(n, k)
 
-  Args:
```
s-noghabi, line 616: why remove these?

examples/deepswe/eval_deepswe.py
```diff
@@ -662,55 +1137,89 @@ def compute_pass_at_k(results):
...
-  in the file is a JSON object representing the evaluation outcome for a single
-  instance.
-
-  Args:
-    results: A list of dictionaries, where each dictionary contains the
-      evaluation results for a single instance, including 'pair_index',
-      'reward', 'num_steps', 'status', 'guard_blocked_steps', and
-      'guard_reasons'.
-
-  Returns:
```
s-noghabi, line 677: why remove these?

examples/deepswe/eval_deepswe.py
```diff
@@ -662,55 +1137,89 @@ def compute_pass_at_k(results):
...
+            "instance_id": entry.get("instance_id", r["instance_id"]),
+            "docker_image": entry.get("docker_image", ""),
+            "reward": r["reward"],
+            "num_steps": r["num_steps"],
+            "status": r["status"],
+            "guard_blocked_steps": r["guard_blocked_steps"],
+            "guard_reasons": r["guard_reasons"],
+        }
+        f.write(json.dumps(record) + "\n")
+    logger.info("Results saved to %s", output_file)
```
s-noghabi, line 1205: a lot of similar code copied in these if/else branching. can we simplify and reuse code with helper fns?

tunix/rl/agentic/trajectory/trajectory_collect_engine.py
```diff
@@ -543,13 +543,13 @@ def _safe_model_call():
...
     # Align trajectory prompt tokens with the rollout worker's actual
     # tokenization on the first turn to prevent prompt token desync.
-    if (
-        not self.agent.trajectory.steps
-        and rollout_output.left_padded_prompt_tokens is not None
-    ):
-      self.agent.trajectory.prompt_tokens = (  # pyrefly: ignore[missing-attribute]
-          rollout_output.left_padded_prompt_tokens[0]
-      )
+    prompt_tokens = getattr(
```
s-noghabi, line 546: why use getattr? won't rollout_output always have left_padded_prompt_tokens?

tunix/rl/agentic/trajectory/trajectory_collect_engine.py
s-noghabi, line 1: please make sure you add tests for all changes added,, especially to tunix/ core directories.

examples/deepswe/eval_deepswe.py
```diff
@@ -87,90 +76,410 @@
...
+      if position is not None:
+        # Case 1: vLLM (3, N, 1) -> (N, 1, 3)
+        if (
+            position.ndim == 3
+            and position.shape[0] == 3
+            and position.shape[-1] == 1
+        ):
+          position = jnp.transpose(position.squeeze(-1), (1, 0))[:, None, :]
+        # Case 2: vLLM (3, N) -> (N, 1, 3)
+        elif position.ndim == 2 and position.shape[0] == 3:
```
s-noghabi, line 427: The check elif position.ndim == 2 and position.shape[0] == 3: assumes that any 2D position tensor with leading dimension 3 is a vLLM 3D coordinate tuple (3, N). If a standard 2D batch of shape (batch_size, seq_len) is evaluated with batch_size = 3, this branch incorrectly triggers, no?

examples/deepswe/eval_deepswe.py
```diff
@@ -387,18 +838,19 @@ def model_call(chat_completions, env_unused):
...
     else:
       with sampler_lock:
         out = sampler(
             prompt,
-            max_generation_steps=MAX_RESPONSE_LENGTH,
+            max_generation_steps=max_gen_steps,
             echo=False,
             eos_tokens=qwen_eos_tokens,
         )
+    gc.collect()
```
s-noghabi, line 853: calling gc collect synchronously inside model_call on every single generation step seems like a poor performance choice

examples/deepswe/eval_deepswe.py
```diff
@@ -187,142 +496,278 @@
...
+      "disable_log_stats": False,
+      "tokenizer": tokenizer_path,
+  }
+  if MODEL_SOURCE == "maxtext":
+    engine_kwargs["hf_overrides"] = {"architectures": ["MaxTextForCausalLM"]}
   vllm_config = VllmConfig(
       mesh=mesh,
       hbm_utilization=VLLM_HBM_UTILIZATION,
-      init_with_random_weights=VLLM_INIT_RANDOM_WEIGHTS,
+      init_with_random_weights=False,
```
s-noghabi, line 659: why false and not VLLM_INIT_RANDOM_WEIGHTS

examples/deepswe/eval_deepswe.py
```diff
@@ -1,81 +1,70 @@
 #!/usr/bin/env python
 """DeepSWE evaluation with deepscaler-style task-level parallelism.
 
-This script intentionally does not modify the existing eval entrypoint.
```
tianshub, line 4: why are these doc string removed?

examples/deepswe/eval_deepswe.py
```diff
@@ -1,81 +1,70 @@
...
 import json
 import logging
+import math
 import os
 import sys
 import threading
 import time
 
+# Path Setup before JAX
+workdir = os.getcwd()
```
tianshub, line 22: why do we need this? if you have the venv with dependencies installed, all these should be available no?

examples/deepswe/eval_deepswe.py
```diff
@@ -1,81 +1,70 @@
...
 import numpy as np
-from swe_agent import SWEAgent
-from swe_env import SWEEnv
 from transformers import AutoTokenizer
+
+try:
+  from guarded_swe_env import GuardedSWEEnv
+  from swe_agent import SWEAgent
+  from swe_env import SWEEnv
+except ImportError:
```
tianshub, line 60: in what case the import won't be available?

examples/deepswe/eval_deepswe.py
```diff
@@ -87,90 +76,410 @@
...
-if os.getenv("JAX_PLATFORMS", None) == "proxy":
-  import pathwaysutils
-
-  pathwaysutils.initialize()
+logger.info("JAX backend initialized.")
 
 os.environ["TOKENIZERS_PARALLELISM"] = "true"
 
+# Register MaxText vLLM adapter if using a MaxText model
+if MODEL_SOURCE == "maxtext":
```
tianshub, line 402: please put this into a maxtext_model_helper file

examples/deepswe/eval_deepswe.py
```diff
@@ -87,90 +76,410 @@
...
 # ========================== Dataset ==========================
 
 logger.info("Loading dataset %s split=%s ...", DATASET_NAME, DATASET_SPLIT)
-dataset = load_dataset(
-    DATASET_NAME,
-    split=DATASET_SPLIT,
-    cache_dir=DATASET_CACHE,
-    num_proc=32,
-)
+if DATASET_NUM_PROC > 1:
```
tianshub, line 457: why do we need this knob? do you see data loading being slow?

examples/deepswe/eval_deepswe.py
```diff
@@ -87,90 +76,410 @@
...
+parser_cli.add_argument(
+    "--enable_continue_decode",
+    type=str2bool,
+    default=os.getenv("ENABLE_CONTINUE_DECODE", "true").lower() == "true",
+    help="Whether to enable continue decode in vLLM",
+)
+parser_cli.add_argument(
+    "--node_selector_val",
+    type=str,
+    default=os.getenv("NODE_SELECTOR_VAL", "cpu-np"),
```
tianshub, line 308: why do we want to have two ways to config the CLI, one with py args one with env var, and default one to another? can we just reconcile to one mechanism?

examples/deepswe/eval_deepswe.py
tianshub: +1. Nitin had similar stuff here https://github.com/google/tunix/pull/2051/changes#diff-1a6299c4c9d1a03ec4a737371fb4cd43e6280c5a797d324ab5ebeb9df92b97ea

can you reuse?

examples/deepswe/eval_deepswe.py
```diff
@@ -187,142 +496,278 @@
...
 
+# Mesh Setup
 devices = jax.devices()
-# Force pure tensor parallelism for eval: DP=1, TP=8.
-# Qwen3-32B has tensors such as (5120, 8, 128), so TP must not exceed 8 for
-# shardings that partition that dimension on the tp axis.
-TP_SIZE = 8
-mesh_devices = np.array(devices[:TP_SIZE]).reshape(1, TP_SIZE)
+total_mesh_devices = MESH_FSDP * MESH_TP
+if total_mesh_devices > len(devices):
```
tianshub, line 582: a lot of these functions (similar for tokenizer) can just go to a util file. We should keep the recipe script as lean as possible.

tianshub (changes requested):

examples/deepswe/eval_deepswe.py
niting: Yes, these changes should go into r2e_gym_helper.py if not already there.

PR #2116, opened 2026-09-04
[Tunix] Enforce strict formatting rewards and stop instruction in GSM8K demo

andytwigg:
### Summary
Improves trajectory completion and eliminates partial reward incentives for rambling generations past the numerical answer in the distributed GSM8K GRPO demo.

### Details
1. **Prompt Template Clarification**:
   Updated `GSM8K_PROMPT_TEMPLATE` to explicitly instruct the model:
   `"After the closing </answer> tag, immediately end your response and stop generating."`
2. **Strict Formatting Reward**:
   Updated `score_gsm8k_completion` to assign `0.0` (instead of `0.5`) when `not format_correct and answer_correct`. Without strict formatting enforcement, the model receives partial credit even when generating run-on tokens or failing to properly terminate at `</answer>`.
3. **Tests**:
   Updated `tests/experimental/examples/math_gsm8k_dist/gsm8k_test.py` to verify the stop instruction in `build_prompt` and assert the strict formatting reward behavior (`0.0` for malformed outputs).

### Verification
- Ran `python3 -m unittest tests/experimental/examples/math_gsm8k_dist/gsm8k_test.py`: all 9 unit tests passed.

### DIFF
--- a/tunix/experimental/examples/math_gsm8k_dist/gsm8k.py
+++ b/tunix/experimental/examples/math_gsm8k_dist/gsm8k.py
@@ -41,7 +41,8 @@ GSM8K_PROMPT_TEMPLATE = (
     "<reasoning>...</reasoning> tags.\n"
     "Then, put your final numerical answer inside "
     "<answer>\\boxed{{}}</answer> tags. Do not put anything else in the "
-    "answer tags.\n\n"
+    "answer tags.\n"
+    "After the closing </answer> tag, immediately end your response and stop generating.\n\n"
     "Problem: {}\n"
     "<reasoning>\n"
 )
@@ -175,8 +176,6 @@ def score_gsm8k_completion(
     reward = 1.0
   elif format_correct and not answer_correct:
     reward = 0.1
-  elif not format_correct and answer_correct:
-    reward = 0.5
   else:
     reward = 0.0
   return reward, {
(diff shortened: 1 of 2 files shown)


tunix/experimental/examples/math_gsm8k_dist/gsm8k.py
s-noghabi, line 1: Can you clarify more as to why we are changing this? this is diverging from the main agentic path in tunix today: [google3/third_party/py/tunix/oss/examples/math_gsm8k/qwen3_grpo_demo.py](https://github.com/google/tunix/blob/main/examples/math_gsm8k/qwen3_grpo_demo.py#L242) and it would make it harder for us to compare the two. 

Ideally we would keep these two exactly the same

s-noghabi (changes requested):

tunix/experimental/examples/math_gsm8k_dist/gsm8k.py
andytwigg: Sure - there are 2 reasons:
1) make the reward function match what we used for cloudnext (and IIRC what ds-r1 had), where we had known convergence results for this task. Do you have convergence results for the gsm8k task with the current reward function? If so, we can leave it.
2) update the template so the model is instructed to stop producing tokens after generating the answer. Otherwise it seems like it continues to produce output beyond that.

tianshub (approved):

PR #2083, opened 2026-09-02
[Trellis]Qwen3.5-35B; Weight Conversion and Raiden Weight Sync Integration

YixuanWang-99:
## Description

Integrates **Raiden weight synchronization** with **MaxText trainer** and **vLLM rollout workers** for distributed RL workloads (GRPO / Trellis). Refactors destination-side weight sync into a shared mixin, handles multi-host FFI synchronization, manages cache lifecycles, and optimizes host memory usage.

**Companion MaxText PR**: [AI-Hypercomputer/maxtext#5089](https://github.com/AI-Hypercomputer/maxtext/pull/5089)

---

## Key Changes
- **MaxText Trainer Integration (`run_trainer_node.py`, `maxtext_utils.py`)**: Adds `trainer_backend="maxtext"` support, configures mesh topologies (FSDP, TP, EP), enables weight conversion, and aligns trainer TP with rollout TP.
- **`RaidenDestinationWeightSyncMixin` Refactoring (`raiden_weight_sync_mixin.py`)**: Extracts a reusable mixin consolidating `get_weight_sync_metadata`, `bind_weight_sync`, `sync_weights`, and `release_weight_sync` across sampler adapters (`inprocess_vllm_sampler_adapter`, `vanilla_sampler_adapter`).
- **Multi-Host FFI & Step-0 Sync (`raiden_handler.py`, `raiden_synchronizer.py`, `rl_program.py`)**: Supports comma-separated multi-host `control_plane_rpc_address` registrations, forces weight sync at step 0 to prevent uninitialized generation, and adds tensor/element count verification to checksum diagnostics.
- **Cache Lifecycle & Memory Optimization (`raiden_weight_sync_delegate.py`)**: Adds prefix cache reset and KV cache reinitialization; adds `--enable_prefix_caching` flag (default `false`) to prevent recurrent state desync in hybrid models; releases host arrays post-sync.
- **128-Lane MoE Chunking & Launch Script (`tunix.generate.utils`, `launch_raiden.sh`)**: Aligns MoE interleaving to 128-element lane chunks for TPU GMM v2 kernels, and provides unified cluster deployment and triage automation.

---

## Testing
- **Unit Tests**: Updated and verified `raiden_handler_test.py`, `raiden_synchronizer_test.py`, and `weight_sync_test.py`.
[...]

PR #2146, opened 2026-09-09
[Raiden Weight Sync 1/6] Support multi-host discovery and multi-axis sharding in RaidenHandler

YixuanWang-99:
## Overview

**Stack:**
- **[T1] google/tunix (this PR)**: Multi-host discovery & multi-axis sharding in RaidenHandler
- [T2a] google/tunix: Transport selection, FFI preflight mismatch check, and host-stage normalization
- [T3] google/tunix: Extract RaidenDestinationWeightSyncMixin
- [T4] google/tunix: Cache lifecycle and prefix caching
- [T5] google/tunix: MoE 128-lane interleaving for TPU GMM layout (pairs with MaxText M2)

## Details
- Supports comma-separated `control_plane_rpc_address` for multi-host setups.
- Validates multi-axis mesh sharding specifications.
- **Review thread addressed**: Restores the `if name_resolver is not None:` guard before instantiating `WeightSyncWorkerRpcClient` (addressing @tianshub's comment).

## Verification
`pytest tests/experimental/weight_sync/raiden_handler_test.py` (49/49 passed).

### DIFF
--- a/tunix/experimental/weight_sync/raiden_handler.py
+++ b/tunix/experimental/weight_sync/raiden_handler.py
@@ -35,8 +35,9 @@ from __future__ import annotations
 import asyncio
 import dataclasses
 import logging
+import math
 import threading
-from typing import Any, Mapping, Optional, Sequence
+from typing import Any, Optional, Sequence
 
 from tpu_sync.rpc import controller_service_pb2
 from tpu_sync.rpc import raiden_controller
@@ -52,7 +53,7 @@ class RaidenTransferOptions:
   parallelism: Optional[int] = None
   expected_block_count: Optional[int] = None
   skip_d2h: bool = False
-  skip_tiling: Optional[Mapping[int, bool]] = None
+  skip_tiling: Optional[dict[int, bool]] = None
   group_size: int = 1
 
   def __post_init__(self) -> None:
@@ -217,7 +218,7 @@ class _RaidenTransport:
     self._controller.register_work_unit(
         unit=self._to_raiden_id(metadata.unit),
         shards=list(metadata.shards),
-        control_plane_rpc_address=metadata.control_plane_rpc_address,
+        control_plane_rpc_address=metadata.control_plane_rpc_address or None,
         mesh_shape=metadata.mesh_shape,
         layout=metadata.layout,
         global_shape=metadata.global_shape,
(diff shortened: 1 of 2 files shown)


tunix/experimental/weight_sync/raiden_handler.py
```diff
@@ -214,10 +215,12 @@ def register_work_unit(self, metadata: weight_sync.WorkUnitMetadata) -> None:
           " so its assigned ports are known"
       )
     self._validate_metadata(metadata)
+    control_addr = metadata.control_plane_rpc_address
+    first_addr = control_addr.split(",")[0].strip() if control_addr else ""
```
tianshub: in what situation there will be multiple ip address?

tunix/experimental/weight_sync/raiden_handler.py
```diff
@@ -229,6 +232,14 @@ def register_work_unit(self, metadata: weight_sync.WorkUnitMetadata) -> None:
             else None
         ),
     )
+    if control_addr and "," in control_addr:
```
wang2yn84: With the latest tpu-sync, you should not need this. cc @datenglin

tunix/experimental/weight_sync/raiden_handler.py
```diff
@@ -273,16 +284,23 @@ def _validate_metadata(metadata: weight_sync.WorkUnitMetadata) -> None:
                 f" of {tensor.name!r} must have logical mesh size 1, got"
                 f" {logical_size}"
             )
-        elif axis not in physical_axes:
+          continue
+        # A dimension may be sharded over the product of several axes, which
+        # the wire form spells comma-joined; the controller splits it the same
+        # way to fold the sub-axis coordinates into one tensor coordinate.
+        sub_axes = axis.split(",")
```
wang2yn84, line 281: Can you add a test?

tunix/experimental/weight_sync/raiden_handler.py
```diff
@@ -214,10 +215,12 @@ def register_work_unit(self, metadata: weight_sync.WorkUnitMetadata) -> None:
           " so its assigned ports are known"
       )
     self._validate_metadata(metadata)
+    control_addr = metadata.control_plane_rpc_address
+    first_addr = control_addr.split(",")[0].strip() if control_addr else ""
     self._controller.register_work_unit(
         unit=self._to_raiden_id(metadata.unit),
         shards=list(metadata.shards),
-        control_plane_rpc_address=metadata.control_plane_rpc_address,
+        control_plane_rpc_address=first_addr or None,
```
datenglin: This is wrong? Doing this will register only the first host.

tunix/experimental/weight_sync/raiden_handler.py
datenglin: That's right. This also does not seem to be correct. I do not think Raiden will do transfer for `sub_id`.

tunix/experimental/weight_sync/raiden_handler.py
YixuanWang-99: Added unit tests

tunix/experimental/weight_sync/raiden_handler.py
YixuanWang-99: In my e2e test, the rollout acts with multiple replica, each rollout replica runs as an independent worker process with a single IP and registers its own separate work unit (job_name=f"replica_{server_id}"), so multiple comma-separated IPs are not used. 

I've removed the comma-splitting logic and reverted to forwarding metadata.control_plane_rpc_address or None directly.

tunix/experimental/weight_sync/raiden_handler.py
YixuanWang-99: Thanks for pointing this out! When I tested e2e with multiple rollout replicas, each replica already registers its own work unit independently with its own single IP, so this logic is unnecessary and not used anyway. I have removed the comma-splitting and sub_id registration block and restored direct registration.

tianshub (approved):

tunix/experimental/weight_sync/raiden_handler.py
```diff
@@ -217,7 +218,7 @@ def register_work_unit(self, metadata: weight_sync.WorkUnitMetadata) -> None:
     self._controller.register_work_unit(
         unit=self._to_raiden_id(metadata.unit),
         shards=list(metadata.shards),
-        control_plane_rpc_address=metadata.control_plane_rpc_address,
+        control_plane_rpc_address=metadata.control_plane_rpc_address or None,
```
wang2yn84: Why do we need "or None"?

PR #2148, opened 2026-09-09
[Raiden Weight Sync 3/6] Extract RaidenDestinationWeightSyncMixin for rollout samplers

YixuanWang-99:
## Overview
Part of the stacked Raiden weight-sync enablement PR chain replacing #2083.

**Stack:**
- [T1] https://github.com/google/tunix/pull/2146: Multi-host discovery & multi-axis sharding in RaidenHandler
- [T2a] https://github.com/google/tunix/pull/2147: Transport selection, FFI preflight mismatch check, and host-stage normalization
- **[T3] google/tunix (this PR)**: Extract RaidenDestinationWeightSyncMixin
- [T4] google/tunix: Cache lifecycle and prefix caching
- [T5] google/tunix: MoE 128-lane interleaving for TPU GMM layout (pairs with MaxText M2)
- [T6] google/tunix: MaxText trainer configuration plumbing

## Details
- Net −65 lines pure refactoring.
- Extracts `RaidenDestinationWeightSyncMixin` in `tunix/experimental/rollout/raiden_weight_sync_mixin.py` to eliminate copy-pasted Raiden weight sync hooks (`bind_weight_sync`, `get_weight_sync_metadata`, `pre_weight_sync`, `weight_sync`, `post_weight_sync`, `abort_weight_sync`, `get_weight_sync_status`).
- Adopted by `inprocess_vllm_sampler_adapter`, `vanilla_sampler_adapter`, and external vLLM adapter.

## Verification
- `pytest tests/experimental/weight_sync/weight_sync_test.py` (14/14 passed)

PR #2147, opened 2026-09-09
[Raiden Weight Sync 2/6] FFI transport selection, preflight mismatch check, and host-stage normalization

YixuanWang-99:
## Overview
Part of the stacked Raiden weight-sync enablement PR chain replacing #2083.
**Pairs with MaxText [M4]** (`AI-Hypercomputer/maxtext`).

**Stack:**
- [T1] https://github.com/google/tunix/pull/2146: Multi-host discovery & multi-axis sharding in RaidenHandler
- **[T2a] google/tunix (this PR)**: Transport selection, FFI preflight mismatch check, and host-stage normalization
- [T3] google/tunix: Extract RaidenDestinationWeightSyncMixin
- [T4] google/tunix: Cache lifecycle and prefix caching
- [T5] google/tunix: MoE 128-lane interleaving for TPU GMM layout (pairs with MaxText M2)
- [T6] google/tunix: MaxText trainer configuration plumbing

## Details
- **The Step 0 Fix**: Solves the end-to-end FFI path failure by centralizing transport resolution.
- Adds and exports `resolve_use_ffi(explicit: bool | str | None, *, is_proxy: bool | None) -> bool` and `normalize_host_stage(host_stage: bool | None, *, use_ffi: bool, is_proxy: bool) -> bool`.
- `WeightSyncCoordinator` adds preflight validation rejecting any pairing of mismatched source and destination transport modes before scheduling any transfers.
- `WorkUnitMetadata` carries resolved `transport_mode` ("ffi" / "tcp") and `use_ffi`.
- Integrates `release_host_arrays()` and checksum / tensor element count diagnostics.

## Verification
- `pytest tests/experimental/weight_sync/raiden_synchronizer_test.py` (32/32 passed)
- `pytest tests/experimental/weight_sync/weight_sync_coordinator_test.py` (73/73 passed)

PR #2149, opened 2026-09-09
[Raiden Weight Sync 4/6] Manage cache lifecycle and prefix caching in RaidenWeightSyncDelegate

YixuanWang-99:
## Overview

**Stack:**
- [T1] https://github.com/google/tunix/pull/2146: Multi-host discovery & multi-axis sharding in RaidenHandler
- [T2a] https://github.com/google/tunix/pull/2147: Transport selection, FFI preflight mismatch check, and host-stage normalization
- [T3] https://github.com/google/tunix/pull/2148: Extract RaidenDestinationWeightSyncMixin
- **[T4] google/tunix (this PR)**: Cache lifecycle and prefix caching
- [T5] google/tunix: MoE 128-lane interleaving for TPU GMM layout (pairs with MaxText M2)

## Details
- Resets prefix cache and clears stale KV entries upon weight transfer.
- Adds `--enable_prefix_caching` flag (default `false`) to avoid recurrent state desync in hybrid models (such as Qwen3.5 GDN layers).

### DIFF
--- a/tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
+++ b/tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
@@ -17,10 +17,12 @@
 from __future__ import annotations
 
 import os
-from typing import Any, List
+from typing import Any, List, Mapping
 
 from absl import logging
+import jax
 from tunix.experimental.weight_sync import raiden_synchronizer
+from tunix.experimental.weight_sync import weight_sync_coordinator
 
 
 class RaidenWeightSyncDelegate:
@@ -36,14 +38,24 @@ class RaidenWeightSyncDelegate:
   abort after a partial weight_sync cannot restore the previous weights.
   """
 
-  def __init__(self, *args, worker_index: int = 0, **kwargs):
-    super().__init__(*args, **kwargs)
+  def __init__(
+      self,
+      *args,
+      worker_index: int = 0,
+      **kwargs,
+  ):
+    del args, kwargs
+    # TODO(tunix-dev): add a lock when enabling multiple samplers in one worker.
+    self._sampler = None
     self._synchronizers: List[Any] = [
         raiden_synchronizer.RaidenSynchronizer(
-            "rollout", worker_index=worker_index, auto_h2d=True
+            "rollout",
+            worker_index=worker_index,
+            auto_h2d=True,
         )
     ]
     self._version = 0
+    self._tracker = weight_sync_coordinator.WorkerRoundTracker()
 
   def is_bounded(
       self,
@@ -52,10 +64,12 @@ class RaidenWeightSyncDelegate:
     return all(s.bound for s in self._synchronizers)
 
   async def bind_weight_sync(
-      self, sync_request: Any = None, state: Any = None, **kwargs
+      self, sync_request: Any = None, state: Any = None, sampler: Any = None, **kwargs
   ) -> Any:
     """Binds destination-side transport resources for weight sync."""
     del sync_request, kwargs
+    if sampler is not None:
+      self._sampler = sampler
 
     for sync in self._synchronizers:
       # The state arrays never change, so one bind covers every round.
@@ -69,14 +83,33 @@ class RaidenWeightSyncDelegate:
     del kwargs
     return [s.work_unit_metadata() for s in self._synchronizers]
 
+  def _has_round(self, sync_request: Any) -> bool:
+    extra = getattr(sync_request, "extra_config", None) or {}
+    return extra.get("req_id") is not None
+
   async def pre_weight_sync(self, sync_request: Any = None, **kwargs) -> Any:
     """Pre-sync phase hook executed before weight transfer begins."""
-    del sync_request, kwargs
+    del kwargs
+    if self._has_round(sync_request):
+      if not self._tracker.admit(sync_request, "prepared"):
+        return True
+      self._tracker.complete(sync_request, "prepared")
+
+    if self._sampler is None:
+      raise RuntimeError("Sampler is not available for weight sync")
+
+    self._sampler.delete_cache()
+    jax.effects_barrier()
+
     return True
 
   async def weight_sync(self, sync_request: Any = None, **kwargs) -> Any:
     """Executes weight installation on device from host staging buffer."""
     del kwargs
+    if self._has_round(sync_request):
+      if not self._tracker.admit(sync_request, "h2d_done"):
+        return self._version
+
     for sync in self._synchronizers:
       if not sync.bound:
         raise RuntimeError("bind_weight_sync must run before weight_sync")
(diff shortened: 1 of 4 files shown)


tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
```diff
@@ -36,14 +40,33 @@ class RaidenWeightSyncDelegate:
   abort after a partial weight_sync cannot restore the previous weights.
   """
 
-  def __init__(self, *args, worker_index: int = 0, **kwargs):
-    super().__init__(*args, **kwargs)
+  def __init__(
+      self,
+      *args,
+      job_name: Optional[str] = None,
```
tianshub: you don't need both job_name, server_id and worker_index as they serve duplicated purpose. Just rely on worker_index and compose the job name same as other PR

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
```diff
@@ -36,14 +40,33 @@ class RaidenWeightSyncDelegate:
...
+      **kwargs,
+  ):
+    del args, kwargs
+    # Raiden partitions the weights across every unit sharing a job_name, so
+    # replicas that all call themselves "rollout" get a slice each instead of a
+    # copy each. server_id is already unique per replica and shared across the
+    # hosts within one, which is exactly the grouping job_name needs.
+    self.server_id = server_id
+    self.job_name = job_name or server_id or "rollout"
+    self._sampler = sampler
```
tianshub: I don't see how this sampler is used?

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
```diff
@@ -36,14 +40,33 @@ class RaidenWeightSyncDelegate:
...
     self._synchronizers: List[Any] = [
         raiden_synchronizer.RaidenSynchronizer(
-            "rollout", worker_index=worker_index, auto_h2d=True
+            self.job_name,
+            worker_index=worker_index,
+            auto_h2d=True,
         )
     ]
     self._version = 0
+    self._sync_lock = asyncio.Lock()
```
tianshub: do you expect multiple samplers in one worker? if not I don't think we need this lock

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
wang2yn84: It's used to invoke reset_prefix_cache and delete kv cache.

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
```diff
@@ -36,14 +40,33 @@ class RaidenWeightSyncDelegate:
...
+      sampler: Optional[Any] = None,
+      **kwargs,
+  ):
+    del args, kwargs
+    # Raiden partitions the weights across every unit sharing a job_name, so
+    # replicas that all call themselves "rollout" get a slice each instead of a
+    # copy each. server_id is already unique per replica and shared across the
+    # hosts within one, which is exactly the grouping job_name needs.
+    self.server_id = server_id
+    self.job_name = job_name or server_id or "rollout"
```
wang2yn84: Server id is much worse than "rollout". How can user infer if certain server id is rollout or trainer?

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
wang2yn84: The only reason why we need worker_index here is because there will be multiple rollout servers. That's the same reason why we don't pass job_name and instead using hardcoded "rollout" for simplicity. Please revert it back if it's not necessary.

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
```diff
@@ -69,30 +94,101 @@ async def get_weight_sync_metadata(self, **kwargs) -> Any:
...
+      if sampler is not None:
+        model_runner = getattr(sampler, "_model_runner", None)
+        if model_runner is not None and hasattr(model_runner, "state_leaves"):
+          state = getattr(model_runner, "state", None)
+          if state is not None:
+            model_runner.state_leaves = tuple(
+                jax.tree_util.tree_leaves(state)
+            )
+
+        llm = getattr(sampler, "llm", None)
```
wang2yn84: These are the impl details of sampler. Consider move these out of update_param() to a new API delete_cache() in sampler so that you only need to call self.sampler.delete_cache()?

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
```diff
@@ -69,30 +94,101 @@ async def get_weight_sync_metadata(self, **kwargs) -> Any:
...
-    del sync_request, kwargs
-    return True
+    del kwargs
+    async with self._sync_lock:
+      if self._has_round(sync_request):
+        if not self._tracker.admit(sync_request, "prepared"):
+          return True
+        self._tracker.complete(sync_request, "prepared")
+
+      sampler = getattr(self, "_sampler", None)
```
wang2yn84: Instead of defensive coding like this, you probably should check the existence of sampler and error out if sampler is not available. Then you can just call self.sampler.xxx() everywhere.

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
A9isha: Done. I've removed the job_name and server_id arguments from __init__ and reverted to relying purely on worker_index with the hardcoded "rollout" job name.

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
A9isha: yes it's used to trigger the KV cache clearing and reinitialization during the weight sync lifecycle. Following up on the other review comments, I've also refactored how it's used so we just call a clean `self._sampler.delete_cache()` API rather than leaking the vLLM implementation details here.

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
A9isha: SG - for now, we only expect one sampler per worker right now, so I've removed _sync_lock (and the asyncio import) entirely. I also left a TODO(tunix-dev) to reintroduce a lock if we ever support multiple samplers per worker in the future.

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
A9isha: Done! I've extracted the cache and state_leaves logic into new delete_cache() and reinitialize_cache() methods directly on VllmSampler (with no-op defaults in BaseSampler). The RaidenWeightSyncDelegate now just calls these unified APIs cleanly, and update_params() reuses them as well.

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
A9isha: Done. I've replaced the defensive getattr checks with an explicit if self._sampler is None: raise RuntimeError(...).

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
A9isha: yes hardcoded this to "rollout"

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
```diff
@@ -36,14 +39,25 @@ class RaidenWeightSyncDelegate:
...
-  def __init__(self, *args, worker_index: int = 0, **kwargs):
-    super().__init__(*args, **kwargs)
+  def __init__(
+      self,
+      *args,
+      worker_index: int = 0,
+      sampler: Optional[Any] = None,
+      **kwargs,
+  ):
+    del args, kwargs
```
wang2yn84, line 47: Just fyi, I'm having a small cl/979409558 to fix the broken Tunix head. Let's try to land this PR first if possible so that you don't have to rebase again. But if this PR still needs iterations, I might land that cl first.

tunix/experimental/weight_sync/raiden_weight_sync_delegate.py
```diff
@@ -52,10 +66,12 @@ def is_bounded(
     return all(s.bound for s in self._synchronizers)
 
   async def bind_weight_sync(
-      self, sync_request: Any = None, state: Any = None, **kwargs
+      self, sync_request: Any = None, state: Any = None, sampler: Any = None, **kwargs
   ) -> Any:
     """Binds destination-side transport resources for weight sync."""
     del sync_request, kwargs
+    if sampler is not None:
```
wang2yn84, line 71: Why do we pass sampler in again when sampler is passed in init?

A9isha: Great point! In the actual rollout adapters, the delegate is instantiated before the sampler is fully ready, so we were only passing it to `__init__` in the tests anyway. I've removed sampler from `__init__` entirely to avoid confusion and updated the tests to pass it via `bind_weight_sync()` exactly like the real code does.

wang2yn84: In the future, we might want to have set_samper() for clearer API.

PR #2150, opened 2026-09-09
[Raiden Weight Sync 5/6] MoE 128-lane weight interleaving for TPU GMM layout

YixuanWang-99:
## Overview
**Pairs with MaxText [M2]** (`AI-Hypercomputer/maxtext`).

**Stack:**
- [T1] https://github.com/google/tunix/pull/2146: Multi-host discovery & multi-axis sharding in RaidenHandler
- [T2a] https://github.com/google/tunix/pull/2147: Transport selection, FFI preflight mismatch check, and host-stage normalization
- [T3] https://github.com/google/tunix/pull/2148: Extract RaidenDestinationWeightSyncMixin
- [T4] https://github.com/google/tunix/pull/2149: Cache lifecycle and prefix caching
- **[T5] google/tunix (this PR)**: MoE 128-lane interleaving for TPU GMM layout (pairs with MaxText M2)

## Details
- Implements 128-lane interleaving in `_interleave_moe_weights` using `TPU_V5P_SUBCORE_LANE_SIZE = 128` to match TPU GMM kernel expectations (`gmm_v2.py`).
- Supports negative-axis indexing.
- Handles `flat_state` updating in `transfer_state_directly`.
- Adds dedicated unit test in `tests/generate/utils_test.py` verifying 128-lane interleaving per shard.

## Verification
- `pytest tests/generate/utils_test.py` (65/65 passed).

### DIFF
--- a/tunix/generate/utils.py
+++ b/tunix/generate/utils.py
@@ -157,7 +157,8 @@ def np_find_first_non_pad_idx(ids: np.ndarray, pad_id: int) -> int:
 
 
 def np_find_first_eos_idx(
-    ids: np.ndarray, eos_id: int | jax.Array,
+    ids: np.ndarray,
+    eos_id: int | jax.Array,
 ) -> int:
   """Numpy version of find_first_eos_idx. Works on CPU arrays."""
   assert ids.ndim == 1, f'ids should be a 1d array. Got: {ids.shape}'
@@ -663,16 +664,24 @@ def _align_shape(
           new_tgt_shape = tgt_shape[:-1] + (repeated_dim, padded_dim)
     elif re.compile(r'layers\..*\.moe\.gating_einsum').match(src_key):
       tp_size = kwargs['tp_size']
-      num_experts, expert_dim, embed_dim = val.shape[0], val.shape[2], val.shape[3]
+      num_experts, expert_dim, embed_dim = (
+          val.shape[0],
+          val.shape[2],
+          val.shape[3],
+      )
       gate_chunks, up_chunks = val[:, 0, :, :], val[:, 1, :, :]
       chunk_size = expert_dim // tp_size
-      padded_expert_chunk_dim = ((chunk_size + 127)//128)*128
+      padded_expert_chunk_dim = ((chunk_size + 127) // 128) * 128
       pad_amount = padded_expert_chunk_dim - chunk_size
       gate_chunks = gate_chunks.reshape(num_experts, tp_size, -1, embed_dim)
       up_chunks = up_chunks.reshape(num_experts, tp_size, -1, embed_dim)
       if pad_amount > 0:
-        gate_chunks = jnp.pad(gate_chunks, ((0, 0), (0, 0), (0, pad_amount), (0, 0)))
-        up_chunks = jnp.pad(up_chunks, ((0, 0), (0, 0), (0, pad_amount), (0, 0)))
+        gate_chunks = jnp.pad(
+            gate_chunks, ((0, 0), (0, 0), (0, pad_amount), (0, 0))
+        )
+        up_chunks = jnp.pad(
+            up_chunks, ((0, 0), (0, 0), (0, pad_amount), (0, 0))
+        )
       val_chunks = jnp.stack([gate_chunks, up_chunks], axis=2)
       val_chunks = val_chunks.reshape(num_experts, -1, embed_dim)
       val_chunks = val_chunks.transpose(0, 2, 1)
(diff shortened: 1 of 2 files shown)


tunix/generate/utils.py
```diff
@@ -1210,23 +1210,34 @@ def _align_per_axis(
...
+TPU_V5P_SUBCORE_LANE_SIZE: int = 128
+
+
 def _interleave_moe_weights(
     wi_0: jax.Array | np.ndarray,
     wi_1: jax.Array | np.ndarray,
     tgt_shape: Tuple[int, ...],
     n_shards: int,
     axis: Optional[int] = None,
+    lane_size: int = TPU_V5P_SUBCORE_LANE_SIZE,
```
tianshub: don't you need to config this from the caller side? currently it's always defaults to v5p config, will this still work for other TPU versions?

YixuanWang-99: Updated. 

- For TPU v5p and TPU v7 / v7x (`v5p`, `v7`, `7x`), `get_default_moe_lane_size()` returns 128 (matching the TPU GMM subcore lane requirement).
- For other TPU versions (e.g. v4, v5e), it defaults to `0` (concatenation along the padded axis without interleaving).

tunix/generate/utils.py
```diff
@@ -1210,23 +1210,66 @@ def _align_per_axis(
...
+    if devices:
+      kind = getattr(devices[0], "device_kind", "").lower()
+      if any(arch in kind for arch in ("v5p", "v7", "7x")):
+        return TPU_MOE_SUBCORE_LANE_SIZE
+      if "tpu" in kind:
+        # Other TPU architectures (e.g. v4, v5e) do not require 128-lane interleaving
+        return 0
+  except Exception:
+    pass
+  return TPU_MOE_SUBCORE_LANE_SIZE
```
s-noghabi: should the default fallback be 0 instead?

tunix/generate/utils.py
```diff
@@ -1210,23 +1210,66 @@ def _align_per_axis(
...
+  env_override = os.environ.get("MOE_INTERLEAVE_LANE_SIZE")
+  if env_override is not None:
+    try:
+      return int(env_override)
+    except ValueError:
+      logging.warning(
+          "Invalid MOE_INTERLEAVE_LANE_SIZE=%r; using default.", env_override
+      )
+  try:
+    devices = jax.devices()
```
s-noghabi: can you also add unit test for this recent change?

tunix/generate/utils.py
YixuanWang-99: added fallback to 0

tunix/generate/utils.py
YixuanWang-99: added test here: test_interleave_moe_weights_fallback_when_lane_size_zero

tunix/generate/utils.py
s-noghabi: I still see   return TPU_MOE_SUBCORE_LANE_SIZE as the last return and default fallback.

s-noghabi (changes requested):

s-noghabi (approved):

tunix/generate/utils.py
wang2yn84: I think sublane of 128 has never changes for all the TPU generations. Do we really need to pass this arg around?

tunix/generate/utils.py
```diff
@@ -1210,23 +1210,66 @@ def _align_per_axis(
   return _jit_repeat_axes(arr, tuple(repeats))
 
 
+TPU_MOE_SUBCORE_LANE_SIZE: int = 128
```
wang2yn84: Same, for which TPU we have non 128 value?

tunix/generate/utils.py
```diff
@@ -1210,23 +1210,66 @@ def _align_per_axis(
...
+TPU_V5P_SUBCORE_LANE_SIZE: int = TPU_MOE_SUBCORE_LANE_SIZE
+
+
+def get_default_moe_lane_size() -> int:
```
wang2yn84: Remove if the lane size is always 128

tunix/generate/utils.py
YixuanWang-99: I think tpu v4 isn't... If it is out of scope, we can remove it.

wang2yn84: Check with duckie and all the TPU generations so far have 128 lanes. sublane varies for v7.

tunix/generate/utils.py
wang2yn84: Resolved offline, it's because prior version of TPU doesn't support gemm kernel, so setting lane to 0 to bypass the calculation.

wang2yn84 (approved):

tianshub:
@YixuanWang-99 please check the failed test

FAILED tests/generate/utils_test.py::UtilsTest::test_transfer_state_directly_with_flat_state - AssertionError: expected (frozen)dict; got <class 'utils_test.UtilsTest.test_transfer_state_directly_with_flat_state.<locals>.FakeFlatStateModel'>

PR #2169, opened 2026-09-09
Add various improvements to train_maxtext_nb.py.

niting:
Plumbs various options related to RL training including but not limited to prefix caching, importance sampling, error thresholds, overlong filtering, force_on_policy_ratio etc. Disables mrope for vllm sampling since that is known to be slow and being improved separately but not needed for non-multimodal workloads.

Also, adds appropriate flags for JAX Profiling.

Updates the .yml configuration used to rl.yml from MaxText as this has the correct sharding specs as compared to base.yml.

I ran this change on a v5p-256 cluster.

### DIFF
--- a/examples/deepswe/train_maxtext_nb.py
+++ b/examples/deepswe/train_maxtext_nb.py
@@ -54,6 +54,14 @@ import vllm  # pytype: disable=import-error
 maxtext_vllm_adapter.register()
 logging.info("Successfully registered MaxTextForCausalLM model with vLLM.")
 
+# Disable MRoPE in vLLM to prevent slow 3D position ID generation for text-only sampling
+try:
+  from vllm.config import ModelConfig
+  ModelConfig.uses_mrope = property(lambda self: False)
+  logging.info("Successfully patched vLLM ModelConfig.uses_mrope to False for sampling.")
+except Exception as e:
+  logging.warning("Could not patch vLLM ModelConfig.uses_mrope: %s", e)
+
 faulthandler.register(signal.SIGINT, all_threads=True)
 
 Dataset = datasets_lib.Dataset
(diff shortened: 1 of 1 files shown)


examples/deepswe/train_maxtext_nb.py
tianshub, line 551: we probably don't need this alias...which can reduce quite some boilerplate code...

examples/deepswe/train_maxtext_nb.py
```diff
@@ -996,19 +1073,38 @@ def get_lora_model(base_model, model_mesh):
...
-optimizer = optax.schedules.inject_hyperparams(optax.adamw)(
-    learning_rate=LEARNING_RATE,
+warmup_steps = int(MAX_STEPS * args.warmup_steps_fraction)
+if warmup_steps > 0 or args.learning_rate_final_fraction < 1.0:
+  lr_schedule = optax.warmup_cosine_decay_schedule(
+      init_value=0.0 if warmup_steps > 0 else LEARNING_RATE,
+      peak_value=LEARNING_RATE,
+      warmup_steps=warmup_steps,
+      decay_steps=MAX_STEPS,
+      end_value=LEARNING_RATE * args.learning_rate_final_fraction,
```
tianshub, line 1083: removing the alias will make this more straightforward and it's clear the args value are source of truth.

hgao327 (approved):

examples/deepswe/train_maxtext_nb.py
```diff
@@ -1173,6 +1270,13 @@ def _generate_maxtext_config_with_no_remat(vllm_config_param):
     "loss_agg_mode": LOSS_AGG_MODE,
     "advantage_estimator": ADVANTAGE_ESTIMATOR,
     "use_rollout_logps": USE_ROLLOUT_LOGPS,
+    "force_on_policy_ratio": FORCE_ON_POLICY_RATIO,
+    "sampler_is": args.sampler_is,
+    "overlong_loss_masking": args.overlong_loss_masking,
```
hgao327, line 1275: Seems these arguments are not defined in GRPOConfig or its parent classes, so this may raise an unexpected-keyword TypeError?

hgao327: from 1275 to 1279

PR #2201, opened 2026-09-11
[E2E] End-to-end integration fixes for MaxText training, rollout, and Raiden weight sync

YixuanWang-99:
## Overview
End-to-end integration fixes across MaxText training, rollout, and Raiden weight sync for multi-host RL workflows.

## Changes
- **Raiden Weight Sync (`raiden_synchronizer.py`)**:
  - Filter out inference-only KV cache leaves (`cache`, `cached_prefill_key`, `cached_prefill_value`) during weight binding to avoid manifest preflight destination-only mismatch rejections.
  - Fix `devices_per_host` calculation under Pathways proxy devices (cap at 4 chips/host on Cloud TPU v4/v5p when proxy is used).
  - Canonicalize variable names in tensor metadata manifest using `_param_key(name)` to align trainer (`['base'][...]`) and rollout (`['model'][...]`) key paths.
- **Rollout Node (`run_rollout_node.py`)**:
  - Plumb `MaxTextForCausalLM` architecture override and MaxText config overrides (`model_name`, `model_call_mode="inference"`, `enable_dp_attention=False`, `prefuse_moe_weights`, etc.) via `additional_config` to the in-process vLLM sampler.
- **Trainer Node (`run_trainer_node.py`)**:
  - Default `--prefuse_moe_weights` to `False` for the trainer.
- **K8s Launcher & Config (`k8s_launcher.sh`, `maxtext_utils.py`)**:
  - Support `DISABLE_CHECKPOINTING` environment variable to bypass Orbax checkpoint saving during smoke tests while preserving checkpoint restore.
  - Plumb `ROLLOUT_PREFUSE_MOE_WEIGHTS` (default `true`) to rollout and pass `DISABLE_CHECKPOINTING` to trainer worker startup.
- **GSM8K GRPO Runner (`run_gsm8k_dist_grpo.py`)**:
  - Set root logging level to `DEBUG` when `--debug` is specified.

### DIFF
--- a/tunix/experimental/examples/common/run_rollout_node.py
+++ b/tunix/experimental/examples/common/run_rollout_node.py
@@ -30,6 +30,7 @@ from typing import Any
 from tunix.experimental.examples.common import models
 from tunix.experimental.weight_sync import weight_sync as weight_sync_lib
 from tunix.rl.agentic.parser.chat_template_parser import parser as chat_parser_lib
+from tunix.utils import maxtext_utils
 
 REPO_ROOT = os.path.abspath(
     os.path.join(os.path.dirname(__file__), "..", "..", "..", "..")
@@ -390,6 +391,27 @@ def _create_inprocess_vllm_sampler(args, tokenizer):
       "max_model_len": max_model_len,
       "enable_prefix_caching": args.enable_prefix_caching,
   }
+  # Select MaxText's `MaxTextForCausalLM` as rollout model.
+  # `additional_config` must be set on VllmConfig (not engine_kwargs): the
+  # sampler overwrites args["additional_config"] from the VllmConfig field.
+  maxtext_additional_config = None
+  if args.maxtext_model_name:
+    logging.info(
+        "Loading MaxText model %r natively via maxtext_vllm_adapter's"
+        " MaxTextForCausalLM (architectures override).",
+        args.maxtext_model_name,
+    )
+    engine_kwargs["hf_overrides"] = dict(
+        maxtext_utils.VLLM_MAXTEXT_HF_OVERRIDES
+    )
+    maxtext_additional_config = (
+        maxtext_utils.build_vllm_maxtext_additional_config(
+            args.maxtext_model_name,
+            attention=args.maxtext_attention,
+            prefuse_moe_weights=args.prefuse_moe_weights,
+        )
+    )
+
   if multihost_backend:
     engine_kwargs["distributed_executor_backend"] = multihost_backend
   server_mode = True if multihost_backend else None
@@ -419,6 +441,7 @@ def _create_inprocess_vllm_sampler(args, tokenizer):
       return_logprobs=True,
       lora_config=lora_config,
       mapping_config=mapping_config,
+      additional_config=maxtext_additional_config,
       engine_kwargs=engine_kwargs,
   )
   sampler_adapter = inprocess_vllm_sampler_adapter.InprocessVllmSamplerAdapter(
@@ -483,23 +506,16 @@ def _create_vllm_sampler(args):
         " MaxTextForCausalLM (architectures override).",
         args.maxtext_model_name,
     )
-    engine_kwargs["hf_overrides"] = {"architectures": ["MaxTextForCausalLM"]}
-    # MaxText inference config.
-    maxtext_config_overrides = {
-        "model_name": args.maxtext_model_name,
-        "model_call_mode": "inference",
-        "enable_dp_attention": False,
-        "allow_split_physical_axes": True,
-        "log_config": False,
-        "weight_dtype": "bfloat16",
-    }
-    if args.prefuse_moe_weights is not None:
-      maxtext_config_overrides["prefuse_moe_weights"] = args.prefuse_moe_weights
-    if args.maxtext_attention:
-      maxtext_config_overrides["attention"] = args.maxtext_attention
-    engine_kwargs["additional_config"] = {
-        "maxtext_config": maxtext_config_overrides
-    }
+    engine_kwargs["hf_overrides"] = dict(
+        maxtext_utils.VLLM_MAXTEXT_HF_OVERRIDES
+    )
+    engine_kwargs["additional_config"] = (
+        maxtext_utils.build_vllm_maxtext_additional_config(
+            args.maxtext_model_name,
+            attention=args.maxtext_attention,
+            prefuse_moe_weights=args.prefuse_moe_weights,
+        )
+    )
   engine_args = AsyncEngineArgs(**engine_kwargs)  # pytype: disable=bad-argument-type  # type: ignore[arg-type]
   sampler_adapter = vllm_sampler_adapter.VllmSamplerAdapter(  # pytype: disable=bad-instantiation  # type: ignore[abstract]
       server_id=args.worker_id,
--- a/tunix/experimental/examples/common/run_trainer_node.py
+++ b/tunix/experimental/examples/common/run_trainer_node.py
@@ -198,10 +198,12 @@ def _parse_args(argv: list[str]) -> argparse.Namespace:
   parser.add_argument(
       "--prefuse_moe_weights",
       type=_str2bool,
-      default=True,
+      default=False,
       nargs="?",
       const=True,
-      help="Prefuse MoE weights for MaxText training.",
+      help=(
+          "Prefuse MoE weights (w0/w1). Off for the trainer."
+      ),
   )
   parser.add_argument(
       "--use_weight_converter",
@@ -368,16 +370,37 @@ class _MeshBoundTrainer:
       self._trainer.close()
 
 
+def _checkpointing_options(args) -> Any:
+  """Builds the Orbax options; `save_interval_steps=0` means "never save".
+
+  Orbax's FixedIntervalPolicy takes `step % save_interval_steps`, so 0 has to
+  become `read_only` rather than being passed through. Restoring is a separate
+  path and keeps working, which is what makes 0 usable for short smoke tests.
+  """
+  if args.checkpoint_save_interval_steps < 0:
+    raise ValueError(
+        "checkpoint_save_interval_steps must be non-negative, got"
+        f" {args.checkpoint_save_interval_steps}."
+    )
+  if args.checkpoint_save_interval_steps == 0:
+    logging.info(
+        "checkpoint_save_interval_steps=0; checkpoint saving is disabled "
+        "(restore is unaffected)."
+    )
+    return ocp.CheckpointManagerOptions(read_only=True)
+  return ocp.CheckpointManagerOptions(
+      save_interval_steps=args.checkpoint_save_interval_steps,
+      max_to_keep=args.checkpoint_max_to_keep,
+  )
+
+
 def _create_maxtext_trainer_factory(args) -> Any:
   """Creates the trainer factory function for MaxText's MaxTextTrainingEngine."""
   logging.info("Trainer backend: MaxText's MaxTextTrainingEngine.")
   pad_id = maxtext_utils.get_tokenizer_pad_id(
       args.model_id, args.tokenizer_path, args.model_dir
   )
-  checkpointing_options = ocp.CheckpointManagerOptions(
-      save_interval_steps=args.checkpoint_save_interval_steps,
-      max_to_keep=args.checkpoint_max_to_keep,
-  )
+  checkpointing_options = _checkpointing_options(args)
   grad_accumulation_steps = max(
       1, math.ceil(args.mini_batch_size / args.train_micro_batch_size)
   )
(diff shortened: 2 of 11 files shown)


tunix/experimental/examples/common/run_rollout_node.py
```diff
@@ -395,6 +390,31 @@ def _create_inprocess_vllm_sampler(args, tokenizer):
...
+  # sampler overwrites args["additional_config"] from the VllmConfig field.
+  maxtext_additional_config = None
+  if args.maxtext_model_name:
+    logging.info(
+        "Loading MaxText model %r natively via maxtext_vllm_adapter's"
+        " MaxTextForCausalLM (architectures override).",
+        args.maxtext_model_name,
+    )
+    engine_kwargs["hf_overrides"] = {"architectures": ["MaxTextForCausalLM"]}
+    maxtext_config_overrides = {
```
s-noghabi: can we have this in a helper function, or even better in maxtext_utils?

tunix/experimental/examples/math_gsm8k_dist/k8s_launcher.sh
```diff
@@ -66,6 +66,9 @@ export USE_ROLLOUT_LOGPS=${USE_ROLLOUT_LOGPS:-true}
 export CHECKPOINT_SAVE_INTERVAL_STEPS=${CHECKPOINT_SAVE_INTERVAL_STEPS:-1}
 export CHECKPOINT_MAX_TO_KEEP=${CHECKPOINT_MAX_TO_KEEP:-10}
 export CHECKPOINT_ROOT_DIRECTORY=${CHECKPOINT_ROOT_DIRECTORY:-checkpoints}
+# Skip all Orbax checkpoint saves
+# Restoring from MAXTEXT_CKPT is unaffected. Intended for short smoke tests.
+export DISABLE_CHECKPOINTING=${DISABLE_CHECKPOINTING:-false}
```
s-noghabi: you can use CHECKPOINT_SAVE_INTERVAL_STEPS=0 to detect this.

tunix/experimental/examples/math_gsm8k_dist/k8s_launcher.sh
```diff
@@ -79,6 +82,7 @@ export ROLLOUT_MAXTEXT_ATTENTION=${ROLLOUT_MAXTEXT_ATTENTION:-}
 
 # MoE & Weight Sync Flags
 export PREFUSE_MOE_WEIGHTS=${PREFUSE_MOE_WEIGHTS:-true}
+export ROLLOUT_PREFUSE_MOE_WEIGHTS=${ROLLOUT_PREFUSE_MOE_WEIGHTS:-true}
```
s-noghabi: iiuc we don't prefuse weights for rollouts, do we? do we really need this?

tunix/experimental/weight_sync/raiden_synchronizer.py
```diff
@@ -291,6 +316,11 @@ def _devices_per_host(devices: List[Any]) -> int:
         "no task_id on any of %d device(s); assuming a single host",
         len(devices),
     )
+    if "proxy" in os.environ.get("JAX_PLATFORMS", ""):
+      # Cloud TPU v4/v5p VMs are always 4 chips/host. Under Pathways every
```
s-noghabi: this seems very v4/v5p specific. would this even work for others?

tunix/utils/maxtext_utils.py
```diff
@@ -268,7 +268,18 @@ def build_maxtext_config(
       ),
   ])
   logging.info("MaxText config argv: %s", argv)
-  return pyconfig.initialize(argv)
+  config = pyconfig.initialize(argv)
+
+  # DISABLE_CHECKPOINTING turns off checkpoint *saving* only; restoring from
+  # `load_parameters_path` still works.
+  if os.environ.get("DISABLE_CHECKPOINTING", "").lower() in ("1", "true", "yes"):
```
s-noghabi: rely on checkpoint_step=0

tunix/experimental/weight_sync/raiden_synchronizer.py
```diff
@@ -238,13 +254,22 @@ def _filter_bindable(
...
+    if _is_kv_cache(name):
+      cache_dropped.append(name)
+    elif _bindable(arr, allow_proxy=allow_proxy):
       arr.block_until_ready()
       keep_names.append(name)
       keep_arrays.append(arr)
     else:
       dropped.append(name)
+  if cache_dropped:
+    logging.info(
```
s-noghabi: would this clutter the log? should we make it debug?

s-noghabi (changes requested):

tunix/experimental/examples/math_gsm8k_dist/k8s_launcher.sh
YixuanWang-99: We need prefuse moe weights for qwen3.5-35B for rollout with maxtext runner model.

tunix/experimental/weight_sync/raiden_synchronizer.py
YixuanWang-99: set RAIDEN_DEVICES_PER_HOST=8 in env covers v7 and escape the hard-code here. Patched the comment.

tunix/experimental/examples/common/run_rollout_node.py
YixuanWang-99: Moved

tunix/experimental/examples/math_gsm8k_dist/k8s_launcher.sh
YixuanWang-99: Done

tunix/utils/maxtext_utils.py
YixuanWang-99: Done

tunix/experimental/weight_sync/raiden_synchronizer.py
YixuanWang-99: done

tunix/experimental/examples/common/run_trainer_node.py
```diff
@@ -368,16 +370,33 @@ def close(self) -> None:
...
 
 
+def _checkpointing_options(args) -> Any:
+  """Builds the Orbax options; `save_interval_steps=0` means "never save".
+
+  Orbax's FixedIntervalPolicy takes `step % save_interval_steps`, so 0 has to
+  become `read_only` rather than being passed through. Restoring is a separate
+  path and keeps working, which is what makes 0 usable for short smoke tests.
+  """
+  if args.checkpoint_save_interval_steps <= 0:
```
s-noghabi: we should not allow negative.

tunix/experimental/examples/math_gsm8k_dist/k8s_launcher.sh
s-noghabi: Do we need two different flags for rollout and trainer? wouldn't a single flag be enough?

tunix/experimental/weight_sync/raiden_synchronizer.py
s-noghabi: I still see the 4. Can you double check? also, can we remove the unnecessary comment?

tunix/experimental/weight_sync/raiden_synchronizer.py
s-noghabi: I still see logging.info. can you double check

s-noghabi (changes requested):

tunix/experimental/weight_sync/raiden_synchronizer.py
wang2yn84: Is RAIDEN_DEVICES_PER_HOST different from normal device per host?

wang2yn84 (commented):
Please squash commits.

tunix/experimental/examples/common/run_trainer_node.py
YixuanWang-99: Updated: added validation to disallow negative values (`--checkpoint_save_interval_steps < 0` raises `ValueError` in both CLI parsing and config builders), allowing only `0` as the sentinel to disable saving. Added unit test coverage.

tunix/experimental/examples/math_gsm8k_dist/k8s_launcher.sh
YixuanWang-99: Consolidated: removed `ROLLOUT_PREFUSE_MOE_WEIGHTS` and kept the single `PREFUSE_MOE_WEIGHTS` flag for the launcher (passed to rollout, matching the recipe).

tunix/experimental/weight_sync/raiden_synchronizer.py
YixuanWang-99: Updated: removed the hardcoded `4` and the comment. `_devices_per_host` now groups devices dynamically using `mesh.device_host_key(d)`, which resolves Pathways proxy devices via `logical_task` and non-Pathways devices via `process_index`. This automatically determines the correct devices per host across topologies (e.g. 4 for v4/v5p, 8 for v7x).

tunix/experimental/weight_sync/raiden_synchronizer.py
YixuanWang-99: It was changed to `logging.debug` in commit 14cba3b6 (line 268 at HEAD). The thread diff view was anchored to the earlier commit.

tunix/experimental/weight_sync/raiden_synchronizer.py
YixuanWang-99: RAIDEN_DEVICES_PER_HOST is exactly normal device per host: For v7, it is 8 devices per host. And for v5p, it is 4 devices per host. But standard JAX APIs could not distinguish hosts under Pathways, and Raiden is sensitive to this count during weight sync.

wang2yn84: I think we do have a way to differentiate hosts with Pathways and McJax: https://github.com/google/tunix/blob/f3da0fe6c2ef14ce54551acc7c7f55aac57b7fe2/tunix/utils/mesh.py#L231

tunix/utils/topology.py
```diff
@@ -378,10 +380,14 @@ def _device_host_key(device: Any) -> tuple[Any, ...] | None:
     task_id = _pathways_device_host_attr(device, "logical_task")
   else:
     task_id = _device_attr(device, "process_index")
+  if task_id is None:
```
s-noghabi: this is duplicate with above.

tunix/utils/topology.py
```diff
@@ -378,10 +380,14 @@ def _device_host_key(device: Any) -> tuple[Any, ...] | None:
...
+    task_id = _pathways_device_host_attr(device, "logical_task")
   if task_id is None:
     return None
 
   slice_id = _device_attr(device, "slice_index", None)
+  if not isinstance(slice_id, int):
```
s-noghabi: when will it not be int (other than none)

tunix/experimental/weight_sync/raiden_synchronizer.py
wang2yn84: Do we still need RAIDEN_DEVICES_PER_HOST then? Shall we remove it?

wang2yn84: And there are quite a few places uses RAIDEN_DEVICES_PER_HOST. Maybe also remove them?

tunix/utils/topology.py
YixuanWang-99: Done. Removed the duplicate lookup.

tunix/utils/topology.py
YixuanWang-99: Done. Removed the check.

tunix/experimental/weight_sync/raiden_synchronizer.py
YixuanWang-99: Done. Removed RAIDEN_DEVICES_PER_HOST completely across the repo (in `raiden_synchronizer.py`, `k8s_launcher.sh`, `trellis_gsm8k_qwen3p5_35b.sh`, and unit tests). Now that `mesh.device_host_key` dynamically resolves hosts for both Pathways and McJax, `_devices_per_host` auto-derives the counts across topologies without needing the environment variable.

wang2yn84: Thank you!

tunix/utils/topology.py
```diff
@@ -232,13 +233,14 @@ def _shape_fits_within(
   return all(dim <= limit for dim, limit in zip(shape, available_shape))
 
 
-@functools.cache
```
s-noghabi, line 235: why remove this?

s-noghabi (approved):

tunix/utils/topology.py
YixuanWang-99: Restored `@functools.cache`. (It was inadvertently dropped earlier when consolidating the Pathways backend detection helpers into `_is_pathways_backend_used`).

PR #2228, opened 2026-09-15
fix(checkpointing, infra): orchestrator save interval, save-sync serialization, memory requests, and fail-fast

YixuanWang-99:
### Summary
This PR addresses several checkpointing lifecycle, host-memory pressure, and stability issues during distributed RL training with MaxText and Raiden weight sync:

1. **Configurable Orchestrator Save Interval**:
   - Honor `checkpoint_save_interval_steps` inside `StandardRLProgram.train_stage` so checkpoints are only saved on designated step boundaries.
   - Expose `--checkpoint_save_interval_steps` in `run_gsm8k_dist_grpo.py` and `k8s_launcher.sh`.
   - When `checkpoint_save_interval_steps=0` (save disabled), keep `enable_checkpointing=True` in `maxtext_utils.py` with a large `checkpoint_period` when `load_parameters_path` is present so base weights restore without error.
   - Suppress redundant final checkpoint writes in `_MeshBoundTrainer.close()` if step was already checkpointed or saving is disabled.

2. **Serialize Checkpoint Staging Against Weight Sync**:
   - In `_MeshBoundTrainer`, call `_drain_inflight_checkpoint()` before `prepare_weight_sync()` to ensure asynchronous Orbax device-to-host staging completes before Raiden initiates weight transfer, preventing concurrent host-memory OOM.
   - Forward `CKPT_D2H_CONCURRENT_GB` in `maxtext_utils.py` to tune device-to-host staging concurrency without rebuilding containers.

3. **Container Memory Requests & Limits in Deployment**:
   - Add `--user_container_memory`, `--user_container_memory_limit`, and `--pathways_worker_memory` in `yaml_generator.py` and `jobset.pathways.yaml`.
   - Rebalance container reservations to prevent BestEffort eviction during checkpoint staging spikes while protecting node allocatable limits.
   - Set `restartPolicy: Always` on user container so it restarts cleanly in place upon unrecoverable errors.

4. **Fail-Fast on Accelerator Device-Set Loss**:
   - In `remote_execution.py`, detect unrecoverable Pathways/device loss signatures and schedule a graceful exit so Kubernetes/JobSet can rebuild the worker rather than leaving a zombie trainer process.

### Counterpart PR
- Paired with MaxText PR: https://github.com/AI-Hypercomputer/maxtext/pull/5234


tunix/experimental/distributed/deployment/yamls/jobset.pathways.yaml
```diff
@@ -69,7 +69,7 @@ ${QUEUE_LABEL}spec:
               resources:
                 limits:
                   cpu: "24"
-                  memory: 32G
+                  memory: 16G
```
tianshub: why is this changed?

tunix/experimental/orchestrator/rl_program.py
```diff
@@ -903,6 +905,19 @@ async def train_stage(self) -> None:
 
       async def _maybe_save_checkpoint() -> None:
         nonlocal checkpoint_saved
+        interval = self.checkpoint_save_interval_steps
```
tianshub: why do we want to control the CP save interval here? the current behavior is at every global step, which suppose to be the best option for full checkpoint. Also orbax will overwrite previous ones based on config. Is there a reason we want to make it more coarse grained?

tunix/experimental/worker/remote_execution.py
```diff
@@ -224,6 +226,60 @@ def unwrap(self) -> Any:
     return self.result
 
 
+_UNRECOVERABLE_HARDWARE_SIGNATURES = (
```
tianshub: do you see failure cases that didn't properly propagate up by remote_execution? we didn't want to bake in too much specific failure handling mechanism other than network errors, and rest should just propagate up.

tianshub (changes requested):

tunix/experimental/examples/common/run_trainer_node.py
```diff
@@ -541,20 +561,98 @@ def compile(self, *args, **kwargs) -> None:
     with self._mesh:
       self._trainer.compile(*args, **kwargs)
 
+  def _drain_inflight_checkpoint(self) -> None:
+    """Blocks until any in-flight checkpoint write has finished.
+
+    Orbax saves asynchronously, so save_checkpoint() returns while staging
```
s-noghabi: we usually do these b2b, where are the end of a global step you capture a ckpt and in parallel do weight transfer. this would add a hidden noticable latency. (we will be similar to a file based transfer -- at least the write path). Do we really want to do this?

tunix/experimental/distributed/deployment/yamls/jobset.pathways.yaml
```diff
@@ -122,10 +122,11 @@ ${QUEUE_LABEL}spec:
               imagePullPolicy: Always
               securityContext:
                 privileged: true
-              # resources:
-              #   limits:
-              #     cpu: "24"
-              #     memory: 32G
+              resources:
```
s-noghabi, line 135: Because the head proc pod is co-located with the pw-node pod via podAffinity, omitted memory requests on pathways-proxy default to requests = limits (up to 190G). Total node memory requested by the co-located containers reaches ~431G (or 341G at minimum), which exceeds the allocatable RAM of standard TPU v5e (208 GB) and TPU v5p (256 GB) VMs, causing the head pod to hang in Pending: Insufficient memory.

tunix/experimental/worker/remote_execution.py
s-noghabi: +1. this is breaking the intended abstractions.

tunix/experimental/examples/common/run_trainer_node.py
```diff
@@ -541,20 +561,98 @@ def compile(self, *args, **kwargs) -> None:
...
   def save_checkpoint(self, metadata: Any = None, **kwargs) -> None:
+    if not self._save_enabled:
+      logging.info(
+          "checkpoint_save_interval_steps=0; skipping save_checkpoint request."
+      )
+      return
     with self._mesh:
       self._trainer.save_checkpoint(metadata, **kwargs)
+      self._last_saved_train_step = getattr(
+          self._trainer, "train_step", None
```
s-noghabi: in peft_trainer, train_step is a method, not a property. this logic will break for that path.

s-noghabi (changes requested):

tunix/experimental/distributed/deployment/yamls/jobset.pathways.yaml
YixuanWang-99: Reverted. 

This is was legacy attempt to fix OOM. Never mind.

tunix/experimental/orchestrator/rl_program.py
YixuanWang-99: Good catch, fell back to save at every step. 

checkpoint_save_interval_steps was fed in the legacy `train_rl` path as Orbax's CheckpointManagerOptions
(same as train_rl.py:325). Now it is removed.

tunix/experimental/worker/remote_execution.py
YixuanWang-99: Removed. 

This is another problem I met, where the trainer stayed alive serving errors forever instead of letting the JobSet rebuild it. I removed it from here

tunix/experimental/distributed/deployment/yamls/jobset.pathways.yaml
YixuanWang-99: Fixed, now the budget is 
```
  pathways-rm      4G   (limit 32G)
  pathways-proxy  16G   (limit 190G)
  user container  48G   (limit 70G)
  pathways-worker 100G
  ---------------------
  total          168G
  ```
  Now it fits v5e and v5p.

File: tunix/experimental/examples/common/run_trainer_node.py (hunk 6 above)
Comment (YixuanWang-99) [reply tunix/experimental/examples/common/run_trainer_node.py]: Good catch. 

Added _current_train_step(), which probes ("train_steps", "train_step") in that order behind an isinstance(value, int) guard. That's unambiguous for both backends: PeftTrainer has train_steps as the int property with train_step as a method, MaxTextTrainingEngine has train_step as an int property and no train_steps attribute at all. Regression tests added for both.

File: tunix/experimental/examples/common/run_trainer_node.py (hunk 4 above)
Comment (YixuanWang-99) [reply tunix/experimental/examples/common/run_trainer_node.py]: You are correct. 

we want them to run in parallel, fixed the code. But I leave the flag just in case the raiden weight sync and checkpointing overlapping OOM again, in that case we can quickly catch it.

File: tunix/experimental/examples/common/run_trainer_node.py [hunk 7]
```diff
@@ -512,9 +523,20 @@ def _load_actor_model(args, mesh: Mesh, *, lora: bool):
 class _MeshBoundTrainer:
   """Binds generic PeftTrainer v2 calls to this worker's JAX mesh."""
 
-  def __init__(self, trainer: peft_trainer_v2.PeftTrainer, mesh: Mesh):
+  def __init__(
```
tianshub, line 516: arr...we are making change in this _MeshBoundTrainer which suppose to be removed... this is a tech debt which we wanted to clean up, also it's awkward to have critical path logic exist in example folder. I certainly don't wanna put this burden on you Yixuan...but have we thought about putting this in the maxtext trainer directly?

tianshub (approved):

PR #2229, opened 2026-09-15
fix(rollout): enable data parallelism and manifest name canonicalization for SAMPLER=vllm

YixuanWang-99:
### Summary
This PR enables correct tensor and data parallel rollout execution and weight synchronization when using `SAMPLER=vllm` (tpu-inference):

1. **Forward Data Parallelism to vLLM AsyncEngineArgs**:
   - In `run_rollout_node.py`, compute `dp_size = max(1, int(getattr(args, "mesh_fsdp", 1) or 1))` and forward `data_parallel_size=dp_size` to `AsyncEngineArgs`.
   - Validate that `use_lora` is not combined with rollout DP (`mesh_fsdp > 1`).
   - Enable debug logging and set `TUNIX_LOG_ROLLOUT_TEXT=1` when `args.debug` is passed.

2. **Canonicalize Variable Names in Raiden Manifest**:
   - In `vllm_sampler_adapter.py`, add `_canonicalize_variable_names` using `raiden_synchronizer._param_key`.
   - `tpu_inference.rl.raiden_worker_sync` formats destination variable names with bracketed syntax (e.g. `['base']['decoder']...['scale'].value`). Canonicalizing to dotted notation ensures exact-name preflight pairing against source parameters published by `RaidenSynchronizer` without modifying data-plane positional binding.


tunix/experimental/examples/common/run_rollout_node.py
```diff
@@ -566,17 +566,26 @@ def _create_vllm_sampler(args, tokenizer):
...
-      "max_model_len=%d...",
+      "data_parallel_size=%d (%d chips) max_model_len=%d...",
       vllm_model,
       tp_size,
+      dp_size,
+      tp_size * dp_size,
       max_model_len,
   )
+  if dp_size > 1 and args.use_lora:
+    raise ValueError(
```
tianshub, line 580: ah is this a vLLM limitation?

tianshub (approved):

s-noghabi (approved):

tunix/experimental/examples/common/run_rollout_node.py
YixuanWang-99: Yes, actually it comes from tpu-inference rather than core vLLM. 

ShardingConfigManager.validate raises "LoRA is not supported with data parallelism (DP size: N)" whenever total_dp_size > 1 and lora_config is set: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/common/sharding.py#L321-L329

PR #2233, opened 2026-09-15
Overlap detokenization and logprob extraction with decode in VllmSampler

gutianyu-google:
## What

`VllmSampler.generate` only starts decoding text and extracting per-token logprobs after the last request has finished, serially in the main thread: offline mode blocks in `LLM.generate`, server mode blocks on every driver future, and `detokenize` then walks all completions. On a 2048 x 8 rollout that tail is a few seconds per RL step during which the accelerators idle.

This PR hands each request to a small thread pool as soon as it finishes, so its decode and logprob extraction overlap with the remaining decode steps:
- offline mode: the sampler drives the engine loop itself (the same `add_request` / `step` loop as `LLM._run_engine`, request ids from the LLM's own counter, progress bar kept);
- server mode: driver futures are consumed with `as_completed`.

`detokenize` reads the precomputed results and falls back to the synchronous path for anything not precomputed. Outputs are identical. `VllmConfig.overlap_postprocessing` (and `RolloutConfig.rollout_vllm_overlap_postprocessing`) default to `True`; `False` restores the previous behaviour. `postprocessing_threads` (default 4) is validated to be >= 1 and the pool is shut down in `stop()`. Beam search keeps the `LLM.generate` path.

## Effect

Qwen3-0.6B GRPO on TPU v7x: about 2 s less per RL step (detokenize 3.9 s -> ~0 on the critical path, logprob extraction overlapped), outputs identical.

## Tests

`tests/generate/vllm_sampler_test.py` (mocked engine / driver, CPU):
- offline overlap: request ordering, `LLM.generate` bypassed, `detokenize` uses the precomputed results, each request decoded exactly once;
- knob off: `LLM.generate` is called exactly as before;
- beam search stays on `LLM.generate`;
- server mode overlap decodes as futures resolve;
- `detokenize` without precomputed results decodes inline;
- `postprocessing_threads` validation; `stop()` shuts the pool down.


tunix/common/configs.py
```diff
@@ -189,6 +189,11 @@ class RolloutConfig:
   # Set to a smaller value to reduce peak HBM pressure on large models.
   rollout_vllm_reshard_chunk_size: Optional[int] = None
 
+  # Decode text / extract logprobs of each finished request in a thread pool
+  # while vLLM keeps decoding the rest of the batch (offline and server mode).
+  # Outputs are identical; False post-processes after generation as before.
+  rollout_vllm_overlap_postprocessing: bool = True
```
tianshub, line 201: are there any concerns to always overlap? or this is more like a defensive knob?

tunix/generate/vllm_sampler.py
```diff
@@ -476,6 +502,96 @@ def detokenize(
...
+    pool = self._get_postprocess_pool()
+    self._postprocessed = None
+    engine = self.llm.llm_engine  # pyrefly: ignore[missing-attribute]
+    counter = self.llm.request_counter  # pyrefly: ignore[missing-attribute]
+    sampling_params.output_kind = RequestOutputKind.FINAL_ONLY
+    for prompt in prompts:
+      engine.add_request(str(next(counter)), prompt, sampling_params)
+    outputs: List[RequestOutput] = []
+    futures: Dict[str, concurrent.futures.Future[Any]] = {}
+    progress = tqdm.tqdm(
```
tianshub, line 600: do you need tqdm for this? curious how this will render in the console

tianshub (approved):

tunix/common/configs.py
gutianyu-google: Yes, this is more like a defensive knob.

tunix/generate/vllm_sampler.py
gutianyu-google: tqdm is already imported. It renders as one in-place-updating line:

```
Processed prompts:  63%|██████▎   | 1291/2048 [00:41<00:24, 31.2it/s]
```

PR #2239, opened 2026-09-15
fix(trainer): plumb max_seq_token_per_tpu into MaxText's max_target_length

igorts-git:
### Summary

Under sequence packing the trainer is fed rows of `max_seq_token_per_tpu` tokens, but `build_maxtext_config` derives `max_target_length` from `max_prompt_length + max_response_length` — the width of *one* trajectory — and the trainer node has no way to learn the orchestrator's packing budget. This plumbs it through so the MaxText config declares the width of the rows it actually sees.

This is bookkeeping, not a shape requirement. MaxText takes its shapes from the batch it is handed, so a packed row wider than `max_target_length` runs fine (verified by building a small MaxText model with `max_target_length=16` and pushing a 64-token row through the `dot_product` training path). Nothing in tunix or MaxText imposes an upper bound tying the packing budget to `max_target_length` — `validate_packing_budget` enforces only a lower bound. What `max_target_length` does drive is everything `MaxTextConfig` derives from it: per-device TFLOPs accounting, and the divisibility checks for `num_vocab_tiling`, context parallelism and `num_moe_token_chunks`. At `max_seq_token_per_tpu=4096` with a `max_prompt + max_response` floor of 1536, those are all computed for a row 2.7x narrower than the ones being trained on.

### Key Changes
- **`tunix/utils/maxtext_utils.py`**:
  - Accept `max_seq_token_per_tpu: int | None = 0` in `build_maxtext_config`.
  - Validate non-negative input (`ValueError` on `< 0`).
  - Raise `max_target_length` to `max_seq_token_per_tpu` when it exceeds the `max_prompt_length + max_response_length` floor.
  - Warn when `0 < max_seq_token_per_tpu < max_target_length` — not a legal packing budget, `validate_packing_budget` rejects it on the learner.
- **`tunix/experimental/examples/common/run_trainer_node.py`**:
  - Add `--max_seq_token_per_tpu` CLI argument (documented as MaxText-backend only).
  - Plumb it to `build_maxtext_config` (and drop the legacy `inspect.signature` reflection).
- **`tunix/experimental/examples/math_gsm8k_dist/`**:
  - Pass `--max_seq_token_per_tpu` to the trainer in both `k8s_launcher.sh` and local `launcher.sh`. Both launchers already passed it to the orchestrator.
- **Unit Tests**:
  - `tests/utils/maxtext_utils_test.py`: `max_seq_token_per_tpu` raising `max_target_length`, sub-floor warning, unset, `None` handling, negative validation.
  - `tests/experimental/examples/common/run_trainer_node_test.py`: CLI parsing (default and custom), propagation to `build_maxtext_config`, and that it stays *off* `TrainingConfig`.

### Why the flag is MaxText-only

It is deliberately not forwarded to `peft_trainer_v2.TrainingConfig`. Under the orchestrator the Tunix trainer never packs — the `SequencePackedBatchAssembler` does, and the weighting packing needs rides in on the payload (`_fwd_bwd_step` accumulates with `denom=aux.primary_loss.denominator`). Walking the three things the field gates:

- `supports_sequence_packing` is `True` for `PeftTrainer`, so the `__init__` guard is a no-op.
- The fused `train_step()` it also gates is unreachable here: `distributed_rl_engine.train_step` always calls `fwd_bwd` then `update`.
- What remains is `_is_single_microstep()` → `GradientAccumulator(allocate_grads)`, which only differs when `mini_batch_size * num_generations / train_micro_batch_size == 1`. The assembler emits `trainer_fsdp * trainer_dp` rows per micro-batch, so packing cannot push the row count below the unpacked nominal unless `train_micro_batch_size` exceeds that product — which the recipes rule out by pinning `TRAIN_MICRO_BATCH_SIZE` to `TRAINER_MESH_FSDP`. Even then the failure is a loud `ValueError: The gradient accumulator is empty`, not silent mis-normalization.

A test pins that it stays off `TrainingConfig` so this does not get re-added by accident.

### Verification

Measured on Qwen3.5-35B-A3B with `max_prompt=512`, `max_response=1024` and a packing budget of 4096 tokens per row: a step packs ~7.6 trajectories into each row and drops from 32 microbatches to 5.

Unit tests:
```
tests/utils/maxtext_utils_test.py: 20 passed
tests/experimental/examples/common/run_trainer_node_test.py: 47 passed
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tunix/experimental/examples/math_gsm8k_dist/launcher.sh
```diff
@@ -497,6 +497,9 @@ echo "Launching trainer node on TPU chips $TRAINER_TPU_CHIPS..."
   if [[ -n "$MAXTEXT_MODEL_NAME" ]]; then
     TRAINER_CMD+=(--maxtext_model_name="$MAXTEXT_MODEL_NAME")
   fi
+  if [[ -n "$MAX_SEQ_TOKEN_PER_TPU" ]]; then
+    TRAINER_CMD+=(--max_seq_token_per_tpu="$MAX_SEQ_TOKEN_PER_TPU")
```
lc5211, line 502: Why does the trainer need to know the max_seq_token_per_tpu? Is it for the warmup compilation purpose?

igorts-git: Actually, it is not used for warmup compilation. For Maxtext we use it to set the `--max_target_length` config parameter, which in turn used to compute the TFLOPs and other stats.
After investigating a bit more I dropped the part that also sent this flag to the PeftTrainer, because it is not needed there. Also updated the PR description.

lc5211: Yeah, thanks! PeftTrainer doesn't have stats depending on the max token lengths so it can be ignored from peft trainer's config.

tunix/utils/maxtext_utils.py
```diff
@@ -298,13 +303,45 @@ def build_maxtext_config(
...
+        " the trainer is fed are max_seq_token_per_tpu wide.",
+        max_target_length,
+        max_seq_token_per_tpu,
+    )
+    max_target_length = max_seq_token_per_tpu
+  elif (
+      max_seq_token_per_tpu is not None
+      and 0 < max_seq_token_per_tpu < max_target_length
+  ):
+    logging.warning(
```
lc5211, line 311: Do we want to terminate early or just log warning?

lc5211 (approved):

tunix/utils/maxtext_utils.py
igorts-git: I am fine with just the warning, since it is not critical.

PR #2254, opened 2026-09-16
ci: halve per-PR TPU slice demand and add job timeouts

shz-google:
## Problem

The self-hosted v6e-8 pool is capped at **4 concurrent slices** (two runner sets, 2 each — never exceeded across a 43h measurement window). `tpu-tests.yml` fanned out **4 parallel TPU jobs** off a single `needs:`, so **one PR consumed 100% of the pool** and any second consumer — another PR, or the nightly regression — was starved by construction.

Measured over 2026-09-14 23:16 → 2026-09-16 17:54 UTC (283 TPU-labelled jobs):

| Metric | Value |
| --- | --- |
| Jobs that never received a runner | **47 / 265 (18%)** |
| …of those, waited 4–12h before being discarded | 17 |
| Peak backlog | 14 jobs waiting |
| Time with at least 1 job blocked | 22.0h (52%) |
| Time with **0** slices busy | 27.0h (**63%**) |
| Mean slices busy | 1.00 of 4 |

Demand is bursty and synchronized, not large in aggregate — so this is addressed by reducing fan-out rather than by asking for more hardware.

Note: jobs that appear to run for 6–12h never actually executed. They have an empty `runner_name` and zero steps — they were queued, then discarded. This is starvation, not leaked capacity.

## Changes

**1. Merge the 4 TPU jobs into 2, grouped by container image**

| Job | Image | Now contains |
| --- | --- | --- |
| `run_core` | jax-stable-stack | core unit tests + sglang |
| `run_dev` | vllm-tpu | GRPO/SFT integration + vLLM |

Grouping by image (rather than by test area) means the image pull and `pip install -e .` are paid once instead of twice. Per-PR slice demand drops **4 to 2**, so two PRs can run concurrently where one previously blocked everything.

Two details worth reviewing:
- The sglang stage keeps a **step-level** `if:` rather than inheriting a job-level one, so the unit tests above it **still run on external fork PRs**, exactly as before.
- The sglang stage is pinned **last** in `run_core` because it force-reinstalls `jax[tpu]` and `qwix`; running it earlier would perturb the unit tests. There is a comment to that effect.

**2. `timeout-minutes` on every job** across all 8 workflows — previously **none** had one. TPU jobs are capped at 60–180min based on observed healthy duration, so a hang releases its slice instead of holding it.

**3. pip caching** on the TPU jobs. Each previously re-downloaded its full dependency tree while holding a slice.

**4. Cron fix.** `0 */16 * * *` did not mean "every 16 hours" — cron resets daily, so it fired at 00:00 and 16:00 (a 16h/8h split). Replaced with `0 4,16 * * *` for even 12h spacing, offset from the 10:00 UTC nightly regression.

**5. `paths-ignore`** for docs-only PRs.

## Trade-off

Wall-clock per PR rises from ~17min to ~31min, since merged stages run sequentially. That is the deliberate trade: half the slices for a longer individual run. Given the pool is idle 63% of the time and the bottleneck is 4-way burst contention, this should be a net win for end-to-end latency under load — but it is a visible regression for a single PR on an idle pool, and worth a second opinion.

## Verification

- All 8 workflow files parse.
- `actionlint`: 4 findings, all the pre-existing "unknown self-hosted runner label" warning (`main` has 6 of the same — fewer here only because there are fewer jobs). No new findings.
- Asserted programmatically: slice count == 2; `run_core` has no job-level `if:`; sglang is the last step.
- Confirmed the `main` ruleset has **no** `required_status_checks` rule, so dropping the `run_vllm` / `run_sglang` job names will not block merges, and `paths-ignore` will not strand docs-only PRs on a check that never runs.

**This PR needs the `ready-to-submit` label** to actually exercise the TPU jobs it changes — otherwise `tunix_tpu_unit_tests` is skipped and the merge is untested.

## Not included (deliberately)

- **The `run_vllm` exit-code bug.** `while read -r test; do pytest "$test"; done < file` returns only the *last* iteration's status, so a failure that is not last is silently discarded. It is a two-line fix in code this PR touches, but fixing it will likely turn that job red by surfacing failures it has been hiding — that deserves its own PR with someone watching.
- **Pre-baking sglang-jax / `jax[tpu]` / qwix into the container image.** Would reclaim most of the sglang stage's setup time, but needs image-publishing access rather than a repo change. pip caching recovers part of it.
- **Raising the pool from 4 to 8 slices.** Worth doing, but only after this lands; at 25% aggregate utilization the fan-out is the binding constraint, not the ceiling.


shz-google (commented):
PTAL. Thanks

.github/workflows/tpu-tests.yml
```diff
@@ -250,64 +280,3 @@ jobs:
...
-          # SGLang Jax removed qwix dependency in https://github.com/sgl-project/sglang-jax/pull/734
-          python3 -m pip install qwix --force-reinstall
-          python3 -m pip install 'jax[tpu]<0.11.0' --force-reinstall
-          # TODO(b/470113586): Remove this test once the bug is fixed.
-          python3 -m pip list | egrep 'jax|flax|libtpu'
-
-          cd tunix 
-          python3 tests/generate/sglang_jax_sampler_test.py
-          python3 tests/generate/sglang_jax_lora_test.py
-          python3 -m pytest tests/rl/rl_cluster_test.py -v --tb=short -k "not vllm"
```
s-noghabi, line 311: why are we removing our tests?

.github/workflows/build_and_test_tunix.yml
```diff
@@ -23,10 +23,24 @@ on:
...
+    # names, otherwise docs-only PRs will block on a check that never runs.
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+      - 'LICENSE'
+      - '.gitignore'
   workflow_dispatch:
   schedule:
-    # Run the job every 16 hours
-    - cron:  '0 */16 * * *'
```
s-noghabi, line 29: we dropped it to 16h due to resources. ideally it will be every 4h.

s-noghabi (changes requested):

sizhit2 (approved):

.github/workflows/tpu-tests.yml
shz-google: No tests are removed here — the `run_sglang` *job* is gone, but all three of its commands moved verbatim into `run_core` as its final stage (L163-186). The diff reads as a deletion because GitHub shows the removed job block; the identical steps were added ~170 lines above. Same for `run_vllm` → `run_dev`. I diffed every `pytest` invocation against `main`: the set is unchanged, 4 jobs → 2.

Two deliberate deltas: (1) the fork gate moved from job-level to step-level, so sglang still skips on external-fork PRs but the core unit tests above it now run there — strictly more coverage than today; (2) the stage is pinned last because it force-reinstalls `jax[tpu]`/`qwix` and would otherwise perturb the earlier unit tests. Cost is ~14min of serialized wall-clock per PR, which is the trade for halving per-PR slice demand.

.github/workflows/build_and_test_tunix.yml
shz-google: How about we run it every 12 hours. If we have more resources, we can try 4h.

.github/workflows/build_package.yml
```diff
@@ -26,6 +26,7 @@ jobs:
   build_and_upload:
     name: Build Tunix wheel
     runs-on: ubuntu-latest
+    timeout-minutes: 20
```
s-noghabi, line 29: what is the value of adding these timeouts? does queing time impact them?

shz-google: If one test get stuck, it will take all TPUs and wait w/o timeout until cancelled, which is 6 hours. See: https://docs.google.com/document/d/14-_rMLZNW0VfTFlolKy5pW01ecxQ5zzm4Nj_2V20DGQ/edit?tab=t.0

PR #2265, opened 2026-09-17
Deliver config that was accepted but never passed

entrpn:
Three gaps in the gsm8k distributed GRPO example.

The launcher never passed --temperature, --top_p or --top_k, so they sat at the runner's defaults and could not be configured. They now reach both the orchestrator and the inference node. The inference node scores the reference policy for the KL term and has to use the same temperature the rollout was drawn at, or the KL is biased. Defaults (1.0 / 1.0 / -1) match the previous behaviour.

grad_norm only went to the metrics logger, so it was invisible on a run with no metrics backend attached. It now appears on the train-step line next to loss and reward.

The gsm8k reward scored metadata["text"], which may hold the whole chat message list rather than the model's completion. It now takes the last assistant turn from the trajectory, falls back to metadata["text"] when that is a plain string, and scores an empty completion when there is no assistant turn to score.


tunix/experimental/examples/math_gsm8k_dist/gsm8k.py
```diff
@@ -105,14 +105,47 @@ def gsm8k_env_reward(
...
+
+
 def make_gsm8k_reward_fn(
     debug: bool = False,
 ) -> collections.abc.Callable[[Any], float]:
   """Creates an orchestrator-side reward function scoring completions against gold answers."""
 
   def reward_fn(item: Any) -> float:
     metadata = dict(getattr(item, "metadata", None) or {})
-    text = str(metadata.get("text", ""))
```
tianshub, line 115: @sizhit2 @hgao327 this should just be item.traj["conversation_text"] right? since that's the list of entire conv histories

tunix/experimental/examples/math_gsm8k_dist/gsm8k.py
```diff
@@ -105,14 +105,47 @@ def gsm8k_env_reward(
   return score_gsm8k_completion(str(completion), gold_answer)
 
 
+def _last_assistant_content(messages: Any) -> str | None:
```
tianshub, line 108: @hgao327 @sizhit2  instead of asking users to do this, this should be handled in the critique_stage by the framework

tunix/experimental/examples/math_gsm8k_dist/gsm8k.py
```diff
@@ -105,14 +105,47 @@ def gsm8k_env_reward(
...
+  """The last assistant message's content, or None if there is not one."""
+  if not isinstance(messages, (list, tuple)):
+    return None
+  for message in reversed(messages):
+    if isinstance(message, dict) and message.get("role") == "assistant":
+      return str(message.get("content") or "")
+  return None
+
+
+def _completion_text(item: Any, metadata: dict[str, Any]) -> str:
```
tianshub, line 118: @entrpn I think we can revert this change since same fix is applied on framework side in [cl/981476529](https://critique.corp.google.com/981476529)

PR #2270, opened 2026-09-17
Jfacevedo/trellis grpo recipe

entrpn:
## Summary

A rollout sampler and a trainer run the same weights through different kernels, so
their per-token log-probabilities for the same tokens disagree slightly. That
disagreement lands directly in the importance ratio `exp(log p_trainer − log q_sampler)`,
where a few sequences can dominate an update. This PR adds opt-in options to measure
that disagreement, drop the sequences where it is too large to train on, and correct
for what remains. It also adds a leave-one-out advantage estimator, and forwards a
loss function's aux metrics to the metrics logger, which the v2 trainer previously
discarded.

Every option defaults to off. Stacked on `jfacevedo/trellis-wiring-fixes` — review the
second commit only.

## New options on `GRPOConfig`

| option | default | failure mode it addresses |
| --- | --- | --- |
| `overlong_loss_masking` | `False` | The engine hits the response budget and the sequence is cut mid-answer, so a verifier scores a prefix. Drops it. Needs a per-trajectory truncation status. |
| `seq_logprob_error_threshold` | `None` | Sampler and trainer disagree so much that they are not the same policy — a template, tokenizer or weight-sync fault, not numerics. Drops the sequence when `mean_t exp\|log p_trainer − log q_sampler\|` exceeds this. Absolute per token, so `+d` on half the tokens and `−d` on the rest cannot cancel to a clean 1.0. |
| `truncated_importance_sampling_type` | `None` | Residual sampler/trainer drift. `"seq-mask-tis"` weights each token by its raw ratio and zeroes every weight of a sequence whose geometric-mean ratio (the sequence-level ratio of GSPO) leaves the keep-band. |
| `truncated_importance_sampling_ratio_min` / `_ratio` | `None` | Edges of that band. |
| `sampler_is_report_bands` | `()` | Report what other bands *would* have rejected, before anyone changes the live one. |
| `sampler_is_length_buckets` | `None` | Raw per-bucket sums from which `sqrt(observed_var / iid_var)` separates an offset that shrinks as `1/sqrt(T)` from one that is systematic and does not. |

`advantage_estimator` also accepts `"grpo-loo"`: baseline *and* standard deviation both
exclude the sample being scored. Including a sample in its own baseline scales its
advantage by `1 − 1/G` — 6.25% at G=16. Dividing by the leave-one-out std keeps prompts
of differing difficulty comparable, which is what distinguishes it from the existing
`"rloo"`.

GRPO [arXiv:2402.03300] · GSPO [arXiv:2507.18071] · RLOO [arXiv:2402.14740] ·
DAPO [arXiv:2503.14476]

## The distinction to check while reviewing

- **Masking** removes a sequence from the loss numerator *and its denominator*.
  Survivors keep their gradient magnitude; only the effective batch shrinks.
- **Zeroing an importance weight** removes it from the numerator only. The loss scales
  down with the rejection rate — so a 20% out-of-band rate is an effective
  learning-rate change, not a smaller batch.

| mechanism | leaves the denominator? |
| --- | --- |
| `overlong_loss_masking` | yes |
| `seq_logprob_error_threshold` | yes |
| `seq-mask-tis` | **no** — weights only |

This holds for `token-mean`, `sequence-mean-token-mean`, `sequence-mean-token-scale`
and `seq-mean-token-sum`, all pinned by test. It cannot hold for
`sequence-mean-token-sum-norm`, which normalises by a fixed factor rather than by the
mask, so the loss refuses to combine that mode with either masking option.

Note the keep-band deliberately does **not** clip weights inside the band: a large
in-band token ratio is unclipped and can reach ~1e38 before `exp` overflows and
`nan_to_num` collapses it to 0. `sampler_is/token_weight_max` is there to surface it.

## Two new per-sequence inputs

- **`rollout_per_token_logps`** `[B, C]` — the sampler's own log-probabilities. Kept
  separate from `old_per_token_logps` because a recipe may leave the latter `None` so
  the trainer recomputes it, pinning the PPO ratio to 1, while still wanting the
  sampler comparison. With one field that recipe is inexpressible.
- **`overlong`** `[B]` — 1.0 where generation stopped at the response budget without an
  end-of-sequence token. Read from the collector's trajectory status, accepting either
  a `TrajectoryStatus` member or its name, since both reach this point. Absent rather
  than `0.0` when the rollout reports no status.

`StandardRLProgram` requests rollout log-probabilities whenever any consumer needs
them — including the report-only diagnostics, and excluding `overlong_loss_masking`,
which reads the collector's verdict instead.

## What changes if you set none of these

With every option off `loss_mask == completion_mask`, so each mask substitution in
`grpo_loss_fn` is an exact no-op. Three things change anyway:

1. **New metric keys appear.** `_record_fwd_bwd` previously handed a loss function's
   aux dict only to `_post_process_train_step`, whose base implementation discards it,
   so nothing a loss reported reached the metrics buffer. Now unconditional on the
   train path (eval is untouched): `sample_mask/kept_frac`,
   `sampler_is/token_logdiff_{mean,absmean,abs_max}`, `sampler_is/hist_*`, `pos_*`,
   `conf_*`, plus `tis/is_oob_ratio` and `sampler_is/would_drop_<lo>_<hi>` when the
   corresponding option is on.
2. **Extremum metrics no longer report ±inf** for a micro-batch with no scored tokens.
   `advantage/max`, `advantage/min` and `is_ratio/min` fall back to `0.0`, which pools
   correctly across micro-batches where ±inf did not.
3. **The `sampler_is/*` attribution metrics compute unconditionally** when the rollout
   returns log-probabilities — a few dozen masked reductions per step over arrays the
   loss already holds. Straightforward to gate if that is unwelcome.

No previously-valid config is newly rejected: every new `ValueError` in `__post_init__`
reads a field introduced by this commit.

## Errors this can now raise

Sequence packing with any of the three sequence options; either log-probability gate
with no rollout log-probabilities; `overlong_loss_masking` with no truncation verdict;
a masking option under `sequence-mean-token-sum-norm`. Each refuses rather than letting
a correctness feature silently disable itself — which means a v1 learner, whose
`TrainExample` does not carry these fields, raises at step 1 rather than training with
the option quietly inert.

One asymmetry worth a reviewer's opinion: under packing the three gating options
refuse, while `sampler_is_length_buckets` emits neutral zeros. Report-only versus
correctness was the reasoning, but it is inconsistent on its face.

`GRPOConfig` additionally rejects a reversed or half-specified keep-band, an
unsupported TIS type, non-increasing length buckets, and `sampler_is` combined with
`seq-mask-tis` — both write the per-token weights, so the second would overwrite the
first.

## Where to look first

1. `tunix/rl/algorithm_config.py` — the options and their validation.
2. `tunix/rl/algo_core.py` — `sequence_loss_mask` and `truncated_importance_weights`;
   their docstrings state the denominator contract the tests pin.
3. `tunix/experimental/orchestrator/algorithm_adapter.py` and `batch_assembly.py` —
   how the two new fields reach the loss.
4. `tunix/experimental/train/peft_trainer_v2.py::_aux_to_additional_metrics` — the
   change with the widest blast radius, since it is not opt-in.

## Tests

90 new cases across `tests/rl/algo_core_test.py` and
`tests/experimental/orchestrator/grpo_recipe_wiring_test.py`: the numerics of each
gate, the denominator contract per aggregation mode, that the KL penalty is not
rescaled when a sequence is dropped, finite metrics when a micro-batch is emptied,
config validation, payload and batch-assembly plumbing, and metric forwarding.

Suites: `tests/rl` 397 passed, `tests/experimental/orchestrator` 383,
`tests/experimental/examples` 88, `tests/experimental/common` 39. The five `tests/rl`
collection errors are a missing `chex` in the venv and reproduce unchanged on `main`.

## Follow-ups, not in this PR

- A named `force_on_policy_ratio` option. The behaviour is reachable today via
  `use_rollout_logps=False`, but that field's documented meaning is "recompute on the
  trainer", which this path does not do.
- Reporting the sequence log-probability error when no threshold is set, so the
  distribution is visible before the gate is turned on.


tunix/experimental/examples/math_gsm8k_dist/run_gsm8k_dist_grpo.py
```diff
@@ -229,6 +234,55 @@ def _parse_args(argv: list[str]) -> argparse.Namespace:
           " on-policy ratio=1."
       ),
   )
+  # ---- Optional GRPO algorithm options -------------------------------------
```
andytwigg, line 237: @entrpn will these arguments need to be copied into examples/deepswe_dist and others?

entrpn: yes

PR #2279, opened 2026-09-17
DO NOT REVIEW Add a distributed RL smoke test to TPU CI (MaxText + vLLM + Raiden)

igorts-git:
## What

Adds one job, `run_maxtext_dist_smoke`, to `tpu-tests.yml`: a distributed GSM8K GRPO run that swaps the Tunix-native stack for the one the production RL recipe actually runs, on a single 8-chip host. Trainer and rollout are separate processes on separate chips, synchronising weights through Raiden, driven by the distributed orchestrator over two training steps.

| | existing `run_dist_smoke` | this job |
|---|---|---|
| Trainer | Tunix `peft_trainer_v2` | MaxText `MaxTextTrainingEngine` |
| Rollout | vLLM, Tunix model | vLLM server-mode, `MaxTextForCausalLM` |
| Weight sync | in-process | Raiden, across two live processes |

Both halves run MaxText's model code, so Raiden's name-based tensor pairing is exercised against real MaxText parameter names — the thing that breaks when either repo renames a weight. `SAMPLER=vllm` also puts the rollout behind a real worker RPC boundary; `run_dev` uses `inprocess_vllm`, which never crosses it, so signature skew between Tunix and tpu-inference cannot surface there.

Trainer is FSDP=4/TP=1 against a TP=4 rollout, deliberately: a matched TP=4/TP=4 pair would let a 1:1 copy pass for a working sync. Qwen3-0.6B, 2 steps, `train[:16]`. Gradient accumulation is 2 microbatches per optimizer step so the accumulate-then-apply path is covered — the orchestrator owns step boundaries, so a single-microbatch run would never call `train_step` with `apply_optimizer=False`.

## CI time cost

**~6 minutes, and no added wall-clock to the workflow.** The TPU jobs start together and run in parallel; `run_dev` at ~21 minutes sets the critical path, so this finishes well inside it. The real cost is one TPU runner slot for those ~6 minutes.

It does not run on every PR — `tunix_tpu_unit_tests` is gated on the `ready-to-submit` label — so in practice it runs on labelled PRs and on the 16-hourly schedule.

Two green runs, cold and warm caches: **6m45s** and **5m40s**. The difference is almost entirely container pull and cache restore (1m28s vs 16s); the test itself is stable at ~2m10s.

| | cold | warm |
|---|---|---|
| Container init + caches | 1m28s | 16s |
| Checkout | 9s | 9s |
| Install Tunix deps / MaxText / Raiden | 1m57s | 2m02s |
| TPU preflight | 13s | 13s |
| Checkpoint conversion (HF → Orbax) | 40s | 40s |
| **Smoke test** | **2m10s** | **2m07s** |
| **Total** | **6m45s** | **5m40s** |

If that ever needs trimming, an `actions/cache` on `artifacts/` would take the conversion to zero.

## Cross-repo versions

MaxText installs from a configurable `maxtext_ref`, **defaulting to HEAD**. The training engine lands on MaxText main several times a week and that is the only way cross-repo breakage surfaces the day it happens. Set a sha to pin, or the literal `pinned` to use the commit in `requirements/maxtext_requirements.txt`.

`install_maxtext.sh` captures and restores `jax/jaxlib/libtpu/numpy/torch/vllm/tpu-inference` around the install, so MaxText's dependency closure cannot silently move the versions the image's TPU runtime was built against. vLLM and tpu-inference stay pinned to the container image; Raiden stays pinned by URL+sha256.

## Also fixes a bug this test exposed

`run_trainer_node.py` now imports Raiden before anything touches a device, as `run_rollout_node.py` already did. Raiden's native modules and the TPU backend each statically link their own XLA, and with it their own copy of `xla/pjrt/proto/execute_options.proto`; whichever is dlopened second aborts the process in the protobuf descriptor registry. Without this the trainer runs normally right up to the first weight sync and then dies with no Python frame to point at. **The trainer is a weight-sync source in every RL run, so this is not specific to CI** — it's worth a look independent of the rest of the PR.

## Reviewer notes

- **New job, not a step in `run_dev`.** It mutates the shared interpreter in a way no other test in that job could survive — installing MaxText's dependency closure over the image's pinned vLLM stack. Per the file-header rule, saying so here.
- **This job's image is 4 days newer than `run_dev`'s**, and deliberately the smallest bump that works: same vLLM commit, tpu-inference `4b80060` → `20a6700`, the first nightly whose `TPUWorker.raiden_h2d` takes the `uuid` argument `vllm_sampler_v2.py` passes it. `run_dev` is unaffected because `inprocess_vllm` never crosses that boundary, so its pin is left alone.
- **`USE_ROLLOUT_LOGPS=false`**, against the production default, because `MaxTextTrainingEngine` does not implement `per_token_logps` — an abstract method of `AbstractTrainer`, so this is a gap in the engine, not a quirk of the test. With it on, `_apply_sampler_trainer_agreement` raises `AttributeError`. Cost: no sampler-vs-trainer agreement metrics and no TIS weights; the GRPO ratio pins to 1, which makes epsilon clipping inert. That is harmless at `MINI_BATCH_SIZE == BATCH_SIZE`, where there is a single inner iteration and the ratio would be 1 anyway, but it is why those two are kept equal here. Flip it back on once MaxText implements the method; nothing else needs to change.
- **Both sides run mcJAX.** Pathways is production's configuration and the only mode exercising the single-controller API the orchestrator is built on, so it was attempted — it reaches weight-sync round 0 and fails in Raiden's FFI source-prepare with `No FFI handler registered for init_weight_synchronizer_and_d2h on a platform Host`. FFI handler registration is per-process: the preload registers handlers in the trainer client, but under Pathways the computation compiles and executes in the Pathways server process, which never loads Raiden's shared objects. No client-side change reaches it. Rather than ship a default-off flag for a mode CI cannot run, the Pathways pieces live on **`igorts/pathways-local-support`** — including `LocalJaxContext`, which is a standalone Tunix fix: without it no local Pathways run can register the `proxy` backend at all. `igorts/pathways-ci-archive` holds the full working state for reference.

## Still open

- **Weight-sync correctness is not gated.** Structural breakage does hard-fail (`apply_to_runner` raises on a shape mismatch or unmatched tensors), but `VERIFY_WEIGHTS=true` only *logs* checksums on both sides and nothing compares them. Happy to add a compare step here or as a follow-up.
- **A passing run dumps no node logs.** The trainer/rollout/orchestrator logs are printed only `if: failure()`, so a green run's job log shows the loop exited cleanly but not what it produced. A tail on success would close that.

Separately noted while working on this, not fixed here: the `Cache HF hub` step caches `~/.cache/huggingface`, but `hf_pipeline` passes `local_dir=`, which bypasses `cache_dir` entirely — model weights land in `artifacts/` and are wiped by checkout, so every run re-downloads ~1.5 GB while holding a v6e-8. Affects the existing job too.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


scripts/start_pathways_daemons.sh
s-noghabi, line 1: Do we need to use pathways on this single node test?

Might be worth to add such a test in Trellis repo instead. wdyt?

scripts/install_maxtext.sh
s-noghabi, line 1: Wanted to head your thoughts, but I am thinking if generally this maxtext test should be in tunix ci or trellis ci? seems to me trellis is the right home, although we can temporarliy have it in tunix to catch issues sooner.

.github/workflows/tpu-tests.yml
```diff
@@ -366,3 +392,164 @@ jobs:
...
+  #   2. `Install Raiden weight sync library` MUST run LAST of the installs. It
+  #      force-installs the native `tpu_sync` wheel with --no-deps and then
+  #      compiles the gRPC protos; running it after MaxText means neither can be
+  #      displaced by MaxText's resolution.
+  #   3. `Build MaxText checkpoint` MUST run BEFORE the smoke test and does not
+  #      touch the TPU (JAX_PLATFORMS=cpu). It is a separate step only so its
+  #      runtime is attributable; the recipe calls the same script and the second
+  #      call is a no-op.
+  # ---------------------------------------------------------------------------
+  run_maxtext_dist_smoke:
```
s-noghabi, line 420: have you tested to see how long this test takes. we should aim for <10m, ideally around 5m or so.

igorts-git: It takes about 6m

PR #2317, opened 2026-09-20
[DRAFT] feat(mlperf): add MLPerf RCP logging support for distributed DeepSWE GRPO

susanbao:
## Summary
- **`tunix/utils/mllog_utils.py`**:
  - Add `log_rcp_step_stats` to emit the two MLPerf Reference Convergence Point (RCP) `tracked_stats` events per step (`reduced_train_loss`, `reward`, `grad_norm`, `global_valid_toks`, `global_valid_seqs` and `train_step_time`, `policy_training_time`, `exposed_generation_time`, `weight_sync_time`, `valid_tokens_per_sec_per_gpu`).
  - Support `gs://` destination paths in `logger_init` by writing to a local staging file and flushing to GCS via `fsspec` (`_flush_to_gcs_if_needed`) after `init_print`, `log_rcp_step_stats`, and `run_stop`.
  - Refine `init_print` hyperparameter and parallelism metadata (`grad_accum_steps`, `warmup_steps`, `lr_decay_steps`, `schedule_type`, `expert_parallelism`, `generation_expert_parallelism`).
- **`tunix/experimental/orchestrator/rl_program.py`**:
  - Decouple multi-turn rollout generation (`step_batches` collection) and contiguous microbatch training (`for mb in step_batches:`) in `StandardRLProgram.train_stage()` so `exposed_generation_time` and `policy_training_time` accurately capture non-overlapping rollout generation and TPU training durations without async TPU dispatch pollution.
  - Move `cb.on_step_begin` to the start of each step in `train_stage()` (after initial weight sync `wsync-v0-r0`) so `init_stop` / `run_start` / `block_start` align with Step 0 start and `init_start` -> `init_stop` includes initial weight synchronization.
  - Track and log `policy_training_time`, `exposed_generation_time`, `weight_sync_time`, `global_valid_toks`, and `global_valid_seqs`.
- **`tunix/experimental/examples/deepswe_dist/run_deepswe_dist.py`, `k8s_launcher.sh`, & `tunix/experimental/examples/recipes/mlperf_35b_256.sh`**:
  - Wire `--rcp_logging` (`RCP_LOGGING`) and MLPerf configuration flags (`--seed`, `--warmup_steps`, `--lr_decay_steps`, `--schedule_type`) into the distributed DeepSWE GRPO pipeline and `mlperf_35b_256.sh` recipe.
  - Add `MLLogStepTimerCallback` to invoke `mllog_utils.train_start` on `step == 0` and `mllog_utils.log_rcp_step_stats` at the end of each step.
- **`Dockerfile`**:
  - Install `mlperf-logging` (`git+https://github.com/mlcommons/logging.git@6.1.0-rc1`).

## Test Results (`Qwen3.5-35B-A3B` DeepSWE Distributed GRPO on `trellis`)
- **Configuration**:
  - Model: `Qwen3.5-35B-A3B` (`STAGE=deepswe`, `env_type=deepswe`, `agent_class=CodeActAgent`)
  - Topology: `64` train TPU v5p chips (`sanbao-train`) + `64` rollout TPU v5p chips across `16` vLLM replicas (`sanbao-roll-0..15`) on `bodaborg-v5p-nap` (`trellis`)
  - Batching: `MAX_STEPS=2`, `BATCH_SIZE=16` prompts x `NUM_GENERATIONS=16` = `256` trajectories/step (`8` microbatches x `32` trajectories per `train_step`)
  - Weight Sync: `WEIGHT_SYNC_MODE=direct` (Raiden DCN transport via `tpu_sync_jax`)
  - Output Log: `gs://sanbao-europe/mlperf/qwen35_35b/trellis/rcp/mllog/seed_42.out` (`EXIT_CODE=0`, final reward `mean=0.2461`)

## MLPerf Compliance Checker Results (`training_6.1.0`)
```text
INFO - Running compliance on file: /google/src/cloud/sanbao/sanbao/google3/experimental/users/sanbao/mlperf/mllog/trellis/deepswe_2.out
INFO -  Compliance checks: training_6.1.0/common.yaml
INFO -  Compliance checks: training_6.1.0/closed_common.yaml
INFO -  Compliance checks: training_6.1.0/closed_qwen35_397b_grpo.yaml
WARNING -  Required AT_LEAST_ONE occurrence of 'eval_accuracy' but found 0
ERROR - FAILED
```


tunix/experimental/orchestrator/rl_program.py
```diff
@@ -1390,103 +1426,106 @@ async def _maybe_save_checkpoint() -> None:
...
+          # abort the entire program. Make it configurable on whether to fail
+          # or continue.
+          full_batch_complete = (
+              groups_consumed >= self.full_batch_size or not scored_items
+          )
+          if full_batch_complete:
+            await _maybe_save_checkpoint()
+
+      if not scored_items and not checkpoint_saved and final_minibatch_completed:
+        await _maybe_save_checkpoint()
```
s-noghabi: have you done a real e2e run?

tunix/experimental/orchestrator/rl_program.py
```diff
@@ -1390,103 +1426,106 @@ async def _maybe_save_checkpoint() -> None:
...
+        step_batches.extend(self.assembler.feed(payloads))  # pyrefly: ignore[bad-argument-type]
+      exposed_generation_time = time.monotonic() - _t_gen
 
       if not all_step_items:
         logging.info(
             "Dataset exhausted at step %d before max_steps.", current_step
         )
         break
 
+      for mb in step_batches:
```
niting: Is this just a move of the existing code? I see some changes, but it's hard to figure out what changed.

niting (commented):
One minor question but I think we should get Shadi/Tianshu's review too since they better understand the orchestrator than I do.

tunix/experimental/orchestrator/rl_program.py
susanbao: Yes. This is just the move of the existing code to accurately accumulate the weight sync, training and rollout time.

susanbao:
> One minor question but I think we should get Shadi/Tianshu's review too since they better understand the orchestrator than I do.

@tianshub @s-noghabi Could you review the PR?

tunix/experimental/orchestrator/rl_program.py
```diff
@@ -1390,103 +1426,106 @@ async def _maybe_save_checkpoint() -> None:
...
+        policy_training_time += time.monotonic() - _t_train
+
+        if mb.is_final_batch and mb is step_batches[-1]:
+          # TODO(tunix-dev): Configurable checkpointing frequency. Today we
+          # checkpoint at the same frequency as the weight update.
+          # Save only at a resumable full-batch boundary. An optimizer step
+          # can occur earlier when a full batch contains multiple mini
+          # batches, but the dataset resume cursor advances in full batches.
+          # TODO(tunix-dev): For now any failures in save_checkpoint will
+          # abort the entire program. Make it configurable on whether to fail
```
s-noghabi: If I understand corrctly, you have changed the batching logic. This is wrong.

The PR changes StandardRLProgram.train_stage() from:

Python
get one group
→ assemble microbatches
→ immediately train those microbatches
→ get the next group
to:

Python
get all groups for the step
→ assemble and store all microbatches in step_batches
→ only then train every microbatch
This is a behavioral change, not just instrumentation.

Potential consequences
Higher host memory usage: all assembled microbatches for a full batch are retained before the first train_step() runs.
Different pipeline overlap: rollout polling/critique and TPU training no longer overlap within a step in the same way.
Different backpressure behavior: the trainer no longer drains scored_q while training each microbatch.
Different failure timing: errors from reference-logprob computation, sampler agreement, or training occur after the whole step has been collected.
Potential latency regression: the first optimizer work starts only after the complete batch has been assembled.
Timing semantics are still imperfect: exposed_generation_time measures the entire collection phase, including waiting for scored items and assembling batches—not necessarily only rollout generation.

s-noghabi (changes requested):

tunix/experimental/orchestrator/rl_program.py
susanbao: Yes. I have tested the code by Qwen3.5 r2egym e2e.

tunix/experimental/orchestrator/rl_program.py
susanbao: Thanks for catching this and listing out the tradeoffs! This change was made intentionally for **MLPerf Training v6.1.0 (`qwen35_397b_grpo`) RCP (`:::MLLOG`) logging**, to get accurate per-step timing breakdowns for `exposed_generation_time`, `policy_training_time`, and `weight_sync_time`.

### Why interleaving `train_step()` distorted RCP timing metrics

In our `Qwen3.5-35B-A3B` DeepSWE runs (`global_batch_size=256`, `16` prompt groups × `16` generations = `8` microbatches × `32` sequences), keeping `await self.engine.train_step(mb)` inside `while groups_consumed < self.full_batch_size:` caused severe distortion in the logged `:::MLLOG` metrics:

1. **`policy_training_time` was under-reported by 86.5% in steady state (`26.95s` logged vs. `202.72s` actual TPU time)**:
   - Microbatches 1–7 run with `apply_optimizer=False`, which JAX/XLA dispatches **asynchronously** on the Python host in `~0.07s–0.18s` while the TPU executes each `~25.2s` forward/backward microbatch in the background.
   - Only Microbatch 8 (`apply_optimizer=True`) blocks on the host to materialize training metrics (`~26.95s`).
   - When `train_step(mb)` was interleaved with `scored_q.get()`, `_t_train` only captured the `~0.18s` host dispatch for Microbatches 1–7 plus Microbatch 8 (`26.95s` total), missing `7 × 25.2s ≈ 176s` of actual TPU training execution.
2. **`exposed_generation_time` was polluted by async TPU training (`1045.96s` included `~176s` of TPU training)**:
   - After dispatching Microbatches 1–7 in `~0.18s`, execution immediately returned to `_t_gen` (`scored_q.get()` / `prepare_train_batch` / `compute_ref_logprobs`), absorbing the `~176s` of in-flight TPU training time into `exposed_generation_time`.

With the contiguous two-phase loop (collect `step_batches` under `_t_gen`, then run all 8 `train_step(mb)` calls back-to-back under `_t_train`), the device queue naturally backpressures across the 8 contiguous microbatches (`0.79s, 0.07s, 25.19s, 25.20s, 25.21s, 25.20s, 25.20s, 75.86s`), giving exact hardware timings:

| Metric (Steady-State Step 2, GBS=256) | Before (Interleaved, Run 2) | After (Contiguous `step_batches`, Run 4) | Discrepancy vs. Hardware |
| :--- | :--- | :--- | :--- |
| `exposed_generation_time` | `1045.96 s` *(polluted by async TPU)* | **`1489.19 s`** | **`0.00 s` (Exact)** |
| `policy_training_time` | `26.95 s` *(86.5% under-reported; only MB 8)* | **`202.72 s`** *(all 8 MBs @ ~25.2s/MB)* | **`0.00 s` (Exact)** |
| `weight_sync_time` | `291.52 s` | **`31.12 s`** | **`0.00 s` (Exact)** |
| `train_step_time` (`gen + train + sync`) | `1364.80 s` | **`1723.04 s`** (`1489.19 + 202.72 + 31.12 = 1723.03s`) | **`0.01 s` (`0.0003%`)** |

If this impacts performance or pipeline behavior, we can revert this change back to interleaved `train_step()` execution—the only downside is that the per-step phase times (`policy_training_time` and `exposed_generation_time`) shown in the RCP logging will be inaccurate (though `mlperf_logging.compliance_checker` itself does not fail on `tracked_stats`). Would you prefer that we revert this part?

susanbao: I have reverted the change on this batching logic. Please review again.

tunix/utils/mllog_utils.py
```diff
@@ -30,24 +30,93 @@
...
+    if metric_logger_dir.startswith("gs://"):
+      if metric_logger_dir.endswith(".out") or metric_logger_dir.endswith(
+          ".log"
+      ):
+        _gcs_target_path = metric_logger_dir
+      else:
+        _gcs_target_path = os.path.join(
+            metric_logger_dir.rstrip("/"), f"seed_{seed_val}.out"
+        )
+      filename = os.path.join("/tmp/rcp_logs", f"seed_{seed_val}.out")
```
s-noghabi: could this create race conditions if you have multiple workers on the same VM (single node case). Not an issue for mlperf, but would prevent single machine debugging

tunix/utils/mllog_utils.py
```diff
@@ -725,7 +907,11 @@ def init_print(
       platform = "TPU-Ironwood"
 
   # Gradient accumulation steps
-  grad_accum_steps = max(1, batch_size // mini_batch_size)
+  grad_accum_steps = max(
+      1,
+      batch_size // mini_batch_size,
+      (mini_batch_size * num_generations) // max(1, train_micro_batch_size),
```
s-noghabi, line 914: why add this?

tunix/utils/mllog_utils.py
susanbao: Good catch! In `run_deepswe_dist.py`, only the single CPU orchestrator process invokes `mllog_utils` (and `configure_logger` is additionally gated by `jax.process_index() == 0`), so co-located trainer/rollout workers won't write to this file.

However, to avoid any local file collision or stale-file appends on a shared VM (e.g., concurrent single-node debug runs or pre-init calls), I updated the local staging path in `/tmp/rcp_logs` to include `os.getpid()` (`seed_{seed_val}_{os.getpid()}.out`) while keeping the GCS destination path as `seed_{seed_val}.out`.

tunix/experimental/orchestrator/rl_program.py
```diff
@@ -1406,6 +1422,7 @@ async def _maybe_save_checkpoint() -> None:
             current_batch_idx, scored_items = ordered
         else:
           scored_items = await self.scored_q.get_group_batch(num_groups=1)
+        exposed_generation_time += time.monotonic() - _t_gen
```
s-noghabi, line 1430: why is this generation time? you can only get it by matching the rollout request & responses, or getting it from the rollout worker.

This is just reward calculation step

s-noghabi: we use a long polling, essentially we fire & forget.

s-noghabi (changes requested):

tunix/utils/mllog_utils.py
susanbao: Previously, `batch_size // mini_batch_size` always evaluated to `1` when `mini_batch_size == batch_size` (the default on-policy config), ignoring `train_micro_batch_size`.

In the distributed trainer (`run_trainer_node.py:_gradient_accumulation_steps`), each optimizer update processes `mini_batch_size * num_generations` trajectories in micro-batches of `train_micro_batch_size`, so the actual `gradient_accumulation_steps` passed to `PeftTrainer` is `(mini_batch_size * num_generations) // train_micro_batch_size`. This also keeps the logged MLPerf hyperparameters consistent (`global_batch_size = batch_size * num_generations` vs `micro_batch_size = train_micro_batch_size`).

tunix/experimental/orchestrator/rl_program.py
susanbao: Thank you for pointing this out. I have added TODO to this and will solve it later.

tunix/utils/mllog_utils.py
s-noghabi: These are different and you should decide which you want

batch_size // mini_batch_size: is the number of times we call model update() in a single global step
(mini_batch_size * num_generations) // max(1, train_micro_batch_size): is the number of times we call fwd_bwd() for a single mini batch (not global step). each time we call fwd_bwd() we do gradient accumulation. 

What is the metric you really want?

susanbao: Good point — we want the second one: the number of `fwd_bwd()` micro-steps accumulated per optimizer `update()` (`(mini_batch_size * num_generations) // max(1, train_micro_batch_size)`), which matches the MLPerf `gradient_accumulation_steps` definition and `run_trainer_node.py:_gradient_accumulation_steps`.

Removed `batch_size // mini_batch_size` so we don't mix update count per rollout step with gradient accumulation steps.

s-noghabi (approved):

PR #2335, opened 2026-09-21
Correct the TPU and GPU install steps in the launching guide

dkling-it:
`docs/launching.md` does not match `pyproject.toml`.

`jax[tpu]` is declared in the core `[project] dependencies`, so `pip install -e .` installs the TPU build of JAX on every platform. `[project.optional-dependencies]` declares `prod = []`, so `pip install -e .[prod]` adds nothing on top of that.

The guide still presents `[prod]` as the TPU path and the bare install as accelerator neutral. Someone on a GPU who follows Option B installs libtpu first and then replaces it, which is easy to miss until it turns up in the environment.

This is a documentation change. It describes what the current metadata actually does and says that an accelerator-specific install replaces the TPU build. No dependency, code or test is touched, and the existing `pip install jax[gpu]` example is unchanged.

If you would rather have `[prod]` carry the TPU dependency and keep the core install neutral, that is a `pyproject.toml` change instead and I am happy to redo it that way.

**Checklist**

- [x] I have added all the necessary unit tests for my change. (Not applicable, documentation only.)


dkling-it:
The CLA is signed and the `cla/google` check is passing. The comment above links to the earlier failed invocation and does not update itself.

PR #2368, opened 2026-09-22
fix(deepswe): scope fleet teardown to run selector and allow SANDBOX_NAMESPACE override

niting:
### Summary

1. **Scope fleet teardown to run selector**: Workaround for `agent-sandbox-rl` teardown bug by setting `c.resources.managed_selector = fleet_inst.run_selector` on each cluster registry entry after `SandboxFleet` initialization in `init_global_fleet()`. This prevents destructive sweeps from deleting other concurrent tenants' warm pools/templates in shared namespaces (such as `trellis`).
2. **Configurable SANDBOX_NAMESPACE**: Update `mlperf_35b_256.sh` to allow overriding `SANDBOX_NAMESPACE` via environment variable while defaulting to `trellis`:
   ```bash
   export SANDBOX_NAMESPACE="${SANDBOX_NAMESPACE:-trellis}"
   ```
3. **Unit tests**: Added unit test in `examples/deepswe/sandbox_utils_test.py` verifying that `init_global_fleet()` configures `managed_selector` to point to `run_selector`.


examples/deepswe/sandbox_utils.py
```diff
@@ -289,6 +289,11 @@ def init_global_fleet(
     fleet_cfg = FleetConfig(**fleet_kwargs)
     fleet_inst = SandboxFleet(fleet_cfg)
 
+    # Workaround for agent-sandbox-rl teardown bug: scope teardown to this run's
```
tomergee, line 292: Suggest a removal trigger and link so this gets deleted once the SDK is fixed, e.g.

```python
# TODO(<upstream issue>): remove once agent-sandbox-rl teardown is run-scoped upstream.
# Covers teardown only; unwarm_image / set_pool_replicas still act on pools by name.
```

Also worth naming claims here: teardown sweeps claims first (phase 1 in `_teardown`), and deleting a neighbour's live claim is the most visible symptom.

examples/deepswe/sandbox_utils.py
```diff
@@ -289,6 +289,11 @@ def init_global_fleet(
...
+    # run-id selector so teardown does not delete other concurrent tenants' warm pools/templates.
+    for c in getattr(fleet_inst, "registry", []):
```
tomergee, line 294: `getattr(fleet_inst, "registry", [])` degrades silently while `fleet_inst.run_selector` raises. Silent degradation here means the old namespace-wide sweep comes back with no warning, which is the worse failure. I would be strict on both and raise a clear error if either is missing: an SDK without `run_selector` also lacks the run-id label this selector depends on.

examples/deepswe/sandbox_utils_test.py
```diff
@@ -363,6 +363,21 @@ def test_init_global_fleet_configures_labels_and_teardown_hooks(self):
...
+    mock_cluster = mock.MagicMock()
+    mock_fleet.registry = [mock_cluster]
+    mock_fleet.run_selector.return_value = "agents.x-k8s.io/asrl-run-id=test-run"
+
+    mock_as_rl = mock.MagicMock()
+    mock_as_rl.SandboxFleet.return_value = mock_fleet
+    with mock.patch.dict("sys.modules", {"agent_sandbox_rl": mock_as_rl}):
+      with mock.patch.object(sandbox_utils, "_GLOBAL_FLEET", None):
+        _ = sandbox_utils.init_global_fleet(tasks=None, num_generations=4)
+        self.assertEqual(
```
tomergee, line 377: This proves the attribute was assigned, not that teardown uses it (the whole module is a MagicMock). It matches the file's pattern so not blocking, but at minimum call through:

```python
self.assertEqual(
    mock_cluster.resources.managed_selector(),
    "agents.x-k8s.io/asrl-run-id=test-run",
)
```

and cover a two-cluster registry. The real check belongs in `sandbox_k8s_e2e_test.sh`: after a teardown, a pool labelled with a foreign run id must still exist.

tomergee (commented):
Verified the override against the SDK: `SandboxFleet._teardown()` reads `c.resources.managed_selector()` through the instance (fleet.py:1173), `Cluster.resources` is a stored attribute (cluster.py:65), and the registry is built once in `SandboxFleet.__init__` and never rebuilt, so this does change what teardown lists on every path (atexit, signals, and the duplicate `teardown_global_fleet()` call). Objects carry the run-id label because the fleet stamps `config.labels` before the registry is built. Good stopgap, a few things to tighten:

- **It is one-directional.** It stops this run from sweeping others, but a tenant whose image predates it still deletes every `app=agent-sandbox-rl` claim/pool/template in the namespace on exit, which is exactly what happened on 09-22. With `SANDBOX_NAMESPACE` still defaulting to `trellis`, the next run is not protected until every tenant's image has this change (inline).
- **Not covered here, and not fixable in tunix:** a pool deleted mid-`wait_for_pool_ready` still burns the full 900 s `ready_timeout` because the SDK ignores the DELETED watch event; and `unwarm_image` / `set_pool_replicas` resolve pools by image-derived name, so two runs sharing an image in one namespace can still delete or resize each other's pools mid-training. Both are going into an upstream agent-sandbox-rl PR (run-scoped teardown, fail-fast on DELETED, `{run_id}` placeholder for names).
- **`Dockerfile:72` installs agent-sandbox-rl from upstream `main` unpinned**, so the SDK fix will arrive silently and this override then becomes dead weight that also defeats the planned `teardown_all_managed` opt-in. Worth pinning the SDK to a commit while the workaround is in place, or at least a TODO with a removal trigger (inline).

tunix/experimental/examples/recipes/mlperf_35b_256.sh
```diff
@@ -188,7 +188,7 @@ export DEBUG=1
 export DATASET_PATH="gs://mlperf_dataset/r2e-gym-easy"
 export USE_AGENT_SANDBOX=1
 export SCAFFOLD="openhands"
-export SANDBOX_NAMESPACE="trellis"
+export SANDBOX_NAMESPACE="${SANDBOX_NAMESPACE:-trellis}"
```
tomergee, line 191: Defaulting to `trellis` keeps the run in the shared namespace where old-image tenants still sweep it and same-image runs still collide by name. Suggest `${SANDBOX_NAMESPACE:-trellis-${USER}}` (or per run), or state in the PR description that `trellis` is only safe once every tenant's image includes this change.

Alternatively add a per-run `template_name_prefix` in `init_global_fleet` (e.g. `f"oh-{tag}-img-"`, pool names derive from it) so shared-image collisions stop even inside one namespace.

